PR #17, opened 2020-09-29
deprecating instrumentation libary span processing via config property

### LINKED ISSUE #14: Instrumentation name and version tags listed twice
One prefixed with `otel` and the other with `splunk`.

kubawach:
Resolves #14 

Since new OTLP Collector supports passing zipkin/jeager instrumentaiton library as span tags, span processing needs to be deprecated.
- `splunk.otel.config.span.processor.instrlib.enabled` property, set to false by default
- added unit tests to confirm the behaviour


jkwatson (approved):
Looks good!

iNikem (approved):
Looks good.

But I am not sure if we need these two attributes anymore. As Otel-java both Zipkin and Jaeger exporters write these attributes now, the only case when end user may miss them, if they use OTLP exporter and older Collector.  I am not sure we have such clients :)

kubawach:
> Looks good.
> 
> But I am not sure if we need these two attributes anymore. As Otel-java both Zipkin and Jaeger exporters write these attributes now, the only case when end user may miss them, if they use OTLP exporter and older Collector. I am not sure we have such clients :)

I'm a big fan of a code removal, in this case however wanted to leave us with some flexibility, just on case. I'll create a task to remove this whole code after some time, once we're sure no one will ever need that. Makes sense? :)

PR #27, opened 2020-10-26
Rename JDBC spans to lower cardinality names

mateuszrzeszutek:
Normally JDBC instrumentation puts the same value in `db.operation` and span name: full SQL query.


custom/src/main/java/com/splunk/opentelemetry/JdbcSpanRenamingProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  private boolean isJdbcSpan(AttributeValue dbSystem) {
+    return dbSystem != null
+        && dbSystem.getType() == AttributeValue.Type.STRING
+        && SQL_SYSTEMS.contains(dbSystem.getStringValue());
+  }
+
+  private String getStatementType(String query) {
```
iNikem, line 60: I think if you change this to `SELECT ` etc, then you can ditch `isJdbcSpan`

custom/src/main/java/com/splunk/opentelemetry/JdbcSpanRenamingProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+
+import io.opentelemetry.common.AttributeValue;
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+import java.util.HashSet;
+import java.util.Set;
+
+public class JdbcSpanRenamingProcessor implements SpanProcessor {
```
iNikem, line 27: Add javadoc please

custom/src/main/java/com/splunk/opentelemetry/JdbcSpanRenamingProcessor.java
mateuszrzeszutek: Hmm, you mean detecting SQL spans just by the prefix? Hmm, I think it's a bit safer this way - this processor won't rename any random span (not necessarily DB related) whose name begins with `select ` (e.g. span manually added by the instrumented app).

custom/src/main/java/com/splunk/opentelemetry/JdbcSpanRenamingProcessor.java
mateuszrzeszutek: Done.

custom/src/main/java/com/splunk/opentelemetry/InstrumentationLibraryTracerCustomizer.java
```diff
@@ -36,11 +39,22 @@ private static boolean spanProcessorInstrumentationLibraryEnabled() {
     return ("true".equalsIgnoreCase(value));
   }
 
+  private static boolean jdbcSpanLowCardinalityNameEnabled() {
+    String value = System.getProperty(ENABLE_JDBC_SPAN_LOW_CARDINALITY_NAME_PROPERTY);
+    if (value == null) {
+      value = System.getenv(propertyToEnv(ENABLE_JDBC_SPAN_LOW_CARDINALITY_NAME_PROPERTY));
+    }
+    // enabled by default
+    return value == null || "true".equalsIgnoreCase(value);
```
jkwatson, line 48: How about just `Boolean.valueOf(value)`. I think the logic is the same.

mateuszrzeszutek: Pretty close, but unfortunately not the same:

```java
    public static Boolean valueOf(String s) {
        return parseBoolean(s) ? TRUE : FALSE;
    }

    public static boolean parseBoolean(String s) {
        return "true".equalsIgnoreCase(s);
    }
```

jkwatson: so the difference is? From the javadoc, it looks like the same thing to me.

mateuszrzeszutek: Oh, I just realized that you meant just the  `"true".equalsIgnoreCase(value)` part, not the entire line 🤦‍♂️

mateuszrzeszutek: In that case, yes, you're exactly right -- I'll change that.

jkwatson: right.  :). sorry for the confusion.

custom/src/main/java/com/splunk/opentelemetry/JdbcSpanRenamingProcessor.java
iNikem: Ok, makes sense. I just did not like this hard-copied list of db systems :(

iNikem (approved):

PR #39, opened 2020-11-13
Java Agent Migration - add steps for shim and classpath

brianashby-sfx:
- adds a new step to the migration workflow that includes using the OpenTracing Shim and updating the `classpath`

### DIFF
--- a/migration.md
+++ b/migration.md
@@ -54,7 +54,6 @@ distribution of Splunk Distribution of OpenTelemetry Java Instrumentation:
    annotation in the SignalFx Java Agent, ensure it's in your application's
    `classpath` as well. For an example of what this looks like, see this
    [SignalFx Java Agent example application](https://github.com/signalfx/tracing-examples/blob/master/signalfx-tracing/signalfx-java-tracing/okhttp-and-jedis/src/main/java/com/signalfx/tracing/examples/javaagent/App.java).
-   
 
 ## Changes in functionality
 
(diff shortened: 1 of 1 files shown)


migration.md
```diff
@@ -46,6 +46,14 @@ distribution of Splunk Distribution of OpenTelemetry Java Instrumentation:
    setting.
 4. In your application startup script, replace `-javaagent:./signalfx-tracing.jar`
    with `-javaagent:/path/to/splunk-otel-javaagent-all.jar`.
+5. If you manually instrumented any code with an OpenTracing tracer, expose
+   the OpenTelemetry tracer as an implementation of an OpenTracing tracer with
+   the OpenTracing Shim. For more information, see
+   [OpenTelemetry - OpenTracing Shim](https://github.com/open-telemetry/opentelemetry-java/tree/master/opentracing-shim).
```
jkwatson, line 52: Is there anything special you need to do to enable the OT Shim when running with the java agent? @iNikem ?

iNikem: Does [this PR](https://github.com/signalfx/tracing-examples/pull/60/files) answer your question?

iNikem (approved):

migration.md
jkwatson: If that code works, then yes!  😉 
I thought that it would probably "just work" since the shim only uses the API. Thanks!

PR #42, opened 2020-11-20
Use alwaysOn() sampler

mateuszrzeszutek:
Will always export the trace, even if the incoming b3 headers do not contain the `x-b3-sampled` one

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/SplunkTracerCustomizer.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/SplunkTracerCustomizer.java
@@ -44,8 +44,7 @@ public class SplunkTracerCustomizer implements TracerCustomizer {
       tracerManagement.addSpanProcessor(new JdbcSpanRenamingProcessor());
     }
 
-    tracerManagement.updateActiveTraceConfig(tracerManagement.getActiveTraceConfig().toBuilder()
-            .setSampler(Sampler.alwaysOn())
-            .build());
+    tracerManagement.updateActiveTraceConfig(
+        tracerManagement.getActiveTraceConfig().toBuilder().setSampler(Sampler.alwaysOn()).build());
   }
 }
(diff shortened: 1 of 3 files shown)


custom/src/main/java/com/splunk/opentelemetry/SplunkTracerCustomizer.java
```diff
@@ -42,5 +43,9 @@ public void configure(TracerSdkManagement tracerManagement) {
     if (jdbcSpanLowCardinalityNameEnabled()) {
       tracerManagement.addSpanProcessor(new JdbcSpanRenamingProcessor());
     }
+
+    tracerManagement.updateActiveTraceConfig(tracerManagement.getActiveTraceConfig().toBuilder()
+            .setSampler(Sampler.alwaysOn())
```
jkwatson, line 48: The SDK itself defaults to alwaysOn. Is the otel agent doing something different?

mateuszrzeszutek: The SDK actually defaults to

```java
  private static final Sampler DEFAULT_SAMPLER = Sampler.parentBased(Sampler.alwaysOn());
```

which works slightly different: for spans whose parent is not sampled it also returns the "not sampled" flag

jkwatson: aha! You are totally correct. I conflated the two in my head.

jkwatson (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/SpringBootSmokeTest.java
```diff
@@ -28,7 +28,9 @@
 class SpringBootSmokeTest extends SmokeTest {
 
   private String getTargetImage(int jdk) {
-    return "open-telemetry-docker-dev.bintray.io/java/smoke-springboot-jdk" + jdk + ":latest";
+    return "open-telemetry-docker-dev.bintray.io/java/smoke-springboot-jdk"
+        + jdk
+        + ":20201105.347264626";
```
mateuszrzeszutek, line 33: I'm not entirely sure that this change is OK, but it seems that the `latest` image is simply too new, something related to @WithSpan must have changed.

jkwatson: Yes, it got moved to a different package, and a different artifact in the past couple of weeks.

iNikem (approved):

PR #43, opened 2020-11-23
Scripts to build docker images for acceptance test matrix

iNikem:


jkwatson (approved):
Excellent. Do we think this is something we'll eventually contribute upstream?

iNikem:
> Excellent. Do we think this is something we'll eventually contribute upstream?

Yes, we have such plans :)

mateuszrzeszutek (approved):

vovencij:
Will the support for pushing images to some repo follow?

iNikem:
> Will the support for pushing images to some repo follow?

Good point, let me investigate that...

vovencij:
Also, why shell scripts? This will not scale for building Windows images and we will need those next. Using gradle, for example, would work nicely cross-platform.

iNikem:
You cannot build both linux and windows images from the same dockerfiles, can you? How can this work "cross-platform"? Or what do I miss?

vovencij:
My point is that you have a bunch of scripts, that replace variables in the templates. For windows, there will be just another set of templates and variables. So this replacement logic will have to be reimplemented from scratch (because shell script won't work in windows) and there will be 2 ways for doing almost the same thing.

vovencij:
And then there will be 3rd place where similar logic will be duplicated - when test would want to use an image from the matrix. Super-ideally this image-name composition logic could be in just one place. But I don't know how to share this logic between build scripts and tests.

iNikem:
@vovencij replaces shell scripts with gradle tasks. Image publishing still does not work as we don't have permissions to setup it

vovencij:
Can you make `matrix` a proper sub-project with gradle tasks under there? As a next step I will add building of `app.war` in that same project.

vovencij:
✅  I would now approve, if I had the right :)

PR #50, opened 2020-12-01
Java Phase 1, Tomcat support and test.

vovencij:


instrumentation/tomcat/src/main/java/com/splunk/opentelemetry/middleware/TomcatAttributesInstrumentationModule.java
```diff
@@ -0,0 +1,77 @@
...
+@AutoService(InstrumentationModule.class)
+public class TomcatAttributesInstrumentationModule extends InstrumentationModule {
+
+  public TomcatAttributesInstrumentationModule() {
+    super("tomcat");
+  }
+
+  @Override
+  public ElementMatcher.Junction<ClassLoader> classLoaderMatcher() {
+    return hasClassesNamed("org.apache.catalina.startup.Catalina");
```
mateuszrzeszutek, line 46: Consider adding `ServerInfo` here:
```suggestion
    return hasClassesNamed("org.apache.catalina.startup.Catalina", "org.apache.catalina.util.ServerInfo");
```

mateuszrzeszutek (approved):

instrumentation/tomcat/src/main/java/com/splunk/opentelemetry/middleware/TomcatAttributesInstrumentationModule.java
vovencij: What's the general approach to listing classes in `classLoaderMatcher`? Should I list classes, that I will be instrumenting, or all classes that I will be somehow touching?

mateuszrzeszutek: In this repo I'd list all used classes in `InstrumentationModule#classLoaderMatcher()`. That's because there's no muzzle reference check here, which normally (in the otel agent) ensures that all classes/methods/fields used are present.

smoke-tests/src/test/java/com/splunk/opentelemetry/TomcatSmokeTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class TomcatSmokeTest extends AppServerTest {
+
+  public static final ExpectedServerAttributes TOMCAT7_SERVER_ATTRIBUTES =
+      new TomcatAttributes("7.0.107.0");
```
iNikem, line 30: How fragile are these versions? I believe we build docker images with "current latest for this major version".

vovencij: There are multiple scenarios to this :) If we publish our matrix images and will not rebuild them - then the version won't change :) If we build every time before tests - they most probably will break on next released version.

I see 2 potential solutions - fix the version in the matrix or use partial matching in assertions. I'd vote for fixing versions in the matrix.

iNikem (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/TomcatSmokeTest.java
vovencij: As discussed in slack, versions will be fixed while upstreaming the matrix for fully open-source app-servers.

PR #53, opened 2020-12-09
Detect liberty version

laurit:


build.gradle.kts
```diff
@@ -21,7 +21,7 @@ subprojects {
 
   extra.set("versions", mapOf(
       "opentelemetry" to "0.12.0",
-      "opentelemetryJavaagent" to "0.12.0-SNAPSHOT"
+      "opentelemetryJavaagent" to "0.13.0-SNAPSHOT"
```
iNikem, line 24: Do you need 0.13 snapshot or 0.12? The latter has been published at last: https://search.maven.org/search?q=g:io.opentelemetry.javaagent

instrumentation/compile-stub/README.md
```diff
@@ -0,0 +1,4 @@
+# Compile stubs
+In case instrumentation code needs a class that is not available in a public repository you can add a stub class that has the methods called from integration code to this module.
```
iNikem, line 2:
```suggestion
In case instrumentation code needs a class from the instrumented library that is not available in a public repository you can add a stub class that has the methods called from integration code to this module.
```

instrumentation/compile-stub/src/main/java/com/ibm/ws/kernel/productinfo/ProductInfo.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright Splunk Inc.
```
iNikem, line 2: I don't think we can put Splunk copyright here, if you have copied a class verbatim. We have to honour the original license.

instrumentation/liberty/src/main/java/com/splunk/opentelemetry/middleware/LibertyAttributesInstrumentationModule.java
```diff
@@ -0,0 +1,87 @@
...
+        // ibm distribution has both
+        if ("openliberty.properties".equals(p.getFile().getName())
+            || "WebSphereApplicationServer.properties".equals(p.getFile().getName())) {
+          version = p.getVersion();
+          break;
+        }
+      }
+
+      MiddlewareHolder.trySetVersion(version);
+      MiddlewareHolder.trySetName("websphere liberty");
```
iNikem, line 84:
```suggestion
      MiddlewareHolder.trySetName("WebSphere Liberty");
```

instrumentation/compile-stub/README.md
```diff
@@ -0,0 +1,4 @@
+# Compile stubs
```
iNikem, line 1: That's nice idea 👍

iNikem (approved):

build.gradle.kts
laurit: i currently don't need 0.13-SNAPSHOT, but probably we'll need it soon

instrumentation/compile-stub/src/main/java/com/ibm/ws/kernel/productinfo/ProductInfo.java
laurit: this class is not copied. it only has methods that are needed by integration with the same signature as origin class has, but method bodies just throw UnsupportedOperationException instead of doing what they do in original class. if you have a better idea for what the copyright should be let me know.

iNikem: What was the original license?

laurit: https://github.com/OpenLiberty/open-liberty/blob/master/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/productinfo/ProductInfo.java

instrumentation/liberty/src/main/java/com/splunk/opentelemetry/middleware/LibertyAttributesInstrumentationModule.java
laurit: i would leave it as it is for now because Ivo has changes planned for this anyway. i believe we'll have to change it to include both short name and full version string. also note that tomcat and jetty currently use lowercase. if change is needed i'd change them also.

instrumentation/compile-stub/src/main/java/com/ibm/ws/kernel/productinfo/ProductInfo.java
iNikem: Ok, they are indeed very different.

vovencij (approved):

PR #55, opened 2020-12-10
Convert WLS images to the new matrix format

vovencij:
* Introduced notion of proprietary images to the build matrix. 
* Converted existing weblogic image creation scripts to the new matrix format. 
* In the webapp deployed to splunk distro test images, the span created from `@WithSpan` annotation is still present, because this test was legitimately broken on WLS 14/jdk11 until there was a fix for bytebuddy.

image

P.S. JDK in wls image names is weird because we use images build from oracle Dockerfiles and this is how it is there.


matrix/build.gradle.kts
```diff
@@ -26,7 +26,11 @@ val dockerWorkingDir = project.buildDir.resolve("docker")
 
 val buildTestImagesTask = tasks.create("buildTestImages") {
```
iNikem, line 27: As these are now built by Otel, can we remove this?

vovencij: And leave only proprietary?

vovencij: wait, liberty is also not present in otel?

vovencij: Removed template and matrix definition, updated README.

iNikem (approved):

PR #56, opened 2020-12-10
Enable sending traces directly to ingest

mateuszrzeszutek:


custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
```diff
@@ -16,23 +16,40 @@
...
+              .build());
+    }
+
+    return builder.build();
   }
 
   @Override
   public Set<String> getNames() {
-    return Collections.singleton("jaeger-thrift");
+    return Collections.singleton("jaeger-thrift-splunk");
```
mateuszrzeszutek, line 53: I think we need a separate name since we're adding our custom logic - and the TODO comment is no longer viable here.

custom/src/main/java/com/splunk/opentelemetry/jaeger/AuthTokenInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class AuthTokenInterceptor implements Interceptor {
+  static final String TOKEN_HEADER = "X-SF-TOKEN";
+
+  private final String signalfxAuthToken;
+
+  public AuthTokenInterceptor(String token) {
+    signalfxAuthToken = token;
```
iNikem, line 30:
```suggestion
    this.signalfxAuthToken = token;
```

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
iNikem: TODO comment is indeed no longer applicable. But I am not sure if we need separate exporter name. If `SIGNALFX_AUTH_TOKEN` is not given, it will behave exactly like Otel's `jaeger-thrift`. I think it will be less confusing for clients to keep the same name. Assuming that we will not conflict with Otel when https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1609 is done

mateuszrzeszutek:
> Assuming that we will not conflict with Otel when open-telemetry/opentelemetry-java-instrumentation#1609 is done

Unfortunately I think that we will have a conflict here. Otel javaagent takes the first exporter with a given name, so if we have 2 named `jaeger-thrift` then it's pretty much random which one gets chosen. Having a separate name for ours is more error-proof.

```java
  private static SpanExporterFactory findSpanExporterFactory(String exporterName) {
    ServiceLoader<SpanExporterFactory> serviceLoader =
        ServiceLoader.load(SpanExporterFactory.class, TracerInstaller.class.getClassLoader());

    for (SpanExporterFactory spanExporterFactory : serviceLoader) {
      if (spanExporterFactory.getNames().contains(exporterName)) {
        return spanExporterFactory;
      }
    }
    return null;
  }
```

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
mateuszrzeszutek: > Assuming that we will not conflict with Otel when open-telemetry/opentelemetry-java-instrumentation#1609 is done

Unfortunately I think that we will have a conflict here. Otel javaagent takes the first exporter with a given name, so if we have 2 named jaeger-thrift then it's pretty much random which one gets chosen. Having a separate name for ours is more error-proof.

```java
  private static SpanExporterFactory findSpanExporterFactory(String exporterName) {
    ServiceLoader<SpanExporterFactory> serviceLoader =
        ServiceLoader.load(SpanExporterFactory.class, TracerInstaller.class.getClassLoader());

    for (SpanExporterFactory spanExporterFactory : serviceLoader) {
      if (spanExporterFactory.getNames().contains(exporterName)) {
        return spanExporterFactory;
      }
    }
    return null;
  }
```

kubawach: I agree. Now there is no difference between these two if TOKEN is empty, but the other may change in time so "shadowing" it could lead to some issues in the future.

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
```diff
@@ -16,23 +16,40 @@
...
 import io.opentelemetry.javaagent.spi.exporter.SpanExporterFactory;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.util.Collections;
 import java.util.Properties;
 import java.util.Set;
+import okhttp3.OkHttpClient;
 
-// TODO remove when https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1609
-// is implemented
 public class JaegerThriftSpanExporterFactory implements SpanExporterFactory {
```
kubawach, line 28: Shouldn't we test this (new) behaviour as well? To be sure that everything works properly with ant without token (ie no exception is thrown if not provided).

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
```diff
@@ -16,23 +16,40 @@
...
+    String token = config.getProperty(SIGNALFX_AUTH_TOKEN, "");
+    if (!token.isEmpty()) {
+      builder.setThriftSender(
+          new HttpSender.Builder(config.getProperty(OTEL_EXPORTER_JAEGER_ENDPOINT))
+              .withClient(
+                  new OkHttpClient.Builder()
+                      .addInterceptor(new AuthTokenInterceptor(token))
+                      .build())
+              .build());
+    }
```
kubawach, line 46: As "no token" situation is considered as a valid one, I'd log (at INFO) that no token was provided. Can help in the future when investigating possible issues - we will know that eg auth fails as no token is present.

mateuszrzeszutek: 👍  but at DEBUG

custom/src/main/java/com/splunk/opentelemetry/jaeger/AuthTokenInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class AuthTokenInterceptor implements Interceptor {
+  static final String TOKEN_HEADER = "X-SF-TOKEN";
+
+  private final String signalfxAuthToken;
+
+  public AuthTokenInterceptor(String token) {
```
kubawach, line 29: Shouldn't we fail here if no token is provided (assertion) ?

kubawach (approved):
LGTM, some very minor comments / questions only.

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
mateuszrzeszutek: It's not easy to unit test produced `SpanExporter` - this can be done, but either through checking fields via reflection or changing every constructor call to `Function<?, ?>` to stub that. I don't think that's worth it, and we have smoke tests anyway.

custom/src/main/java/com/splunk/opentelemetry/jaeger/AuthTokenInterceptor.java
mateuszrzeszutek: 🤷‍♂️

iNikem (approved):

PR #57, opened 2020-12-10
Support for WildFly version detection

iNikem:


instrumentation/wildfly/src/main/java/com/splunk/opentelemetry/middleware/WildFlyAttributesInstrumentationModule.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public ElementMatcher<? super TypeDescription> typeMatcher() {
+      return named("org.jboss.as.version.ProductConfig");
+    }
+
+    @Override
+    public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+      return Collections.singletonMap(
+          isMethod().and(isStatic()).and(named("fromFilesystemSlot")),
```
vovencij, line 64: Currently every time this method is called, we try to set middleware attributes, which will result in warning messages from MiddlewareHolder. Have you checked how many times this is called during the startup of the wildfly?

smoke-tests/src/test/java/com/splunk/opentelemetry/WildFlySmokeTest.java
```diff
@@ -0,0 +1,74 @@
...
+            WILDFLY_17_SERVER_ATTRIBUTES),
+        arguments(
+            "ghcr.io/open-telemetry/java-test-containers:wildfly-21.0.0.Final-jdk8-20201207.405832649",
+            WILDFLY_21_SERVER_ATTRIBUTES),
+        arguments(
+            "ghcr.io/open-telemetry/java-test-containers:wildfly-21.0.0.Final-jdk11-20201207.405832649",
+            WILDFLY_21_SERVER_ATTRIBUTES));
+  }
+
+  @ParameterizedTest
```
vovencij, line 64: `@ParameterizedTest(name = "[{index}] {0}")` - image name is enough for a test name.

vovencij (changes requested):

instrumentation/wildfly/src/main/java/com/splunk/opentelemetry/middleware/WildFlyAttributesInstrumentationModule.java
iNikem: Very good point. But why do we write debug log in recurring set? If I make addition check here, that will be 2 CAS instead one of 1.

vovencij: Checking it here is not desired, yes. Initial assumption for this MiddlewareHolder was that there is a possibility to instrument _something_ during server startup, that will happen only once and as early as possible, so that middleware attributes will be set by the time a server will start serving requests. And the log will hint the instrumentation point is not the best. So far there were always some init methods, that would cache server version somewhere for internal needs, e.g. JMX queries etc. Maybe wildfly also has something similar, that will call this method and cache the result for later usage?

iNikem: Fixed

vovencij (approved):

PR #62, opened 2020-12-11
Remove no longer needed JDBC span rename

mateuszrzeszutek:
Low-cardinality JDBC span names are now implemented in the upstream


jkwatson (approved):
Thanks! Do we need to update any release-notes or changelog or anything along with this?

iNikem (approved):

PR #65, opened 2020-12-17
Introduce nightly build with upstream snapshot

vovencij:
No automatic issue creation for now, because until GA lots of breaking changes are expected.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,13 @@ release {
 
 group = "com.splunk.public"
 
+boolean snapshot = getProperty("snapshot") == 'yes'
+
+def versions = [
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -12,17 +12,21 @@ release {
 
 group = "com.splunk.public"
 
+boolean snapshot = getProperty("snapshot") == 'yes'
```
iNikem, line 15: I thought `getProperty` throws an exception if no such property exists?

vovencij: As you can see in PR build - it doesn't.

iNikem (approved):

PR #67, opened 2020-12-21
Update README.md

slernersplunk:
Make instructions for ENV variables clearer for customers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,10 +91,10 @@ A simple wrapper for the Jaeger exporter of opentelemetry-java. gRPC is currentl
 
 | System property                   | Environment variable              | Description                                                                                                         |
 |-----------------------------------|-----------------------------------|---------------------------------------------------------------------------------------------------------------------|
-| otel.exporter=jaeger              | OTEL_EXPORTER=jaeger              | Select the Jaeger exporter                                                                                          |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -91,10 +91,10 @@ A simple wrapper for the Jaeger exporter of opentelemetry-java. gRPC is currentl
 
 | System property                   | Environment variable              | Description                                                                                                         |
 |-----------------------------------|-----------------------------------|---------------------------------------------------------------------------------------------------------------------|
-| otel.exporter=jaeger              | OTEL_EXPORTER=jaeger              | Select the Jaeger exporter                                                                                          |
+| otel.exporter=jaeger              | OTEL_EXPORTER=jaeger              | Optional: Select the Jaeger exporter for gRPC capable environments. This should not be set in a default Kubernetes setuip.                                                                                          |
```
jkwatson, line 94: I wouldn't say "default" here. That makes a lot of assumptions about the world. I'd just mention that it's not compatible with the smart agent.

slernersplunk:
update in next pull req

README.md
slernersplunk: It is compatible with SmartAgent on a host... just not in K8S.

PR #72, opened 2021-01-07
specify compatible otel API versions to extend manual instrumentation

brianashby-sfx:
- clarify that you need to use compatible API versions to extend Splunk distribution manual instrumentation
- specify compatible otel java versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,11 @@ For more options see [here](https://github.com/open-telemetry/opentelemetry-java
 Documentation on how to manually instrument a Java application are available
 [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation#manually-instrumenting).
 
+To extend the instrumentation with the OpenTelemetry Instrumentation for Java,
+you have to use a compatible API version. The Splunk distribution of 
+OpenTelemetry Java Instrumentation version 0.4.0 is compatible with the
+OpenTelemetry Instrumentation for Java version 0.12.0 and 0.12.1.
+
 ## Inject trace and span ID into logs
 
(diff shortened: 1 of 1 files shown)


jkwatson (approved):
Seems reasonable. This info should also go in the release notes.

README.md
```diff
@@ -120,6 +120,11 @@ For more options see [here](https://github.com/open-telemetry/opentelemetry-java
 Documentation on how to manually instrument a Java application are available
 [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation#manually-instrumenting).
 
+To extend the instrumentation with the OpenTelemetry Instrumentation for Java,
+you have to use a compatible API version. The Splunk distribution of 
+OpenTelemetry Java Instrumentation version 0.4.0 is compatible with the
+OpenTelemetry Instrumentation for Java version 0.12.0 and 0.12.1.
```
jkwatson, line 126:
```suggestion
OpenTelemetry Instrumentation for Java version 0.12.0 and API version 0.12.1.
```

vovencij:
Currently with every new release upstream versions have to be updated in 3 places now: `build.gradle`, README, release notes. Maybe particular versions should be in release notes only? And README just say, that `please see release notes for your version for a list of compatible versions`. Because the fixed readme is not included in already released version and it is already outdated for the HEAD, if anyone should release 0.5.0 right now, because HEAD already has a dependency on upstream 0.13.

vovencij:
... aand it just happened :)

jkwatson:
This points to a need to have a documented release process. And, more release automation.

PR #73, opened 2021-01-08
Fix snapshot builds.

breedx-splk:
Upstream opentelemetry-java-instrumentation publishes snapshots to sonatype, so if we're doing a snapshot build let's include that repo.

The snapshot builds that are supposed to run 2 hours after the upstream builds have been failing, and I think this will fix that.


vovencij (approved):
Seems to be working. Thanks!

PR #76, opened 2021-01-12
Expose resource attributes as system properties to use them in logs

vovencij:
Correlating logs with traces requires logging at least some resource attributes. From the agent point of view, proposed solution imposes the least performance overhead, as it requires no constant run-time overhead whatsoever and properties can be access from all major logging frameworks directly.

### DIFF
--- a/smoke-tests/src/test/java/com/splunk/opentelemetry/logger/ResourceAttributesForLoggingTest.java
+++ b/smoke-tests/src/test/java/com/splunk/opentelemetry/logger/ResourceAttributesForLoggingTest.java
@@ -20,8 +20,6 @@ import com.splunk.opentelemetry.SmokeTest;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
 import java.util.concurrent.TimeUnit;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Assumptions;
@@ -61,7 +59,7 @@ public class ResourceAttributesForLoggingTest {
 
   private Process startProcessWithJavaagent() throws IOException {
     File javaBinary = new File(System.getProperty("java.home"), "bin/java");
-    String javaagent = "-javaagent:s" + SmokeTest.agentPath;
+    String javaagent = "-javaagent:" + SmokeTest.agentPath;
     String resourceAttributes =
         "-Dotel.resource.attributes=service.name=MyService,environment=development";
     String classpath = System.getProperty("java.class.path");
@@ -77,13 +75,5 @@ public class ResourceAttributesForLoggingTest {
     System.out.println(String.join(" ", strings));
     processBuilder.command(strings);
     return processBuilder.start();
-  }
-
-  private void transfer(InputStream from, OutputStream to) throws IOException {
-    byte[] buffer = new byte[65535];
-    int read;
-    while ((read = from.read(buffer, 0, buffer.length)) >= 0) {
-      to.write(buffer, 0, read);
-    }
   }
 }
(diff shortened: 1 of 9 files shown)


smoke-tests/src/test/java/com/splunk/opentelemetry/logger/ResourceAttributesForLoggingTest.java
```diff
@@ -0,0 +1,89 @@
...
+    return processBuilder.start();
+  }
+
+  private void transfer(InputStream from, OutputStream to) throws IOException {
+    byte[] buffer = new byte[65535];
+    int read;
+    while ((read = from.read(buffer, 0, buffer.length)) >= 0) {
+      to.write(buffer, 0, read);
+    }
+  }
```
mateuszrzeszutek, line 88: I think you can use `InputStream#transferTo(OutputStream)` instead

smoke-tests/src/test/java/com/splunk/opentelemetry/logger/ResourceAttributesForLoggingTest.java
```diff
@@ -0,0 +1,89 @@
...
+    transfer(process.getInputStream(), out);
+    ByteArrayOutputStream err = new ByteArrayOutputStream();
+    transfer(process.getErrorStream(), err);
+
+    Assertions.assertTrue(process.waitFor(30L, TimeUnit.SECONDS), "Process exited");
+
+    int exitValue = process.exitValue();
+    if (exitValue != 0) {
+      System.err.println(err.toString());
+    }
```
mateuszrzeszutek, line 53: Nitpicking: you can use `Assertions#assertEquals(int, int, Supplier<String>)` to lazily compute assertion message instead of an if statement here.

smoke-tests/src/test/java/com/splunk/opentelemetry/logger/ResourceAttributesForLoggingTest.java
```diff
@@ -0,0 +1,89 @@
...
+      javaagent,
+      resourceAttributes,
+      "-cp",
+      classpath,
+      LoggerTestMain.class.getName()
+    };
+    ProcessBuilder processBuilder = new ProcessBuilder();
+    System.out.println(String.join(" ", strings));
+    processBuilder.command(strings);
+    return processBuilder.start();
```
mateuszrzeszutek, line 79: I must say that this is a clever way of running a separate Java process without using docker/testcontainers 👍

laurit (approved):

mateuszrzeszutek (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/logger/ResourceAttributesForLoggingTest.java
vovencij: Good catch. I somehow was under impression, that we're limited to JDK 8 here. But apparently not.

smoke-tests/src/test/java/com/splunk/opentelemetry/logger/ResourceAttributesForLoggingTest.java
vovencij: This wasn't even about lazy evaluation. It was about having stderr in logs, if exit code was non-zero. But having it in the assertion message is even better, indeed!

smoke-tests/src/test/java/com/splunk/opentelemetry/logger/ResourceAttributesForLoggingTest.java
mateuszrzeszutek: 👍 
Can you remove this function now that it's no longer used?

vovencij: Yes, sure. I was under impression that I did that. Apparently not yet...

PR #79, opened 2021-01-13
Get rid of duplicate files inside agent jar

laurit:
Jar files can have multiple entries with the same name. After this change there is only one entry for each name. Agent size is reduced from 39M to 30M.

### DIFF
--- a/agent/build.gradle
+++ b/agent/build.gradle
@@ -1,20 +1,38 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
 plugins {
     id("com.github.johnrengelman.shadow") version "6.0.0"
 }
 configurations {
+    // dependencies that are moved to inst
     customShadow
+    // dependencies that need package relocation
+    customShadowInclude
+    // dependencies that are not moved to inst and don't need package relocation
+    mainShadow
+    compileOnly.extendsFrom(mainShadow)
 }
 dependencies {
     customShadow project(path: ":custom", configuration: "shadow")
     customShadow project(path: ":instrumentation", configuration: "shadow")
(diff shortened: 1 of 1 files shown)


agent/build.gradle
```diff
@@ -28,33 +36,44 @@ CopySpec isolateSpec() {
...
+    // merge otel agent with our agent
     shadowJar {
-        dependsOn ':custom:shadowJar'
-        dependsOn ':instrumentation:shadowJar'
-        dependsOn ':bootstrap:jar'
-        with isolateSpec()
+        from tasks.customShadow.outputs
 
-        /// TODO - with this set to EXCLUDE, service files will not be merged between instrumentation and custom
-        duplicatesStrategy = DuplicatesStrategy.INCLUDE
```
vovencij, line 39: Is this TODO now solved?

laurit: Yes. I compared extracted jar before and after changes and it had the following differences.
- com/splunk/opentelemetry/SplunkAgent.class has constant pool elements in different order, otherwise classes are identical. New version matches the one in agent/build/classes
- ordering of elements in inst/META-INF/services files is different. Previously entries from splunk were before otel now they are after.
- new jar does not have a directory entry for inst/META-INF/services, idk where it went, maybe mergeServiceFiles does something with it. But it shouldn't be needed anyway, directory entries in jar files are optional and as far as I understand won't affect service loader.

vovencij: The problem with previous TODO was that if there were service files in both upstream and custom, then they *must* be merged, but with `EXCLUDE` Strategy instead of merging just one was randomly kept.

laurit: verified, service files are merged

laurit: verified, service files are merged

vovencij (approved):

PR #80, opened 2021-01-14
Explicitly include guava version to avoid conflicts with upstream.

breedx-splk:
This old version caused an old version of `Preconditions` to sneak into the shadowed jar -- which, especially when combined with #79, made the agent startup fail when configured with oltp exporter.

### DIFF
--- a/custom/build.gradle
+++ b/custom/build.gradle
@@ -11,9 +11,9 @@ dependencies {
   implementation("io.opentelemetry:opentelemetry-exporter-jaeger-thrift:${versions["opentelemetry"]}")
   implementation("io.opentelemetry.javaagent:opentelemetry-javaagent-spi:${versions["opentelemetryJavaagent"]}")
   implementation("io.jaegertracing:jaeger-client:1.5.0")
-  annotationProcessor("com.google.auto.service:auto-service:1.0-rc3")
+  annotationProcessor("com.google.auto.service:auto-service:1.0-rc7")
   annotationProcessor("com.google.auto:auto-common:0.8")
-  implementation("com.google.auto.service:auto-service:1.0-rc3")
+  implementation("com.google.auto.service:auto-service:1.0-rc7")
   implementation("com.google.auto:auto-common:0.8")
 }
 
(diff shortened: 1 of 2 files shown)


jkwatson (approved):
👍

breedx-splk:
Grrrrr..... :rage2: 

Actually, I just looked [in the agent code](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/8d74baa2e4d186b1e78ebdc88ae0a7f73a297e69/gradle/dependencies.gradle#L61) and they are on rc3.  So this might be harmful...

breedx-splk:
Closing this until we can figure out what's up.

breedx-splk:
Dropped back down to rc3 but explicitly include the version of guava that our upstream agent has included.  I wish there would be a better way to do this (like using a bom or similar) because finding this and adding this will be a significant maintenance burden.

jkwatson (approved):

custom/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
   annotationProcessor("com.google.auto:auto-common:0.8")
   implementation("com.google.auto.service:auto-service:1.0-rc3")
   implementation("com.google.auto:auto-common:0.8")
+  implementation("com.google.guava:guava:29.0-android")
```
vovencij, line 18: Why this version? [Upstream has](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/8d74baa2e4d186b1e78ebdc88ae0a7f73a297e69/gradle/dependencies.gradle#L15) `"20.0", // Last version to support Java 7`

vovencij: apparently `custom` depends on both `sdk` and `javaagent` and they themselves have different guavas.

mateuszrzeszutek: It's a wonder that it works at all 😂

breedx-splk:
Yeah I don't think we should do this.  At least the other change that shades the upstream agent first gets us closer without manual maintenance of these versions.

PR #81, opened 2021-01-14
Collect WebLogic specific attributes

agoallikmaa:
This adds collection of WebLogic-specific attributes to servlet spans, and also provides middleware name and version through `MiddlewareHolder`.

Sample output from the logging collector:
```
Attributes:
     -> middleware.name: STRING(WebLogic Server)
     -> middleware.version: STRING(12.2.1.4.0)
     -> middleware.weblogic.server: STRING(admin-server)
     -> middleware.weblogic.application: STRING(weblogic)
     -> middleware.weblogic.domain: STRING(domain1)
```

No spans are created at WebLogic instrumentation level (no custom logic required), so this adds an instrumentation that only collects the attribute values, stores them as a map into `otel.middleware` ServletRequest attribute. The map contents are copied over to the span in another instrumentation where the servlet span is currently active.

WebLogic classes are not published as a public dependency, therefore its classes are accessed with `Class.forName` and `MethodHandle` instead of compiling against its classes directly. These are resolved by `WebLogicEntity` during its static initialization, which also provides wrapper classes for WebLogic classes. As an optimization, the full set of attributes is not collected on every request, but they are cached to a servlet context attribute.

### DIFF
--- a/instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/middleware/weblogic/WebLogicAttributeCollector.java
+++ b/instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/middleware/weblogic/WebLogicAttributeCollector.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.middleware.weblogic;
+
+import com.splunk.opentelemetry.javaagent.bootstrap.MiddlewareHolder;
+import java.util.HashMap;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+
+public class WebLogicAttributeCollector {
+  private static final String CONTEXT_ATTRIBUTE_NAME = "otel.weblogic.attributes";
+  public static final String REQUEST_ATTRIBUTE_NAME = "otel.middleware";
+
+  public static void attachMiddlewareAttributes(HttpServletRequest servletRequest) {
+    WebLogicEntity.Request request = WebLogicEntity.Request.wrap(servletRequest);
+
+    Map<?, ?> attributes = fetchMiddlewareAttributes(request.getContext());
+    request.instance.setAttribute(REQUEST_ATTRIBUTE_NAME, attributes);
+  }
+
+  private static Map<?, ?> fetchMiddlewareAttributes(WebLogicEntity.Context context) {
+    if (context.instance == null) {
+      return null;
+    }
+
+    Object value = context.instance.getAttribute(CONTEXT_ATTRIBUTE_NAME);
+
+    if (value instanceof Map<?, ?>) {
+      return (Map<?, ?>) value;
+    }
+
+    // Do this here to avoid duplicate work
+    storeMiddlewareIdentity(context);
+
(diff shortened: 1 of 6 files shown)


instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/middleware/weblogic/WebLogicEntity.java
```diff
@@ -0,0 +1,266 @@
...
+    Class<?> mbeanClass;
+
+    try {
+      requestGetContext =
+          MethodHandles.publicLookup()
+              .findVirtual(
+                  Class.forName("weblogic.servlet.internal.ServletRequestImpl"),
+                  "getContext",
+                  MethodType.methodType(
+                      Class.forName("weblogic.servlet.internal.WebAppServletContext")));
```
vovencij, line 53: Can you use compile stubs approach for this instead, as is done for liberty? https://github.com/signalfx/splunk-otel-java/tree/main/instrumentation/compile-stub

smoke-tests/src/test/java/com/splunk/opentelemetry/WebLogicSmokeTest.java
```diff
@@ -54,15 +57,29 @@ public void webLogicSmokeTest(String imageName) throws IOException, InterruptedE
     // FIXME: APMI-1300
     //    assertServerHandler(....)
```
vovencij, line 58: You can replace this FIXME with just a comment, that there aren't any server handlers for weblogic because of APMI-1300.

vovencij (changes requested):

instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/middleware/weblogic/WebLogicEntity.java
mateuszrzeszutek: One advantage of the current approach is that it's completely safe even in the absence of muzzle - we should preserve that feature and try-catch everything when converting to compile stub.

instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/middleware/weblogic/WebLogicAttributeCollector.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class WebLogicAttributeCollector {
+  private static final String CONTEXT_ATTRIBUTE_NAME = "otel.weblogic.attributes";
+  public static final String REQUEST_ATTRIBUTE_NAME = "otel.middleware";
+
+  public static void attachMiddlewareAttributes(HttpServletRequest servletRequest) {
+    WebLogicEntity.Request request = WebLogicEntity.Request.wrap(servletRequest);
+
+    Map<?, ?> attributes = fetchMiddlewareAttributes(request.getContext());
+    request.instance.setAttribute(REQUEST_ATTRIBUTE_NAME, attributes);
```
mateuszrzeszutek, line 32: Just wondering: have you considered adding a field to `MiddlewareHolder` and storing those attributes there instead of the request?

sedmelluq: The server and application fields may have different values in a single JVM, so the current static field approach `MiddlewareHolder` would not work there.

instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/middleware/weblogic/WebLogicEntity.java
agoallikmaa: This would add a lot of interfaces and classes with just one method in each, and zero in some - and this is if the interface hierarchy is changed to skip the intermediate ones we don't actually need, which is not so nice. An additional problem is that to have classes such as `ServletRequestImpl` stub also implement `ServletRequest`, we would need to include the servlet as a dependency for the stubs as well, which I am not sure is a good idea. So I'm not sure in this case the stubs would actually make it better, since even the below example uses inaccurate hierarchy, which is not very elegant.

image

vovencij: Let it stay as is then.

vovencij (approved):

iNikem (approved):

PR #87, opened 2021-01-19
Windows smoke tests for Jetty, Tomcat, GlassFish

agoallikmaa:
Adds tasks and Dockerfiles to build Windows Docker images for:

- OpenTelemetry collector
- Fake backend (required duplicating it from OpenTelemetry repository)
- Jetty
- Tomcat
- GlassFish (Payara)

Implements an alternative way of test container management that works with Windows images (by using docker-java directly) as Testcontainers does not support Windows images. Platform and type (proprietary or not) are now also specified for images used in smoke tests to determine which images can be tested with on the current system.

Running Windows smoke tests also added to CI workflow, where for now, the Windows images are built locally before running tests.


vovencij (commented):
What would be the steps if one would like to push windows tests upstream?

iNikem:
Quick question: why `fake-backend` on windows cannot be upstreamed to Otel?

matrix/build.gradle
```diff
@@ -18,53 +18,85 @@ dependencies {
...
+
+def windowsTargets = [
+    "tomcat" : [
+        [version: "7.0.107", majorVersion: "7"]: ["8"],
+        [version: "8.5.61", majorVersion: "8"]: ["8", "11"],
+        [version: "9.0.41", majorVersion: "9"]: ["8", "11"]
+    ],
+    "jetty" : [
+        [version: "9.4.35", sourceVersion: "9.4.35.v20201120"]: ["8", "11", "15"],
+        [version: "10.0.0", sourceVersion: "10.0.0.beta3", classifier: "split"]: ["11", "15"]
```
iNikem, line 53: What is this `split` and why do we need this?

matrix/build.gradle
```diff
@@ -18,53 +18,85 @@ dependencies {
...
+createDockerTasks(buildProprietaryTestImagesTask, proprietaryTargets, false)
+
+def buildWindowsTestImagesTask = tasks.create("buildWindowsTestImages") {
+  group = "build"
+  description = "Builds all Windows Docker images for the test matrix"
+}
+
+def windowsTargets = [
+    "tomcat" : [
+        [version: "7.0.107", majorVersion: "7"]: ["8"],
```
iNikem, line 47: Why in case of Windows we have separate `version` and `majorVersion`?

matrix/build.gradle
```diff
@@ -18,53 +18,85 @@ dependencies {
...
+        [version: "8.5.61", majorVersion: "8"]: ["8", "11"],
+        [version: "9.0.41", majorVersion: "9"]: ["8", "11"]
+    ],
+    "jetty" : [
+        [version: "9.4.35", sourceVersion: "9.4.35.v20201120"]: ["8", "11", "15"],
```
iNikem, line 52: And why here we have third version, `sourceVersion`?

matrix/src/jetty-split.windows.dockerfile
```diff
@@ -0,0 +1,20 @@
+ARG jdk
+ARG sourceVersion
+
+# Unzip in a separate container so that zip file layer is not part of final image
+FROM mcr.microsoft.com/windows/servercore:1809 as builder
+ARG sourceVersion
+ADD https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-home/${sourceVersion}/jetty-home-${sourceVersion}.zip /server.zip
+RUN ["powershell", "-Command", "expand-archive -Path /server.zip -DestinationPath /server"]
+
+FROM winamd64/openjdk:${jdk}-jdk-windowsservercore-1809
```
iNikem, line 10: Is there a reason to use different images for this main container and for the `builder`? If we could use the same image it would make builds faster.

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
```diff
@@ -0,0 +1,524 @@
...
+import org.rnorth.ducttape.ratelimits.RateLimiterBuilder;
+import org.rnorth.ducttape.unreliables.Unreliables;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WindowsTestContainerManager extends AbstractTestContainerManager {
+  private static final Logger logger = LoggerFactory.getLogger(WindowsTestContainerManager.class);
+
+  private static final String NPIPE_URI = "npipe:////./pipe/docker_engine";
+  private static final String COLLECTOR_CONFIG_FILE_PATH = "/collector-config.yml";
```
iNikem, line 67: Why do we need this to differ from `AbstractTestContainerManager#COLLECTOR_CONFIG_RESOURCE`?

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
```diff
@@ -0,0 +1,524 @@
...
+import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+import org.apache.commons.io.IOUtils;
+import org.rnorth.ducttape.TimeoutException;
+import org.rnorth.ducttape.ratelimits.RateLimiter;
+import org.rnorth.ducttape.ratelimits.RateLimiterBuilder;
+import org.rnorth.ducttape.unreliables.Unreliables;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WindowsTestContainerManager extends AbstractTestContainerManager {
```
iNikem, line 63: Are this class and `LinuxTestContainerManager` functionally equivalent? Or is `LinuxTestContainerManager` more powerful?

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
```diff
@@ -66,15 +67,20 @@ protected void assertWebAppTrace(ExpectedServerAttributes serverAttributes)
         1, traces.countFilteredAttributes("http.url", url), "The span for the initial web request");
     Assertions.assertEquals(
         2,
-        traces.countFilteredAttributes("http.url", "http://localhost:8080/app/headers"),
+        traces.countFilteredAttributes("http.url", getUrl("/app/headers", true)),
         "Client and server spans for the remote call");
 
+    if (testImage.isProprietaryImage) {
```
iNikem, line 73: Do we need this `WithSpan`? Or more precisely: why do we need test application which differs from Otel's one? We can publish Otel testing app and reuse it. Will make our tests simpler.

As first step we can remove `WithSpan` method from our testing app.

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
```diff
@@ -100,11 +106,15 @@ private String tryGetResponse(Request request) throws IOException {
     return responseBody;
   }
 
-  protected void additionalWebAppTraceAssertions(
-      TraceInspector traces, ExpectedServerAttributes serverAttributes) {}
+  protected void additionalWebAppTraceAssertions(TraceInspector traces, TestImage testImage) {}
 
-  protected int totalNumberOfSpansInWebappTrace() {
-    return 3;
+  protected int totalNumberOfSpansInWebappTrace(
```
iNikem, line 111: This is never overwritten, so can be private. Also `serverAttributes` is not used.

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
```diff
@@ -100,11 +106,15 @@ private String tryGetResponse(Request request) throws IOException {
     return responseBody;
   }
 
-  protected void additionalWebAppTraceAssertions(
-      TraceInspector traces, ExpectedServerAttributes serverAttributes) {}
+  protected void additionalWebAppTraceAssertions(TraceInspector traces, TestImage testImage) {}
```
iNikem, line 109: Nobody overwrites this, is this needed?

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/TargetWaitStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+  public static class Log extends TargetWaitStrategy {
+    public final String regex;
+
+    public Log(Duration timeout, String regex) {
+      super(timeout);
+      this.regex = regex;
+    }
+  }
+
+  public static class Http extends TargetWaitStrategy {
```
iNikem, line 37: Is this used?

smoke-tests/src/test/java/com/splunk/opentelemetry/SpringBootSmokeTest.java
```diff
@@ -25,18 +28,22 @@
...
+        "open-telemetry-docker-dev.bintray.io/java/smoke-springboot-jdk"
+            + jdk
+            + ":20201105.347264626");
   }
 
   @ParameterizedTest(name = "{index} => SpringBoot SmokeTest On JDK{0}.")
   @ValueSource(ints = {8, 11, 15})
   public void springBootSmokeTestOnJDK(int jdk) throws IOException, InterruptedException {
-    startTarget(getTargetImage(jdk));
+    TestImage image = getTargetImage(jdk);
```
iNikem, line 41: Inline

smoke-tests/src/test/java/com/splunk/opentelemetry/SpringBootSmokeTest.java
```diff
@@ -25,18 +28,22 @@
...
+
+    startTargetOrSkipTest(image);
 
-    String url = String.format("http://localhost:%d/greeting", target.getMappedPort(8080));
+    String url =
+        String.format("http://localhost:%d/greeting", containerManager.getTargetMappedPort(8080));
```
iNikem, line 46: This can now use `AppServerTest.getUrl`, right?

smoke-tests/src/test/java/com/splunk/opentelemetry/TomcatSmokeTest.java
```diff
@@ -30,39 +33,63 @@
       new TomcatAttributes("7.0.107.0");
   public static final ExpectedServerAttributes TOMCAT8_SERVER_ATTRIBUTES =
       new TomcatAttributes("8.5.60.0");
+  public static final ExpectedServerAttributes TOMCAT8561_SERVER_ATTRIBUTES =
+      new TomcatAttributes("8.5.61.0");
```
iNikem, line 37: Why add this version?

iNikem:
Will other app servers follow to Windows as well?

vovencij:
> Quick question: why `fake-backend` on windows cannot be upstreamed to Otel?

I think we have to look at the wider picture here - upstreaming just fake-backend is not very useful without at least one test that would use it. But for that we also would need the windows-dockerized collector. These could be follow up steps within the effort of open-sourcing the test matrix.

matrix/build.gradle
agoallikmaa: Anything beyond `version` (used for image name), `classifier` (used to choose which Dockerfile to use) is just an argument directly passed to the `Dockerfile`, so that we can have clean appserver name and version fields without having to leak Dockerfile specific details into those fields.

matrix/build.gradle
agoallikmaa: Anything beyond `version` (used for image name), `classifier` (used to choose which Dockerfile to use) is just an argument directly passed to the `Dockerfile`, so that we can have clean appserver name and version fields without having to leak Dockerfile specific details into those fields.

matrix/build.gradle
agoallikmaa: `classifier` is used to be able to choose a different `Dockerfile` without changing the name of the appserver. Different versions may require setup that is different enough that they require a separate `Dockerfile`, yet it makes no sense to have the final image have a different name just because of that.

matrix/src/jetty-split.windows.dockerfile
agoallikmaa: I made sure that the former was the base image for the latter, so this is not an issue.

smoke-tests/src/test/java/com/splunk/opentelemetry/SmokeTest.java
```diff
@@ -222,4 +178,14 @@ protected String getCurrentAgentVersion() throws IOException {
         .get(Attributes.Name.IMPLEMENTATION_VERSION)
         .toString();
   }
+
+  private static TestContainerManager createContainerManager() {
+    boolean isWindows = System.getProperty("os.name").toLowerCase().contains("windows");
+
+    if (isWindows && !"1".equals(System.getenv("USE_LINUX_CONTAINERS"))) {
```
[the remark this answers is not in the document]
agoallikmaa: This is in case you are running Docker for Windows in Linux container mode and thus want to run Linux containers instead.

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
agoallikmaa: The difference in values is not actually necessary, but I think separate fields still make sense, because they are used in different places - the `RESOURCE` one refers to the filename in the test code classpath, and the `PATH` refers to the filename used in container filesystem.

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/TargetWaitStrategy.java
agoallikmaa: As the logic for waiting behind http endpoint was internally already implemented as it was used for fake backend, it made sense to keep the list of wait strategies in sync with the capabilities that are already implemented anyway.

matrix/build.gradle
agoallikmaa: Just to specify, for Tomcat this is necessary because the major version alone is also used in the URLs, so it makes more sense to pass this in as an argument than try to perform string manipulation within a Dockerfile.

matrix/build.gradle
agoallikmaa: In this specific case, it causes the `jetty-split.windows.dockerfile` to be used instead, which is used for Jetty versions where HOME and BASE must be split to two different locations, and the download URLs for those are also different, so separate Dockerfile made sense.

smoke-tests/src/test/java/com/splunk/opentelemetry/ProprietaryAppServerTest.java
```diff
@@ -16,33 +16,8 @@
...
-    try {
-      DockerClientFactory.lazyClient().inspectImageCmd(imageName).exec();
-      return true;
-    } catch (Exception e) {
-      System.out.println(e.getMessage());
-      return false;
-    }
-  }
-}
+abstract class ProprietaryAppServerTest extends AppServerTest {}
```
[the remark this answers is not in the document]
agoallikmaa: No. Removed.

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
agoallikmaa: Removed.

smoke-tests/src/test/java/com/splunk/opentelemetry/SpringBootSmokeTest.java
agoallikmaa: Done.

smoke-tests/src/test/java/com/splunk/opentelemetry/SpringBootSmokeTest.java
agoallikmaa: Done, had to make it extend `AppServerTest` first though.

smoke-tests/src/test/java/com/splunk/opentelemetry/TomcatSmokeTest.java
agoallikmaa: Removed. Initially this was because I used a download endpoint in Dockerfile building which only had the latest version, but since I changed that, it is no longer necessary.

agoallikmaa:
> Quick question: why `fake-backend` on windows cannot be upstreamed to Otel?

It could, but then that would be a blocker for this, and I'm not sure if upstreaming the backend alone would make sense until some Windows tests are upstreamed as well.

> What would be the steps if one would like to push windows tests upstream?

Pretty much just adapt all the additional logic in both `matrix` and `smoke-tests` upstream.

matrix/build.gradle
iNikem: Why do we need to test these two Jetty setups? Do we expect them to behave differently with javaagent?

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
iNikem: And that PATH is only relevant for Windows-based images?

matrix/build.gradle
agoallikmaa: No. They are not different setups, the required directory layout and download URLs were just changed starting from version 10. I did not want to expose this internal detail to the final Docker image, therefore the classifier argument, which does not leak to the final image name, is used instead to specify that a different Dockerfile should be used instead. Otherwise there would have to be a separate "jetty-10" server entry, which would cause the image name to become "jetty-10-10.0.0" which would not be nice.

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
agoallikmaa: The path only needs to be consistent across the copy to container operation and command used to launch the collector, which are both done in the same class. Therefore it is an internal implementation details of the container manager and there is no need to expose it to any other class.

smoke-tests/src/test/java/com/splunk/opentelemetry/SmokeTest.java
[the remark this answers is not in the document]
agoallikmaa: Just to clarify, it will have to be provided manually by the user in that case.

vovencij (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
iNikem: @agoallikmaa ^^ ?

iNikem:
Rebase this PR please

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
agoallikmaa: They effectively perform the same task, but `LinuxTestContainerManager` uses `testcontainers`, while `WindowsTestContainerManager` uses `docker-java` directly, as `testcontainers` does not support Windows containers. Another difference is obviously that `WindowsTestContainerManager` uses Windows collector and fake backend images while `LinuxTestContainerManager` uses Linux ones.

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
agoallikmaa: This was present before, and I'd rather leave any functional changes to the test application out of the scope of this PR, especially since it may warrant further discussion that is not relevant to the topic of this PR.

agoallikmaa:
> Will other app servers follow to Windows as well?

I do not see any reason why not, but they are not in the scope of my current task.

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
iNikem: My point here is we use cannot use testcontainers on both platforms and thus brought `docker-java`, can we still have one `ContainerManger` for both platforms? Or do we benefit somehow from using testcontainers for Linux? Currently seems like duplicated effort.

sedmelluq: There are actually some nice-to-haves in the testcontainers implementation that I did not duplicate as that would not have been worth it just for this one use case - such as containers being more reliably cleaned up without relying on the stop/kill being called from Java code.

Also, Testcontainers would be the preferred way to do this, and hopefully they'll get around to supporting Windows images. And if they do, then the `docker-java` implementation could be thrown away and replaced with a few ifs for Windows.

Ironically this cleanup mechanism with an auxiliary container is what's blocking them from supporting Windows, as they seem to be stubborn about this container always having to be a Linux container, even though what runs on it is written in Go which would work just as nicely in a Windows container.

agoallikmaa: There are actually some nice-to-haves in the Testcontainers implementation that I did not implement in the docker-java one as that would not have been worth it just for this one use case. The main one is that it makes sure the containers are cleaned up without relying on the stop/kill being called from Java code.

Also, Testcontainers would be the preferred way to do this, so ideally this docker-java custom implementation could be removed once they add support for Windows images.

iNikem (approved):

PR #92, opened 2021-01-25
Windows smoke tests for WildFly

agoallikmaa:
Creates Windows Docker images for the same set of WildFly version and JDK combinations as upstream OTel publishes for Linux. Adds smoke tests configurations for those images.

Original plan was to add WebSphere Liberty as well, but downloading its distributions requires logging in with an IBM account, so it would not currently be possible to build those images locally in Github Actions. Will have to revisit that when we have a way to use prebuilt proprietary images in the CI pipeline.


vovencij (approved):

iNikem:
> Original plan was to add WebSphere Liberty as well, but downloading its distributions requires logging in with an IBM account

But Otel does build Liberty images for linux. Or is logging required only for windows images?

iNikem (approved):

agoallikmaa:
> > Original plan was to add WebSphere Liberty as well, but downloading its distributions requires logging in with an IBM account
>
> But Otel does build Liberty images for linux. Or is logging required only for windows images?

As OTel uses OpenLiberty for Linux images, added Windows images of that.

PR #98, opened 2021-02-02
Update to otel-java-instrumentation 0.15.0

mateuszrzeszutek:


custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
```diff
@@ -0,0 +1,43 @@
...
+  @Override
+  public void configure(SdkTracerProviderBuilder tracerProvider) {
+    tracerProvider
+        .setTraceConfig(
+            TraceConfig.builder()
+                .setMaxNumberOfAttributes(Integer.MAX_VALUE)
+                .setMaxNumberOfEvents(Integer.MAX_VALUE)
+                .setMaxNumberOfLinks(1000) // this is the default value actually
+                .setMaxNumberOfAttributesPerEvent(Integer.MAX_VALUE)
+                .setMaxNumberOfAttributesPerLink(Integer.MAX_VALUE)
```
iNikem, line 37: Here old values were 256 with the comment "events and links create collections with provided sizes, so we shouldn't set them too high"

custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
```diff
@@ -0,0 +1,43 @@
...
+                .setMaxLengthOfAttributeValues(TraceConfig.UNLIMITED_ATTRIBUTE_LENGTH)
+                .build())
+        .setSampler(Sampler.alwaysOn())
+        .addSpanProcessor(new MiddlewareAttributeSpanProcessor());
```
iNikem, line 41: I think I would still prefer to have a separate customer for this in that same package. So that all related functionality lives in the same place.

custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
mateuszrzeszutek: That has changed some time ago, right now SDK does not create a new collection with expected size for events. It still does for links though.

mateuszrzeszutek:
There's a problem with the `ResourceAttributesToSystemProperties` class: after introducing auto-configuration `Resource.getDefault()` no longer contains entries from `otel.resource.attributes` property. The final `Resource` object is stored in a private field in `SdkTracerProvider`->`TracerSharedState`.
I think that probably the only it can be retrieved by unobfuscating the provider and getting field value via reflection - @jkwatson do you have any idea how to get it in a less hacky way?

jkwatson:
> There's a problem with the `ResourceAttributesToSystemProperties` class: after introducing auto-configuration `Resource.getDefault()` no longer contains entries from `otel.resource.attributes` property. The final `Resource` object is stored in a private field in `SdkTracerProvider`->`TracerSharedState`.
> I think that probably the only it can be retrieved by unobfuscating the provider and getting field value via reflection - @jkwatson do you have any idea how to get it in a less hacky way?

We need to make sure that we have auto-configuration in place, with SPIs properly loaded, and then always merge whatever we want with `Resource.getDefault()`, then we should get everything from `otel.resource.attributes` like normal.

custom/src/main/java/com/splunk/opentelemetry/resource/ResourceAttributesToSystemProperties.java
```diff
@@ -36,12 +42,43 @@ public void beforeByteBuddyAgent() {}
...
+    Attributes attributes = getResource().getAttributes();
     attributes.forEach(
         (k, v) -> {
           if (k.getType() == AttributeType.STRING) {
             System.setProperty("otel.resource." + k.getKey(), v.toString());
           }
         });
   }
+
+  private static Resource getResource() {
```
mateuszrzeszutek, line 54: Okay, I ended up writing this abomination to get the actual resource attributes used by tracer...

I'll create an issue in the SDK to do something about it.

mateuszrzeszutek: I don't have any great idea, the best I could think of right now is adding a `ResourceProvider` inside the auto-configuration module that constructs another `ConfigProperties` object and uses it to read the property. Anything else that came to my mind was too hacky to mention it 😅

mateuszrzeszutek:
There's one more issue that appears when using 0.15 that does not occur in 0.14 (even with the latest fake-backend & collector images): sometimes (randomly) 2 traces appear instead of just one, the additional one (the first one actually) contains just a `Response.sendError` span. I have no idea what might have caused these to appear -- the HttpServletResponse instrumentation is pretty old and was always enabled, so why haven't those additional traces appeared earlier? @vovencij @agoallikmaa @laurit do you have any idea what's causing this?

see: https://github.com/signalfx/splunk-otel-java/pull/98/checks?check_run_id=1815450854

custom/src/main/java/com/splunk/opentelemetry/resource/ResourceAttributesToSystemProperties.java
jkwatson: Thinking more on this...it would be good for the autoconfiguration to expose the auto-configured Resource. This is something that would be useful to have access to, outside of having it built into the SDK itself. 

Maybe a `getResource()` method directly on the OpenTelemetrySdkAutoConfiguration class? Create the issue. I think Anuraag might have an opinion.

vovencij: > it's unlikely we'll do much to support this use-case of turning resource attributes into system properties
Reading existing resource attributes seems like pretty general use case?

jkwatson: well, depends.. Do we want to expose them on the SDK classes? Maybe, but that's something we have to carefully consider, because once we expose them, we can't un-expose them.

breedx-splk:
> There's one more issue that appears when using 0.15 that does not occur in 0.14 (even with the latest fake-backend & collector images): sometimes (randomly) 2 traces appear instead of just one, the additional one (the first one actually) contains just a `Response.sendError` span. I have no idea what might have caused these to appear -- the HttpServletResponse instrumentation is pretty old and was always enabled, so why haven't those additional traces appeared earlier? @vovencij @agoallikmaa @laurit do you have any idea what's causing this?
> 
> see: https://github.com/signalfx/splunk-otel-java/pull/98/checks?check_run_id=1815450854

I created a PR into this branch: #99 

I think the description sums it up over there, but basically the test framework is complicated and sometimes, based on timing, it's possible to have the 2 spans for the single trace spread over 1 or 2 instances of `ExportTraceServiceRequest`.  I don't think that is a bug, but I do think it means that we can't use the size() to tell how many traces we actually had.

I ran with this change a bunch and it seems to fix it.

vovencij:
What about this suggestion by @jkwatson :

> We need to make sure that we have auto-configuration in place, with SPIs properly loaded, and then always merge whatever we want with `Resource.getDefault()`, then we should get everything from `otel.resource.attributes` like normal.

Can't this be done instead of reflection?

mateuszrzeszutek:
> What about this suggestion by @jkwatson :
> 
> > We need to make sure that we have auto-configuration in place, with SPIs properly loaded, and then always merge whatever we want with `Resource.getDefault()`, then we should get everything from `otel.resource.attributes` like normal.
> 
> Can't this be done instead of reflection?

It could be done this way, but it would require parsing `otel.resource.attributes` value though; not much better than the current solution. (I'm not strongly in favor of the current one, I can change it if you want)

iNikem (approved):

PR #99, opened 2021-02-02
Fix test flappy

breedx-splk:
The smoke test framework is rather complicated, and the `TraceInspector` and `ExportTraceServiceRequest` are a bit difficult to grok fully...but what I _did_ observe is that sometimes the data is returned to the test as a single `ExportTraceServiceRequest` which contains both spans, and sometimes the spans are split between two instances of `ExportTraceServiceRequest`.  The latter is the case that causes the test to fail, because we were using the `.size()` of the `TraceInspector` to indicate the number of traces....and this is a problem.

So this change that size() check to instead use the number of unique trace IDs to determine the span count.

This also includes a few other cleanups.


smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
```diff
@@ -132,24 +134,25 @@ protected void assertServerHandler(ExpectedServerAttributes serverAttributes)
...
     log.debug("Response for non-existing page: {}", response.body().string());
-    Assertions.assertEquals(
+    assertEquals(
         404,
         response.code(),
         "404 response code is expected from the app-server for a request to a non-existing page.");
     var traces = waitForTraces();
 
-    Assertions.assertEquals(1, traces.size(), "There is one trace from server handler");
+    var numberOfTraceIds = traces.getSpanStream().map(Span::getTraceId).distinct().count();
```
breedx-splk, line 143: ⭐ This is the important change.  Depending on how the data is reported, `traces.size()` can be 1 or 2, even though both spans are always for a single trace (same ID).

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
```diff
@@ -44,38 +46,38 @@ protected void assertWebAppTrace(ExpectedServerAttributes serverAttributes, Test
...
     Request request = new Request.Builder().get().url(url).build();
-    String responseBody = tryGetResponse(request);
+    String responseHeadersAndBody = tryGetResponse(request);
 
     TraceInspector traces = waitForTraces();
 
     Set<String> traceIds = traces.getTraceIds();
 
-    Assertions.assertEquals(1, traceIds.size(), "There is one trace");
+    assertEquals(1, traceIds.size(), "There is one trace");
```
iNikem, line 55: Should this be changed as well?

iNikem (approved):

smoke-tests/build.gradle
```diff
@@ -10,8 +10,8 @@ dependencies {
   testImplementation("com.fasterxml.jackson.core:jackson-databind:2.11.2")
   testImplementation("com.google.protobuf:protobuf-java-util:3.12.4")
   testImplementation("com.squareup.okhttp3:okhttp:3.12.12")
-  testImplementation("io.opentelemetry:opentelemetry-proto:0.10.0")
-  testImplementation("io.opentelemetry:opentelemetry-api:0.10.0")
+  testImplementation("io.opentelemetry:opentelemetry-proto:0.15.0")
```
mateuszrzeszutek, line 13: I don't really know if it's important, but the `fake-backend` app still uses otel-proto 0.10.0: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/smoke-tests/fake-backend/build.gradle#L28

mateuszrzeszutek (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
vovencij: `traceIds` here is `Set`, so it's size should be correct.

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
vovencij: There is `com.splunk.opentelemetry.TraceInspector#getTraceIds` which does the part up to `distinct`. Maybe we can have just one way of doing this?

vovencij (changes requested):

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
breedx-splk: I consider it to be a different use case.  Here, we are only concerned with the number of distinct IDs, but with `getTraceIds()` the trace ID values are actually used/compared.  I guess since you called it out here _and_ Nikita also got tripped up below, I'll combine them.

mateuszrzeszutek (approved):

smoke-tests/build.gradle
breedx-splk: This is a good catch.  Fortunately, it seems to work ok with the 0.15.0 on the testing side here.  🤷

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
```diff
@@ -50,9 +50,9 @@ protected void assertWebAppTrace(ExpectedServerAttributes serverAttributes, Test
 
     TraceInspector traces = waitForTraces();
 
-    Set<String> traceIds = traces.getTraceIds();
+    assertEquals(1, traces.countTraceIds(), "There is one trace");
 
-    assertEquals(1, traceIds.size(), "There is one trace");
+    Set<String> traceIds = traces.getTraceIds();
```
iNikem, line 55: Old version was faster :) only one stream iteration

smoke-tests/build.gradle
mateuszrzeszutek: As long as it works 🤷

smoke-tests/src/test/java/com/splunk/opentelemetry/AppServerTest.java
breedx-splk: 🙉

breedx-splk:
In the interest of helping #98 unblocked, I am going to merge this now.

PR #100, opened 2021-02-03
Netty 3.8 server timing

mateuszrzeszutek:


instrumentation/netty-3.8/src/main/java/com/splunk/opentelemetry/netty/v3_8/NettyChannelPipelineInstrumentation.java
```diff
@@ -0,0 +1,146 @@
...
+      try {
+        if (handler instanceof HttpServerCodec || handler instanceof HttpResponseEncoder) {
+          pipeline.addLast(
+              ServerTimingHandler.class.getName(), new ServerTimingHandler(contextStore));
+        }
+      } finally {
+        CallDepthThreadLocalMap.reset(ServerTimingHandler.class);
+      }
+    }
+  }
```
vovencij, line 145: Any reason why implementation is duplicated? AFAIU the only difference is advices is the position of `ChannelHandler` argument – 1 vs 2. Other than that logic seems to be the same. Maybe extract that in the abstract super-class?

mateuszrzeszutek: Hmm, I can extract some common parts to an additional helper class - abstract class probably wouldn't work that well, since advice classes are not "normal" classes.

vovencij (approved):

PR #111, opened 2021-02-07
Configure bots

flands:
* Enable auto-close of issues with no response from issue author
* Enable auto-close of stale issues
* Enable release notes to be automatically drafted

### DIFF
--- a/.github/no-response.yaml
+++ b/.github/no-response.yaml
@@ -1,0 +1,13 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 14
+# Label requiring a response
+responseRequiredLabel: more-info-needed
+# Comment to post when closing an Issue for lack of response. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed because there has been no response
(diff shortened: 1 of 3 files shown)


.github/no-response.yaml
```diff
@@ -0,0 +1,13 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 14
+# Label requiring a response
+responseRequiredLabel: more-info-needed
```
iNikem, line 6: We don't use such label at the moment. Should we start doing it in your opinion?

.github/stale.yaml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
```
iNikem, line 4: This puts an obligation on us to react/resolve all issues in 2 months. I am not sure we can do that.

.github/no-response.yaml
flands: that would be my suggestion -- I added the label to the project already

iNikem (approved):

PR #115, opened 2021-02-08
Fix README regarding setting service name

mateuszrzeszutek:
And add a smoke test that verifies that otel resource works correctly

Related to https://github.com/signalfx/splunk-otel-java/discussions/114

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Then the runtime parameters would be updated to:
 ```bash
 $ curl -L https://github.com/signalfx/splunk-otel-java/releases/latest/download/splunk-otel-javaagent-all.jar \
     -o splunk-otel-javaagent.jar
-$ java -javaagent:./splunk-otel-javaagent.jar -Dotel.jaeger.service.name=my-java-app \
+$ java -javaagent:./splunk-otel-javaagent.jar -Dotel.resource.attributes=service.name=my-java-app \
     -jar target/java-agent-example-1.0-SNAPSHOT-shaded.jar https://google.com
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -93,7 +96,6 @@ A simple wrapper for the Jaeger exporter of [opentelemetry-java](https://github.
 |-----------------------------------|-----------------------------------|---------------------------------------------------------------------------------------------------------------------|
 | otel.trace.exporter=jaeger-thrift-splunk | OTEL_TRACE_EXPORTER=jaeger-thrift-splunk              | Select the span exporter to use. `jaeger-thrift-splunk` is the default value.                                                                                          
 | otel.exporter.jaeger.endpoint     | OTEL_EXPORTER_JAEGER_ENDPOINT     | The Jaeger endpoint to connect to. Default is `http://localhost:9080/v1/trace`.
-| otel.exporter.jaeger.service.name | OTEL_EXPORTER_JAEGER_SERVICE_NAME | The service name of this JVM instance. Default is `unknown_service:java`, but any `service.name` provided via other means will supercede this value.
```
iNikem, line 96: Otel java autoconfiguration will not support this property?

mateuszrzeszutek: It doesn't support it right now, and seeing as in the current snapshot version there's no way to set service name other than resource attributes, it probably never will.

jkwatson: we never added this to auto-configuration because it hasn't been in the spec since we created the auto-configuration module. That was always a "java-only" special that was waiting to be removed until the spec finally sorted out the required service name in the Resource.

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
```diff
@@ -33,17 +33,11 @@
 
   static final String SIGNALFX_AUTH_TOKEN = "signalfx.auth.token";
   public static final String OTEL_EXPORTER_JAEGER_ENDPOINT = "otel.exporter.jaeger.endpoint";
-  static final String OTEL_EXPORTER_JAEGER_SERVICE_NAME = "otel.exporter.jaeger.service.name";
 
   @Override
   public SpanExporter createExporter(ConfigProperties config) {
     JaegerThriftSpanExporterBuilder builder = JaegerThriftSpanExporter.builder();
 
-    String serviceName = config.getString(OTEL_EXPORTER_JAEGER_SERVICE_NAME);
```
jkwatson, line 42: if we wanted to support this for a release, we could probably add this value to the resource that we configure into the SDK. I don't know how important that is, but we can probably make it work if we have to.

mateuszrzeszutek: Yep, an additional `ResourceProvider` would work. The question is, do we really need that? We've already had lots of breaking changes in configuration in this release.

README.md
iNikem: Let's not add anything not in spec. Or at least as little as possible :)

iNikem (approved):

PR #116, opened 2021-02-10
Update to upstream 0.16.0

iNikem:


custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
```diff
@@ -19,22 +19,23 @@
...
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 
 @AutoService(SdkTracerProviderConfigurer.class)
 public class SplunkTracerProviderConfigurer implements SdkTracerProviderConfigurer {
   @Override
   public void configure(SdkTracerProviderBuilder tracerProvider) {
     tracerProvider
-        .setTraceConfig(
-            TraceConfig.builder()
+        .setSampler(Sampler.alwaysOn())
```
mateuszrzeszutek, line 30: Sampler is already being set in line 40

mateuszrzeszutek (approved):

custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
iNikem: Oops. fixed

mateuszrzeszutek (approved):

custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
```diff
@@ -19,23 +19,23 @@
...
+        .setSpanLimits(
+            SpanLimits.builder()
                 .setMaxNumberOfAttributes(Integer.MAX_VALUE)
                 .setMaxNumberOfEvents(Integer.MAX_VALUE)
-                .setMaxNumberOfLinks(1000) // this is the default value actually
+                .setMaxNumberOfLinks(1000)
```
jkwatson, line 35: 👍 since it seems like this default might change.

jkwatson (approved):

PR #122, opened 2021-02-17
Add Windows TomEE image and test

agoallikmaa:
Adds Windows Docker images for the same versions of TomEE that already exist for Linux and adds them as test cases for smoke test.


iNikem (approved):

mateuszrzeszutek (approved):

iNikem:
I will have to think how to make our private images work on GHA...

PR #125, opened 2021-02-23
Rename config properties so that they're similarly named as in other Splunk distros

mateuszrzeszutek:
e.g. splunk-otel-python, splunk-otel-js

README will be updated after release

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
@@ -31,6 +31,8 @@ import org.slf4j.LoggerFactory;
 public class JaegerThriftSpanExporterFactory implements ConfigurableSpanExporterProvider {
   private static final Logger log = LoggerFactory.getLogger(JaegerThriftSpanExporterFactory.class);
 
+  @Deprecated
+  static final String SPLUNK_ACCESS_TOKEN_OLD = "signalfx.auth.token";
   static final String SPLUNK_ACCESS_TOKEN = "splk.access.token";
   public static final String OTEL_EXPORTER_JAEGER_ENDPOINT = "otel.exporter.jaeger.endpoint";
 
@@ -40,6 +42,9 @@ public class JaegerThriftSpanExporterFactory implements ConfigurableSpanExporter
 
     String endpoint = config.getString(OTEL_EXPORTER_JAEGER_ENDPOINT);
     String token = config.getString(SPLUNK_ACCESS_TOKEN);
+    if (token == null) {
+      token = config.getString(SPLUNK_ACCESS_TOKEN_OLD);
+    }
     if (token != null && !token.isEmpty()) {
       log.debug("Using authenticated jaeger-thrift exporter");
       builder.setThriftSender(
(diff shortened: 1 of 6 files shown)


jkwatson:
Since this will break existing users, should we support both for at least a little while?

mateuszrzeszutek:
> Since this will break existing users, should we support both for at least a little while?

Good point - I'll do that for the auth token, since it seems to be the only one that's actually used by our users. Thanks!

jkwatson:
> > Since this will break existing users, should we support both for at least a little while?
> 
> Good point - I'll do that for the auth token, since it seems to be the only one that's actually used by our users. Thanks!

It might be interesting to log a warning about the deprecated name at startup time if we see it being used.

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
```diff
@@ -44,6 +43,12 @@ public SpanExporter createExporter(ConfigProperties config) {
     String token = config.getString(SPLUNK_ACCESS_TOKEN);
     if (token == null) {
       token = config.getString(SPLUNK_ACCESS_TOKEN_OLD);
+      if (token != null) {
+        log.warn(
+            "Deprecated property '{}' was set; consider using '{}' instead",
```
jkwatson, line 48:
```suggestion
            "Deprecated property '{}' was set; please use '{}' instead",
```

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
```diff
@@ -31,15 +31,25 @@
 public class JaegerThriftSpanExporterFactory implements ConfigurableSpanExporterProvider {
   private static final Logger log = LoggerFactory.getLogger(JaegerThriftSpanExporterFactory.class);
 
-  static final String SIGNALFX_AUTH_TOKEN = "signalfx.auth.token";
+  @Deprecated static final String SPLUNK_ACCESS_TOKEN_OLD = "signalfx.auth.token";
+  static final String SPLUNK_ACCESS_TOKEN = "splk.access.token";
```
iNikem, line 35: Why the heck did we agree on `splk` and not `splunk` ???

instrumentation/netty-3.8/build.gradle
```diff
@@ -15,5 +15,5 @@ dependencies {
 }
 
 tasks.withType(Test) {
-    jvmArgs '-Dsplunk.context.server-timing.enabled=true'
+    jvmArgs '-Dsplk.context.server-timing.enabled=true'
```
iNikem, line 18: 😭 Why??

custom/src/main/java/com/splunk/opentelemetry/jaeger/JaegerThriftSpanExporterFactory.java
mateuszrzeszutek: I don't know... there's even a conversation on slack about using `SPLK_` and we both left 👍  there 😭 🤣   nobody knows why now 🤷‍♂️

instrumentation/netty-3.8/build.gradle
breedx-splk: I wasn't around for this, but it _is_ the ticker symbol and it _is_ shorter.  😁 It's just unpronounceable.... 🤮

iNikem:
We have agreed on `SPLUNK_` prefix.

iNikem (approved):

PR #127, opened 2021-02-25
OpenJ9 smoke test Docker images

agoallikmaa:
* Adds OpenJ9 images for Windows images and JBoss images.
* The logic to define images to create in `matrix/build.gradle` has been changed to be as close as possible to the logic in OpenTelemetry - only difference currently is that this one has logic to make Windows images, add pull tasks, and can be called for multiple sets of targets as it is wrapped in a function with targets as a parameter.
* Added a more concise way to list test configurations in smoke tests now to remove duplication and to merge adding multiple JDKs and VMs of the same version into one line.
* Updated `junit-jupiter` and `mockito` dependency versions to the latest which also makes them all depend on exactly the same jupiter version. This was done because for some reason after `jcenter` removal, without redownloading dependencies it tried to use a mix of `5.7.0` and `5.6.2` packages which caused an `AbstractMethodError` during build. Did not investigate the root cause further as making sure everything requires the same version was an easy fix.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,9 @@ boolean snapshot = getProperty("snapshot") == 'yes'
 def versions = [
     'opentelemetry'         : snapshot ? "0.17.0-SNAPSHOT"       : "0.16.0",
     'opentelemetryAlpha'    : snapshot ? "0.17.0-alpha-SNAPSHOT" : "0.16.0-alpha",
-    'opentelemetryJavaagent': snapshot ? "0.17.0-SNAPSHOT"       : '0.16.1'
+    'opentelemetryJavaagent': snapshot ? "0.17.0-SNAPSHOT"       : '0.16.1',
+    'mockito'               : '3.8.0',
+    'jupiter'               : '5.7.1'
 ]
 
 subprojects {
@@ -44,12 +46,12 @@ subprojects {
   }
 
   dependencies {
-    testImplementation("org.mockito:mockito-core:3.8.0")
-    testImplementation("org.mockito:mockito-junit-jupiter:3.8.0")
-    testImplementation("org.junit.jupiter:junit-jupiter-api:5.7.1")
-    testImplementation("org.junit.jupiter:junit-jupiter-params:5.7.1")
-    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.7.1")
-    testImplementation("org.junit.jupiter:junit-jupiter-params:5.7.1")
+    testImplementation("org.mockito:mockito-core:${versions.mockito}")
+    testImplementation("org.mockito:mockito-junit-jupiter:${versions.mockito}")
+    testImplementation("org.junit.jupiter:junit-jupiter-api:${versions.jupiter}")
+    testImplementation("org.junit.jupiter:junit-jupiter-params:${versions.jupiter}")
+    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:${versions.jupiter}")
+    testImplementation("org.junit.jupiter:junit-jupiter-params:${versions.jupiter}")
   }
 
   tasks {
--- a/build.gradle
+++ b/build.gradle
@@ -44,12 +44,12 @@ subprojects {
   }
 
   dependencies {
-    testImplementation("org.mockito:mockito-core:3.5.15")
-    testImplementation("org.mockito:mockito-junit-jupiter:3.5.15")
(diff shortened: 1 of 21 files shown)


matrix/build.gradle
```diff
@@ -25,14 +25,14 @@ def buildProprietaryTestImagesTask = tasks.create("buildProprietaryTestImages")
...
 
 def proprietaryTargets = [
-    "weblogic" : [
-       "12.1.3" : ["developer"],
-       "12.2.1.4" : ["developer"],
-       "14.1.1.0" : ["developer-8", "developer-11"]
+    "weblogic": [
+        [version: ["12.1.3", "12.2.1.4"], vm: ["hotspot"], jdk: ["8"], args: [tagSuffix: "developer"]],
+        [version: ["14.1.1.0"], vm: ["hotspot"], jdk: ["8"], args: [tagSuffix: "developer-8"]],
+        [version: ["14.1.1.0"], vm: ["hotspot"], jdk: ["11"], args: [tagSuffix: "developer-11"]]
```
iNikem, line 31: Can these 2 lines be one? With an array of `jdk`?

iNikem (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/TomcatSmokeTest.java
```diff
@@ -37,42 +33,12 @@
...
-        arguments(
-            proprietaryWindowsImage("splunk-tomcat:8.5.60-jdk11-windows"),
-            TOMCAT8_SERVER_ATTRIBUTES),
-        arguments(
-            proprietaryWindowsImage("splunk-tomcat:9.0.40-jdk8-windows"),
-            TOMCAT9_SERVER_ATTRIBUTES),
-        arguments(
-            proprietaryWindowsImage("splunk-tomcat:9.0.40-jdk11-windows"),
-            TOMCAT9_SERVER_ATTRIBUTES));
+    return configurations("tomcat").otelLinux("7.0.107", TOMCAT7_SERVER_ATTRIBUTES, VMS_ALL, "8")
```
mateuszrzeszutek, line 36: Nit formatting:
```suggestion
    return configurations("tomcat")
        .otelLinux("7.0.107", TOMCAT7_SERVER_ATTRIBUTES, VMS_ALL, "8")
```

mateuszrzeszutek (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/TomcatSmokeTest.java
agoallikmaa: This (and some other odd formatting here) was done by spotless. Before that I had them all on separate lines.

matrix/build.gradle
agoallikmaa: The reason is that in some cases the `tagSuffix` needs to contain the JDK version and some cases not (the 12. versions). Therefore it cannot be just appended in the Dockerfile and needs to be manually provided, which means the arguments must be different for version 14. where the suffix contains JDK version.

smoke-tests/src/test/java/com/splunk/opentelemetry/TomcatSmokeTest.java
mateuszrzeszutek: Nothing we can do here then 🤷‍♂️

build.gradle
```diff
@@ -44,12 +44,12 @@ subprojects {
...
-    testImplementation("org.junit.jupiter:junit-jupiter-api:5.6.2")
-    testImplementation("org.junit.jupiter:junit-jupiter-params:5.6.2")
-    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.6.2")
-    testImplementation("org.junit.jupiter:junit-jupiter-params:5.6.2")
+    testImplementation("org.mockito:mockito-core:3.8.0")
+    testImplementation("org.mockito:mockito-junit-jupiter:3.8.0")
+    testImplementation("org.junit.jupiter:junit-jupiter-api:5.7.1")
+    testImplementation("org.junit.jupiter:junit-jupiter-params:5.7.1")
+    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.7.1")
+    testImplementation("org.junit.jupiter:junit-jupiter-params:5.7.1")
```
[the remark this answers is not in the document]
agoallikmaa: Added.

mateuszrzeszutek (approved):

PR #130, opened 2021-03-04
Update migration.md with jaeger exporter vars

gtrog:
Hi, I noticed that the migration docs still referenced the old zipkin exporter variables, so, wanted to update this to accurately reflect how to currently migrate to the OTEL agent.

### DIFF
--- a/migration.md
+++ b/migration.md
@@ -137,13 +137,3 @@ annotation instead of the `@Trace` annotation. For more information, see
 [Configure a WithSpan annotation](https://github.com/open-telemetry/opentelemetry-java-instrumentation#configure-a-withspan-annotation).
 
 The `@TraceSetting` annotation to allow an exception isn't supported.
-
-### SignalFX Agent Configuration changes
-
-The previous Java Agent used `zipkin` as the trace exporter format.  In order to start using the OpenTelemetry Java Instrumentation agent, you will need to add the following to the SignalFX Smart Agent configuration:
-
-```
-writer:
-  traceExportFormat: sapm
-```
-
--- a/migration.md
+++ b/migration.md
@@ -137,3 +137,13 @@ annotation instead of the `@Trace` annotation. For more information, see
 [Configure a WithSpan annotation](https://github.com/open-telemetry/opentelemetry-java-instrumentation#configure-a-withspan-annotation).
 
 The `@TraceSetting` annotation to allow an exception isn't supported.
+
+### SignalFX Agent Configuration changes
+
+The previous Java Agent used `zipkin` as the trace exporter format.  In order to start using the OpenTelemetry Java Instrumentation agent, you will need to add the following to the SignalFX Smart Agent configuration:
+
+```
+writer:
+  traceExportFormat: sapm
+```
+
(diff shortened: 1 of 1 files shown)


jkwatson (approved):
Thanks!

gtrog:
I added another bit of documentation of an issue I ran into where I didn't configure `traceExportFormat` to `sapm` and encountered issues when the SignalFX Agent would try to ship the spans.  Once I added this config, everything seemed to work.  Let me know if this is correct/accurate.

migration.md
```diff
@@ -137,3 +137,13 @@ annotation instead of the `@Trace` annotation. For more information, see
...
 
 The `@TraceSetting` annotation to allow an exception isn't supported.
+
+### SignalFX Agent Configuration changes
+
+The previous Java Agent used `zipkin` as the trace exporter format.  In order to start using the OpenTelemetry Java Instrumentation agent, you will need to add the following to the SignalFX Smart Agent configuration:
+
+```
+writer:
+  traceExportFormat: sapm
```
jkwatson, line 147: this confuses me. the new java agent uses jaeger by default for export. why would you have to change the export format in the smart agent?

gtrog: According to the docs, `traceExportFormat` defaults to `zipkin`.  Without changing it, I was getting errors like this:
```
time="2021-03-04T19:01:55Z" level=error msg="Error shipping spans to SignalFx" error="{\"valid\":0,\"invalid\":{\"invalidSpanName\":[\"6fdf7bf5ff746991\",\"0d03beb990ebcbf5\",\"5a1da0ce77d77404\"]}}"
```

jkwatson: But why would this be different with a change in the javaagent you were using? That seems unrelated to which input format was being used.

gtrog: This is just something I encountered in migrating from one java agent to the other:

We're currently using the old µAPM agent on-prem, and there, the SignalFx doesn't have `traceExportFormat` explicitly configured, so, it's defaulting to `zipkin`, and this configuration overall seems to work.

I'm working on a second environment in Kubernetes and figured since it's a greenfield project, I figured I'd try using the recommended OTEL java agent instead.  Once I had everything seemingly configured correctly and similar to what we have on prem, I kept seeing the errors above.  After a few hours of trial and error and trying to understand what's going on, I realized that the old apm java agent was using zipkin internally and the new one is using jaeger.  I figured `traceExportFormat` may have been something to do with it, explicitly set it to `sapm` in our Kubernetes environment, at which point I was able to start seeing APM traces come in to SignalFx.  Just seemed related to migration, that's why I mentioned it here.  I don't fully understand why this had to be explicitly set when it didn't before or what exactly the setting does.  I can remove the documentation if it's unrelated.

gtrog: I'll remove this documentation for now, this can be revisited later if it seems related, just thought it was something noteworthy.

iNikem (changes requested):
Thank you for your contribution, @gtrog. Splunk requires for all external contribution to sign this [Contributor License Agreement ](https://www.splunk.com/en_us/form/contributions.html) before we can accept PRs from them. Can you do that?

PR #132, opened 2021-03-09
Remove jfrog oss-snapshot-local repository

laurit:
According to @iNikem it shouldn't be needed anymore. For some reason having it before `mavenCentral` slows down build by quite a bit.


jkwatson (approved):
👍

PR #135, opened 2021-03-12
Export JVM metrics from the javaagent using micrometer

mateuszrzeszutek:
README changes will be added in a separate PR


custom/src/main/java/com/splunk/opentelemetry/micrometer/OtelNamingConvention.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * <ol>
+ *   <li>To match what OTel metrics spec says about runtime metrics: and pretty much the only thing
+ *       that's specified there is that runtime metrics should start with {@code
+ *       runtime.{environment}.}
+ *   <li>To avoid conflicts with the {@code opentelemetry-java-contrib/jmx-metrics} tool that starts
+ *       all metrics with {@code jvm.}
+ * </ol>
+ */
+class OtelNamingConvention implements NamingConvention {
```
jkwatson, line 34: I don't see any unit tests for this class. I think that would be good to have.

jkwatson (approved):
This look really great. Just would like some unit tests for `OtelNamingConvention`, but that could come in a separate PR if you would prefer.

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/GlobalMetricsTags.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class holds the list of tags that should be added to all meters registered by the javaagent.
+ * It needs to be loaded by the bootstrap classloader to be accessible inside instrumentations.
+ */
+public final class GlobalMetricsTags {
```
breedx-splk, line 31: I don't want this thought to get in the way of this PR...but I would love for us to also consider ways of trying to avoid this static/global pattern down the road...even if it means passing the tags into all instrumentations that need it.  Or something.  I dunno....I just worry about this being hard to maintain.

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
```diff
@@ -0,0 +1,73 @@
...
+    if (environment != null) {
+      globalTags.add(Tag.of("deployment.environment", environment));
+    }
+    if (service != null) {
+      globalTags.add(Tag.of("service", service));
+    }
+    if (runtime != null) {
+      globalTags.add(Tag.of("runtime", runtime));
+    }
+    if (pid != null) {
```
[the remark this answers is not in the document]
breedx-splk: Would be good to have unit test coverage on this stuff too.

custom/src/main/java/com/splunk/opentelemetry/micrometer/OtelNamingConvention.java
```diff
@@ -0,0 +1,56 @@
...
+  private final NamingConvention signalfxNamingConvention;
+
+  OtelNamingConvention(NamingConvention signalfxNamingConvention) {
+    this.signalfxNamingConvention = signalfxNamingConvention;
+  }
+
+  @Override
+  public String name(String name, Meter.Type type, String baseUnit) {
+    String metricName = signalfxNamingConvention.name(name, type, baseUnit);
+    return metricName.startsWith("jvm.") ? "runtime." + metricName : metricName;
```
breedx-splk, line 44: Unit tests please!

custom/src/main/java/com/splunk/opentelemetry/micrometer/SplunkMetricsConfig.java
```diff
@@ -0,0 +1,67 @@
...
+import io.micrometer.signalfx.SignalFxConfig;
+import io.opentelemetry.instrumentation.api.config.Config;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+import java.time.Duration;
+
+class SplunkMetricsConfig implements SignalFxConfig {
+  @Override
+  public boolean enabled() {
+    return Config.get().getBooleanProperty("splunk.metrics.enabled", true);
```
breedx-splk, line 28: Let's limit the number of global/static calls and create a config field on this class that is passed into the constructor.  Same with `Resource` (below).

custom/src/main/java/com/splunk/opentelemetry/micrometer/SplunkMetricsConfig.java
```diff
@@ -0,0 +1,67 @@
...
+    // non-empty token MUST be provided; we can just send anything because collector/SmartAgent will
+    // use the real one
+    return Config.get().getProperty("splunk.access.token", "no-token");
+  }
+
+  @Override
+  public String uri() {
+    // right now the default value points to SmartAgent endpoint
+    return Config.get()
+        .getProperty("splunk.metrics.endpoint", "http://localhost:9080/v2/datapoint");
```
breedx-splk, line 42: Promote the value to a constant to raise visibility/findability?

custom/src/main/java/com/splunk/opentelemetry/micrometer/SplunkMetricsConfig.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  public String source() {
+    return Resource.getDefault().getAttributes().get(ResourceAttributes.SERVICE_NAME);
+  }
+
+  @Override
+  public Duration step() {
+    long stepMillis =
+        Long.parseLong(Config.get().getProperty("splunk.metrics.export.interval", "30000"));
+    return Duration.ofMillis(stepMillis);
```
breedx-splk, line 54: Would like to see unit test coverage on this.

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/GlobalMetricsTags.java
breedx-splk: I also realize that passing it into an AutoService may be difficult/impossible...but still.  😁

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/jvmmetrics/JvmMetricsInstaller.java
```diff
@@ -0,0 +1,50 @@
...
+            singleton("jvm-metrics"),
+            Config.get().getBooleanProperty("splunk.metrics.enabled", true))) {
+      return;
+    }
+
+    List<Tag> tags = GlobalMetricsTags.get();
+    new ClassLoaderMetrics(tags).bindTo(Metrics.globalRegistry);
+    new JvmGcMetrics(tags).bindTo(Metrics.globalRegistry);
+    new JvmMemoryMetrics(tags).bindTo(Metrics.globalRegistry);
+    new JvmThreadMetrics(tags).bindTo(Metrics.globalRegistry);
```
breedx-splk, line 48: 👍

breedx-splk (changes requested):
This is awesome stuff!  Had a few ideas/comments, bu t looks good overall to me.

custom/src/main/java/com/splunk/opentelemetry/micrometer/OtelNamingConvention.java
mateuszrzeszutek: Definitely -- I must have overlooked this class, I was so focused on making gradle work 😄

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
```diff
@@ -0,0 +1,73 @@
...
+@AutoService(ComponentInstaller.class)
+public class MicrometerInstaller implements ComponentInstaller {
+  @Override
+  public void beforeByteBuddyAgent() {
+    List<Tag> tags = getGlobalTags();
+    GlobalMetricsTags.set(tags);
+    Metrics.addRegistry(createSplunkMeterRegistry());
+  }
+
+  private static List<Tag> getGlobalTags() {
```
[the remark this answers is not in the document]
mateuszrzeszutek: Unfortunately `GlobalMetricsTags` is in the bootstrap classloader and has no access to the otel SDK - I had to do it in this class (which is in the agent classloader and thus can see SDK classes)

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/GlobalMetricsTags.java
mateuszrzeszutek: I first looked into setting global tags in the meter registry itself, which would be a much cleaner solution -- no need to remember about passing global tags to your new meter, the registry would it automagically. Unfortunately it required digging really deep into `MeterRegistry` internals and overriding >10 different methods to make this work 🥶  

And the fact that those tags need to be accessible both in the agent and application classloaders makes "fancy" (non-global variable 😂 ) implementation much harder...

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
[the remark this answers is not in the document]
breedx-splk: Awww shucks! I didn't consider that.  I guess this is a form of bridging....

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/GlobalMetricsTags.java
breedx-splk: Yeah, of course.  Seems reasonable for now then.

custom/src/main/java/com/splunk/opentelemetry/micrometer/SplunkMetricsConfig.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  public String source() {
+    return Resource.getDefault().getAttributes().get(ResourceAttributes.SERVICE_NAME);
+  }
+
+  @Override
+  public Duration step() {
+    long stepMillis =
+        Long.parseLong(Config.get().getProperty("splunk.metrics.export.interval", "30000"));
```
iNikem, line 53: Is this method called often or just once? In the former case we should extract this to a constant, I think.

gradle/instrumentation.gradle
```diff
@@ -37,17 +36,20 @@ shadowJar {
...
 
+evaluationDependsOn(":testing:agent-for-testing")
+
 tasks.withType(Test).configureEach {
+    dependsOn shadowJar
+    dependsOn ":testing:agent-for-testing:shadowJar"
+
     jvmArgs "-Dotel.javaagent.debug=true"
-    jvmArgs "-javaagent:${configurations.testAgent.files.first().absolutePath}"
+    jvmArgs "-javaagent:${project(":testing:agent-for-testing").tasks.shadowJar.archiveFile.get().asFile.absolutePath}"
```
iNikem, line 46: I think you should change it back :) https://docs.gradle.org/current/userguide/cross_project_publications.html

instrumentation/jvm-metrics/src/test/java/com/splunk/opentelemetry/jvmmetrics/JvmMetricsTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(AgentInstrumentationExtension.class)
+class JvmMetricsTest {
+  @Test
+  void shouldRegisterJvmMeters() {
+    var meterNames = TestMetricsAccess.getMeterNames();
+
+    // classloader metrics
+    assertTrue(meterNames.contains("jvm.classes.loaded"));
```
iNikem, line 33: Looking at `OtelNamingConvention`, I would expect seeing `runtime` somewhere here. Why not?

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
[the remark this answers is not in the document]
kubawach: I think that this (meaning - creating a list of tags) could even be turned into a builder attached to the GlobalMetricsTags class. This suggestion is a bit in line with @breedx-splk previous idea of moving most of the code there which could not be implemented due to class loading issue.

custom/src/main/java/com/splunk/opentelemetry/micrometer/OtelNamingConvention.java
```diff
@@ -0,0 +1,56 @@
...
+ * <ol>
+ *   <li>To match what OTel metrics spec says about runtime metrics: and pretty much the only thing
+ *       that's specified there is that runtime metrics should start with {@code
+ *       runtime.{environment}.}
+ *   <li>To avoid conflicts with the {@code opentelemetry-java-contrib/jmx-metrics} tool that starts
+ *       all metrics with {@code jvm.}
+ * </ol>
+ */
+class OtelNamingConvention implements NamingConvention {
+  private final NamingConvention signalfxNamingConvention;
```
kubawach, line 35: I'd cut "signalfx" prefix here (does not seem to be now-defunct-signalfx related).

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
```diff
@@ -0,0 +1,73 @@
...
+    if (pid != null) {
+      globalTags.add(Tag.of("process.pid", pid.toString()));
+    }
+    return globalTags;
+  }
+
+  private static SignalFxMeterRegistry createSplunkMeterRegistry() {
+    SignalFxMeterRegistry signalFxRegistry =
+        new SignalFxMeterRegistry(new SplunkMetricsConfig(), Clock.SYSTEM);
+    NamingConvention signalFxNamingConvention = signalFxRegistry.config().namingConvention();
```
kubawach, line 69: Just wondering - is it possible to get a null naming conv here?

kubawach (approved):
Some minor comments.

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
mateuszrzeszutek: No -- `SignalFxMeterRegistry` always sets its own convention.

custom/build.gradle
```diff
@@ -10,15 +10,23 @@ def relocatePackages = ext.relocatePackages
 
 dependencies {
   compileOnly(project(":bootstrap"))
-  implementation("io.opentelemetry:opentelemetry-sdk:${versions["opentelemetry"]}")
+  implementation("io.opentelemetry:opentelemetry-sdk:${versions.opentelemetry}")
   implementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:${versions.opentelemetryAlpha}")
+  implementation("io.opentelemetry:opentelemetry-semconv:${versions.opentelemetryAlpha}")
   implementation("io.opentelemetry:opentelemetry-exporter-jaeger-thrift:${versions.opentelemetry}")
+  implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:${versions.opentelemetryJavaagentAlpha}")
   implementation("io.opentelemetry.javaagent:opentelemetry-javaagent-spi:${versions.opentelemetryJavaagentAlpha}")
```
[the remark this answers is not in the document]
mateuszrzeszutek: 🤔 
Most of them, probably -- I'll change that.

custom/src/main/java/com/splunk/opentelemetry/micrometer/SplunkMetricsConfig.java
mateuszrzeszutek: Good catch - micrometer does not cache this at all (despite parsing being present in the base `PushRegistryConfig` class) and calls it every time a meter is created.

instrumentation/jvm-metrics/src/test/java/com/splunk/opentelemetry/jvmmetrics/JvmMetricsTest.java
mateuszrzeszutek: Because `OtelNamingConvention` is not actually present in the test javaagent, we don't include `project(":custom")` there. In another PR I'd like to add metrics verification to smoke tests (once I figure out how...) so the full name will be checked there.

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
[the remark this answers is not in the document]
mateuszrzeszutek: I decided to keep the `GlobalTagsBuilder` class in the `custom` module so that it can use the OTel Resource API.

gradle/instrumentation.gradle
mateuszrzeszutek: I think I fixed it 😅

mateuszrzeszutek:
Seems like the tests have passed (at least the linux ones...) -- @iNikem @kubawach @breedx-splk @jkwatson I've addressed all your comments, please re-review this PR 🙏

custom/src/test/java/com/splunk/opentelemetry/micrometer/OtelNamingConventionTest.java
```diff
@@ -0,0 +1,68 @@
...
+@ExtendWith(MockitoExtension.class)
+class OtelNamingConventionTest {
+  @Mock NamingConvention namingConventionMock;
+
+  NamingConvention otelNamingConvention;
+
+  @BeforeEach
+  void setUp() {
+    otelNamingConvention = new OtelNamingConvention(namingConventionMock);
+  }
```
jkwatson, line 39: Putting the class under test here I think is less readable than putting it in each of the test cases. and, it would be fewer lines of code to just put this in each of the test cases, as well.

custom/src/test/java/com/splunk/opentelemetry/micrometer/GlobalTagsBuilderTest.java
```diff
@@ -0,0 +1,66 @@
...
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+import org.junit.jupiter.api.Test;
+
+class GlobalTagsBuilderTest {
+  @Test
+  void shouldBuildEmptyTagsList() {
+    var resource = Resource.create(Attributes.empty());
```
jkwatson, line 32: I think you lost a `// given` here.

custom/src/test/java/com/splunk/opentelemetry/micrometer/GlobalTagsBuilderTest.java
```diff
@@ -0,0 +1,67 @@
...
+                "prod",
+                ResourceAttributes.SERVICE_NAME,
+                "my-service",
+                ResourceAttributes.PROCESS_RUNTIME_NAME,
+                "OpenJDK Runtime Environment",
+                ResourceAttributes.PROCESS_PID,
+                12345L));
+
+    // when
+    var tags = new GlobalTagsBuilder(resource).build();
```
jkwatson, line 58: java 11 in tests.. is there gradle magic that makes sure our code doesn't start using java 11 APIs that would break users on java 8?

breedx-splk (approved):

PR #144, opened 2021-03-29
Add metrics for JDBC connection pool: Apache Commons DBCP2

mateuszrzeszutek:
* Micrometer-core actually contains a ready-made [instrumentation for commons-pool2](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/commonspool2/CommonsObjectPool2Metrics.java) - the object pool used by dbcp2. I tried using it, but unfortunately it turned out to be not so useful - it relies on `BasicDataSource#jmxName` being set (`BasicDataSource` adds itself to the mbean server when you set the JMX name), unfortunately Spring completely ignores that property and registers the data source MXBean by itself. This results in the underlying object pool (and `PooledConnection`) MXBeans not being registered, because they're not spring beans - just objects created inside `BasicDataSource`. Because of that I had to ignore the commons-pool2 instrumentation and implement metrics just for dbcp2.
* I've decided to use dbcp types (like `BasicDataSourceMXBean`) in this instrumentation - using plain JMX without compileOnly dependency is possible, but it's considerably less readable - you can see the example in the micrometer `CommonsObjectPool2Metrics` class.


instrumentation/commons-dbcp2/src/test/java/com/splunk/opentelemetry/commonsdbcp2/CommonsDbcp2InstrumentationTest.java
```diff
@@ -0,0 +1,102 @@
...
+    connectionPool.getConnection();
+
+    // then
+    await()
+        .atMost(20, TimeUnit.SECONDS)
+        .untilAsserted(
+            () ->
+                assertThat(TestMetricsAccess.getMeterNames())
+                    .containsExactlyInAnyOrder(
+                        "db.pool.connections.active",
```
mateuszrzeszutek, line 69: Meter tags/units are not being verified yet, I'm planning to implement that in the next PR

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/DataSourceMetrics.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.commons.dbcp2.BasicDataSourceMXBean;
+
+public final class DataSourceMetrics {
+  public static void registerMetrics(BasicDataSourceMXBean dataSource, ObjectName objectName) {
+    List<Tag> tags = getTags(objectName);
+
+    gauge("db.pool.connections.active", tags, dataSource::getNumActive);
+    gauge("db.pool.connections.idle", tags, dataSource::getNumIdle);
+    gauge("db.pool.connections.min", tags, dataSource::getMinIdle);
+    gauge("db.pool.connections.max", tags, dataSource::getMaxTotal);
```
breedx-splk, line 36: Do we think it might be confusing here to be mixing "min idle" with "max total"?  I think users might not realize the disparity.  Is the goal to track the min and max size of the pool itself?  Maybe we want to consider tracking both "max idle" along with "max total", but give them different names?  So it would end up being something like:

* `db.pool.connections.active`
* `db.pool.connections.idle.num`
* `db.pool.connections.idle.min`
* `db.pool.connections.idle.max`
* `db.pool.connections.capacity.max`

breedx-splk (changes requested):
Looks good overall, but I had a question about the metric names and how they mapped over from the pool size info.  👍

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/DataSourceMetrics.java
mateuszrzeszutek: > Is the goal to track the min and max size of the pool itself?

Yes, exactly.

Metrics and their naming was inspired by HikariCP built-in metrics instrumentation ([here](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/metrics/micrometer/MicrometerMetricsTracker.java) and [here](https://github.com/brettwooldridge/HikariCP/blob/f9b2c3d372b3f28911a872191ad62953c84bafa4/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L661)). I tried to measure things that every connection pool exposes (still, it's just a guess) and Hikari does not expose idle.max in its metrics tracker API.

breedx-splk: But don't you think it's confusing to have min come from idle and max come from total?

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/DataSourceMetrics.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private static List<Tag> getTags(ObjectName objectName) {
+    // use the "name" property if available: Spring sets it to the bean name
+    String name = objectName.getKeyProperty("name");
+    // if its unavailable just use the whole mbean name
+    if (name == null) {
+      name = objectName.toString();
+    }
+    return GlobalMetricsTags.concat(Tag.of("type", "dbcp2"), Tag.of("name", name));
```
jkwatson, line 46: maybe `pool.type` and `pool.name` ?

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/DataSourceMetrics.java
mateuszrzeszutek: I thought about this a bit more and I came to a conclusion that `...connections.min` does not make any sense, especially for dbcp2. Idle connection eviction may be disabled (it is in fact disabled by default) and actual min will always = 0, regardless what `idleMin` is actually set to.
You're right, it is confusing when you think about it a bit deeper -- although the min/max (over)simplification is tempting at first.

I changed the names you suggested a bit.

breedx-splk: Looks good, thanks!

breedx-splk (approved):

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/CommonsDbcp2InstrumentationModule.java
```diff
@@ -0,0 +1,84 @@
...
+@AutoService(InstrumentationModule.class)
+public class CommonsDbcp2InstrumentationModule extends InstrumentationModule {
+  public CommonsDbcp2InstrumentationModule() {
+    super("commons-dbcp2");
+  }
+
+  @Override
+  protected boolean defaultEnabled() {
+    return Config.get().getBooleanProperty("splunk.metrics.enabled", true)
+        && super.defaultEnabled();
```
iNikem, line 47: Can we somehow tie this to the presence of metric exporter? Instead of having a separate configuration.

iNikem (approved):

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/CommonsDbcp2InstrumentationModule.java
mateuszrzeszutek: 👍 
Done

mateuszrzeszutek:
> have we verified that these metrics make it to the splunk backends and are properly chartable, etc?

image

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/DataSourceMetrics.java
```diff
@@ -0,0 +1,86 @@
...
+  private static final String CONNECTIONS_IDLE_MIN = "db.pool.connections.idle.min";
+  /** The maximum number of idle open connections allowed. */
+  private static final String CONNECTIONS_IDLE_MAX = "db.pool.connections.idle.max";
+  /** The maximum number of open connections allowed. */
+  private static final String CONNECTIONS_MAX = "db.pool.connections.max";
+
+  public static void registerMetrics(BasicDataSourceMXBean dataSource, ObjectName objectName) {
+    List<Tag> tags = getTags(objectName);
+
+    gauge(CONNECTIONS_TOTAL, tags, new TotalConnectionsUsed(dataSource));
```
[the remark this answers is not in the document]
mateuszrzeszutek: Oh, that is a very good point -- I'll fix that in another PR. Thanks!

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/CommonsDbcp2InstrumentationModule.java
```diff
@@ -0,0 +1,84 @@
...
+  private static class BasicDataSourceInstrumentation implements TypeInstrumentation {
+    @Override
+    public ElementMatcher<? super TypeDescription> typeMatcher() {
+      return named("org.apache.commons.dbcp2.BasicDataSource");
+    }
+
+    @Override
+    public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+      return singletonMap(
+          isPublic().and(named("preRegister")).and(takesArguments(2)),
```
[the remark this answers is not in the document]
mateuszrzeszutek: In a Spring application Spring does it automatically -- for manually created and managed connection pools you have to set the pool's JMX name for it to be registered (which is not set by default).
I think it was reasonable to hook on the `preRegister` method, since it provides us a unique object name - even if for some users (which I suspect won't be many) it won't ever execute.

PR #146, opened 2021-03-31
Unregister DBCP2 metrics when removed from JMX

mateuszrzeszutek:
See https://github.com/signalfx/splunk-otel-java/pull/144#discussion_r604700266

### DIFF
--- a/instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/CommonsDbcp2InstrumentationModule.java
+++ b/instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/CommonsDbcp2InstrumentationModule.java
@@ -25,6 +25,7 @@ import com.google.auto.service.AutoService;
 import io.micrometer.core.instrument.Metrics;
 import io.opentelemetry.javaagent.tooling.InstrumentationModule;
 import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import javax.management.ObjectName;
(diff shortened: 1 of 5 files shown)


instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/CommonsDbcp2InstrumentationModule.java
```diff
@@ -68,9 +67,16 @@ protected boolean defaultEnabled() {
 
     @Override
     public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
-      return singletonMap(
+      Map<ElementMatcher<MethodDescription>, String> transformers = new java.util.HashMap<>();
```
laurit, line 70: Could import HashMap

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/DataSourceMetrics.java
```diff
@@ -41,15 +45,29 @@
   /** The maximum number of open connections allowed. */
   private static final String CONNECTIONS_MAX = "db.pool.connections.max";
 
+  // a weak map does not make sense here because each Meter holds a reference to the dataSource
+  private static final Map<BasicDataSourceMXBean, List<Meter>> dataSourceMetrics =
```
laurit, line 49: This could behave in unexpected ways when there is a `BasicDataSourceMXBean` implementation that defines `equals` and `hashCode`. Idk whether this is something we should care about as probably `BasicDataSource` is the only implementation and it does not have `equals` and `hashCode`. `Collections.synchronizedMap(new IdentityHashMap<>())` would be the safe choice. I'll leave it to you to decide whether to change this or not.

laurit (approved):

instrumentation/commons-dbcp2/src/main/java/com/splunk/opentelemetry/commonsdbcp2/DataSourceMetrics.java
mateuszrzeszutek: > Idk whether this is something we should care about as probably BasicDataSource is the only implementation and it does not have equals and hashCode. 

Exactly this ^^

I'll add the comment explaining why it's safe to use a plain `ConcurrentHashMap` though

PR #152, opened 2021-04-01
Add more detailed Server-Timing docs

mateuszrzeszutek:


docs/server-timing.md
```diff
@@ -0,0 +1,28 @@
...
+* https://www.w3.org/TR/trace-context/#traceparent-header
+
+## Supported frameworks and libraries
+
+The following server frameworks and libraries are supported:
+
+| Framework/Library | Versions                            |
+|-------------------|-------------------------------------|
+| Servlet API       | 2.2-4.X (5.0+ is not supported yet) |
+| Netty             | 3.8-4.0 (4.1+ is not supported yet) |
```
iNikem, line 28: Why 4.1 is not supported? It should not be hard?

iNikem (approved):

docs/server-timing.md
mateuszrzeszutek: It shouldn't - when I was writing it it was missing the `CallDepth#get()` method, now that it's there it probably should be pretty easy to implement that.
I think there's a story for that (and all other HTTP server frameworks) in our backlog.

jkwatson (approved):

PR #185, opened 2021-04-06
Rename the Server-Timing header property

mateuszrzeszutek:
And use `true` as the default value.

README changes in another PR, they need to be merged after release.


breedx-splk (approved):
LGTM

iNikem (approved):

mateuszrzeszutek:
> docs will get updated when we release, I assume?

Yes, exactly that

Issue #137, opened 2021-03-22
Use otel-collector-contrib image in Windows smoke tests


mateuszrzeszutek:
And enable micrometer metrics for windows smoke tests.

But first, this PR: https://github.com/open-telemetry/opentelemetry-collector/pull/2272 needs to be ported to otel-collector-contrib repo.

iNikem:
Why contrib?

mateuszrzeszutek:
For signalfx metrics receiver, the "vanilla" otel-collector doesn't have it.

PR #190, opened 2021-04-08
Build splunk-otel-collector Windows docker image for tests

mateuszrzeszutek:
Resolves https://github.com/signalfx/splunk-otel-java/issues/137

Ideally we should use splunk-otel-collector everywhere, but that needs to wait until splunk-otel-collector has a Windows docker image.

### DIFF
--- a/smoke-tests/otel-collector-contrib-windows/Dockerfile
+++ b/smoke-tests/otel-collector-contrib-windows/Dockerfile
@@ -1,0 +1,4 @@
+FROM mcr.microsoft.com/windows/servercore:ltsc2019
+ADD https://github.com/signalfx/splunk-otel-collector/releases/latest/download/otelcol_windows_amd64.exe otelcol_windows_amd64.exe
+ENV NO_WINDOWS_SERVICE=1
+ENTRYPOINT /otelcol_windows_amd64.exe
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yaml
```diff
@@ -32,5 +32,11 @@ jobs:
           distribution: adopt
           java-version: 11
 
+      - name: Build Windows Collector image
```
iNikem, line 35: Do we need to build this for every PR? Or we can just pull the image built by CI job?

iNikem (commented):
Can we instead contribute to https://github.com/open-telemetry/opentelemetry-collector/issues/2715 ?

.github/workflows/pr.yaml
mateuszrzeszutek: I'll change the CI job 👍

mateuszrzeszutek:
> Can we instead contribute to open-telemetry/opentelemetry-collector#2715 ?

We're using the signalfx metrics receiver, so actually we'd like to get this one done: https://github.com/signalfx/splunk-otel-collector/issues/260

It'd be nice to have both of them, but I have no idea how go/collector build works 😞

smoke-tests/otel-collector-contrib-windows/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM mcr.microsoft.com/windows/servercore:ltsc2019
+ADD https://github.com/signalfx/splunk-otel-collector/releases/latest/download/otelcol_windows_amd64.exe otelcol_windows_amd64.exe
```
iNikem, line 2: I think we need contrib here, no?

mateuszrzeszutek: 🤦  I forgot about the most important thing

iNikem (approved):

PR #192, opened 2021-04-08
Test jaeger-thrift-splunk exporter (especially the splunk access token)

mateuszrzeszutek:
This was probably the only major thing that was completely not covered by the smoke tests.

### DIFF
--- a/custom/build.gradle
+++ b/custom/build.gradle
@@ -27,8 +27,15 @@ dependencies {
   }
 
   testImplementation("io.opentelemetry:opentelemetry-sdk:${versions.opentelemetry}")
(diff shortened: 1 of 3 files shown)


jkwatson (approved):
any thoughts to using TestContainers and using an actual Jaeger backend for this kind of test?

breedx-splk (approved):

mateuszrzeszutek:
> any thoughts to using TestContainers and using an actual Jaeger backend for this kind of test?

Hmm, will I be able to verify that the Splunk token was passed?

jkwatson:
> > any thoughts to using TestContainers and using an actual Jaeger backend for this kind of test?
> 
> Hmm, will I be able to verify that the Splunk token was passed?

Excellent point. I'm not sure; something to investigate at some point, though.

PR #203, opened 2021-04-14
Update README and CHANGELOG after v0.10.0 release

mateuszrzeszutek:


.github/workflows/release.yaml
```diff
@@ -37,3 +37,4 @@ jobs:
         with:
           name: Release ${{ steps.get_version.outputs.VERSION }}
           files: splunk-otel-javaagent-all.jar
+          draft: true
```
mateuszrzeszutek, line 40: The release job should create draft release by default - we will hit the "make it official" button manually once we fill in the release notes.

iNikem (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,46 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [v0.10.0]
+
+### Added
+
+- Implement gathering basic JVM metrics with micrometer.
```
jkwatson, line 14: Personally, I'd rather these changelog entries were more end-user centric. So, rather than just copying all the PRs, instead explain to the user what the new feature is, and how they can get access to it.

mateuszrzeszutek: Good point, done. I'll also update the release notes so that they're a bit more user-friendly.

jkwatson (approved):

PR #210, opened 2021-04-20
Versioning strategy

kubawach:
First sketch of versioning strategy.

Maintainers - please have a look.

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,0 +1,32 @@
+# Splunk Distribution of OpenTelemetry Java Instrumentation Versioning
+
+## Version numbers
+
+The project strictly follows [Semantic Versioning 2.0.0](https://semver.org/). This means
+that all artifacts have a version of the format `MAJOR.MINOR.PATCH`.
+No backwards-incompatible changes will be made unless incrementing the `MAJOR` version number. 
+Most releases will be made by incrementing the `MINOR` version. 
+Patch releases will be made by incrementing the `PATCH` version.
+
+## Compatibility  
+
+Backwards-incompatible changes will be avoided - if they are necessary, the `MAJOR` version of the artifact will be
+incremented.
+
+This can occur in following situations:
+- OpenTelemetry Java Instrumentation breaking change - please see [upstream repository](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/VERSIONING.md) for their versioning policy
+- change in a Splunk-managed configuration property or code
+
+In the latter case, we will first **deprecate** (at least one release earlier) before introducing the change, so that customers can prepare accordingly.
+
+All the changes are to be communicated in the [changelog](CHANGELOG.md).
+
+### Old versions support
+
+Splunk is committed to support the customers using this distribution. All versions will get critical (eg security) patches up to **one year** after the release date. 
+
+## Release cadence
+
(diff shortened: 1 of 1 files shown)


VERSIONING.md
```diff
@@ -0,0 +1,32 @@
...
+
+All the changes are to be communicated in the [changelog](CHANGELOG.md).
+
+### Old versions support
+
+Splunk is committed to support the customers using this distribution. All versions will get critical (eg security) patches up to **one year** after the release date. 
+
+## Release cadence
+
+Currently, the upstream is released on a **monthly** cadence. This distribution will be updated no later than **2 weeks** after an upstream release.
```
mateuszrzeszutek, line 30: The [FAQ](https://github.com/signalfx/splunk-otel-java/blob/main/docs/faq.md#how-often-do-you-release) says that we try to release at most 2 days after the upstream. Do we want to specify a strict SLA here? Or is saying "most of the cases it's less than 2 days" okay? @flands

VERSIONING.md
```diff
@@ -0,0 +1,32 @@
...
+- OpenTelemetry Java Instrumentation breaking change - please see [upstream repository](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/VERSIONING.md) for their versioning policy
+- change in a Splunk-managed configuration property or code
+
+In the latter case, we will first **deprecate** (at least one release earlier) before introducing the change, so that customers can prepare accordingly.
+
+All the changes are to be communicated in the [changelog](CHANGELOG.md).
+
+### Old versions support
+
+Splunk is committed to support the customers using this distribution. All versions will get critical (eg security) patches up to **one year** after the release date.
```
laurit, line 26: Perhaps this should read all "All major versions". I doubt that we want to backport stuff to all minor releases. Also up to one year sounds like definitely not more that a year, but could be less.

VERSIONING.md
```diff
@@ -0,0 +1,32 @@
...
+Most releases will be made by incrementing the `MINOR` version. 
+Patch releases will be made by incrementing the `PATCH` version.
+
+## Compatibility  
+
+Backwards-incompatible changes will be avoided - if they are necessary, the `MAJOR` version of the artifact will be
+incremented.
+
+This can occur in following situations:
+- OpenTelemetry Java Instrumentation breaking change - please see [upstream repository](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/VERSIONING.md) for their versioning policy
```
jkwatson, line 17: It's also possible that this happens because of an upstream-upstream breaking change in OpenTelemetry-Java itself, yes? Like, if something with the autoconfiguration artifact changes in a breaking way? (especially since that's not marked stable at this point)

kubawach: Yes, but I believe that from our and customer point of view the major release will be triggered by major release of the upstream (ie if otel-java inroduces breaking change, that will be followed by otel-java agent release and then - splunk release). So I'd leave this wording to avoid confusion.

mateuszrzeszutek (approved):

VERSIONING.md
mateuszrzeszutek: OTel Java Instrumentation VERSIONING file itself relies on the OTel Java versioning policy.
Still, it wouldn't hurt to be a bit more explicit.

PR #221, opened 2021-05-06
Implement configuration changes required by GDI spec

mateuszrzeszutek:
* set default traces exporter to otlp
* support `SPLUNK_ACCESS_TOKEN` in otlp
* set default propagators to `tracecontext,baggage`
* log a warning if `service.name` resource attribute is missing

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
@@ -16,8 +16,6 @@
 
 package com.splunk.opentelemetry;
 
-import static com.splunk.opentelemetry.jaeger.JaegerThriftSpanExporterFactory.OTEL_EXPORTER_JAEGER_ENDPOINT;
-
 import com.google.auto.service.AutoService;
 import io.opentelemetry.javaagent.spi.config.PropertySource;
 import java.util.HashMap;
@@ -25,6 +23,9 @@ import java.util.Map;
 
 @AutoService(PropertySource.class)
 public class SplunkConfiguration implements PropertySource {
+  public static final String SPLUNK_ACCESS_TOKEN = "splunk.access.token";
+  public static final String OTEL_EXPORTER_JAEGER_ENDPOINT = "otel.exporter.jaeger.endpoint";
+
   @Override
   public Map<String, String> getProperties() {
     Map<String, String> config = new HashMap<>();
@@ -34,11 +35,11 @@ public class SplunkConfiguration implements PropertySource {
     // disable otel runtime-metrics instrumentation; we use micrometer metrics instead
     config.put("otel.instrumentation.runtime-metrics.enabled", "false");
 
-    config.put("otel.traces.exporter", "jaeger-thrift-splunk");
+    config.put("otel.traces.exporter", "otlp");
     // http://localhost:9080/v1/trace is the default endpoint for SmartAgent
     // http://localhost:14268/api/traces is the default endpoint for otel-collector
     config.put(OTEL_EXPORTER_JAEGER_ENDPOINT, "http://localhost:9080/v1/trace");
-    config.put("otel.propagators", "b3multi");
+    config.put("otel.propagators", "tracecontext,baggage");
 
     // enable experimental instrumentation
     config.put("otel.instrumentation.spring-batch.enabled", "true");
--- a/smoke-tests/src/test/java/com/splunk/opentelemetry/helper/AbstractTestContainerManager.java
+++ b/smoke-tests/src/test/java/com/splunk/opentelemetry/helper/AbstractTestContainerManager.java
@@ -35,6 +35,7 @@ public abstract class AbstractTestContainerManager implements TestContainerManag
     environment.put("OTEL_BSP_SCHEDULE_DELAY", "10ms");
     environment.put(
         "OTEL_EXPORTER_JAEGER_ENDPOINT", "http://" + COLLECTOR_ALIAS + ":14268/api/traces");
+    environment.put("OTEL_EXPORTER_OTLP_ENDPOINT", "http://" + COLLECTOR_ALIAS + ":4317");
     // export metrics every 1s
     environment.put("SPLUNK_METRICS_EXPORT_INTERVAL", "1000");
     environment.put("SPLUNK_METRICS_ENDPOINT", "http://" + COLLECTOR_ALIAS + ":9943/v2/datapoint");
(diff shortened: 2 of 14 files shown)


custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
```diff
@@ -34,11 +35,11 @@
     // disable otel runtime-metrics instrumentation; we use micrometer metrics instead
     config.put("otel.instrumentation.runtime-metrics.enabled", "false");
 
-    config.put("otel.traces.exporter", "jaeger-thrift-splunk");
+    config.put("otel.traces.exporter", "otlp");
     // http://localhost:9080/v1/trace is the default endpoint for SmartAgent
     // http://localhost:14268/api/traces is the default endpoint for otel-collector
     config.put(OTEL_EXPORTER_JAEGER_ENDPOINT, "http://localhost:9080/v1/trace");
```
iNikem, line 41: If we use otlp then we should not set this property, right?

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/AbstractTestContainerManager.java
```diff
@@ -35,6 +35,7 @@
     environment.put("OTEL_BSP_SCHEDULE_DELAY", "10ms");
     environment.put(
         "OTEL_EXPORTER_JAEGER_ENDPOINT", "http://" + COLLECTOR_ALIAS + ":14268/api/traces");
```
iNikem, line 37: Do we need this property now?

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
```diff
@@ -34,11 +35,11 @@
     // disable otel runtime-metrics instrumentation; we use micrometer metrics instead
     config.put("otel.instrumentation.runtime-metrics.enabled", "false");
 
-    config.put("otel.traces.exporter", "jaeger-thrift-splunk");
+    config.put("otel.traces.exporter", "otlp");
```
iNikem, line 38: We should put this into changelog as breaking change.

iNikem:
MIGRATING.md should be changed as well

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
mateuszrzeszutek: This is just the default value, it should always be set.

smoke-tests/src/test/java/com/splunk/opentelemetry/helper/AbstractTestContainerManager.java
mateuszrzeszutek: Hmm, nope - I'll move it to that one test that actually uses it.

mateuszrzeszutek:
> MIGRATING.md should be changed as well

I'll prepare a separate PR for documentation changes - it'll be merged just before the release. The current version is still using jaeger-thrift so the docs shouldn't say anything about otlp until we're ready to release.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
jkwatson: Isn't this the otel default? do we actually need to put this in here at all if we're switching to otlp?

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
```diff
@@ -34,11 +35,11 @@
...
     // http://localhost:9080/v1/trace is the default endpoint for SmartAgent
     // http://localhost:14268/api/traces is the default endpoint for otel-collector
     config.put(OTEL_EXPORTER_JAEGER_ENDPOINT, "http://localhost:9080/v1/trace");
-    config.put("otel.propagators", "b3multi");
+    config.put("otel.propagators", "tracecontext,baggage");
```
jkwatson, line 42: this is the default, correct? So we shouldn't need to set it at all?

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
mateuszrzeszutek: We don't, same for propagators: but I thought that being a little explicit won't hurt.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
mateuszrzeszutek: Yes, it's the default value in the SDK - I thought that explicitly putting this here won't hurt, and it makes checking if we adhere to the GDI spec a little bit easier.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
iNikem: I agree with John: no need to explicitly set default values

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
jkwatson: That's not a bad reason; would like other thoughts on whether we should re-iterate the defaults in our configuration.

iNikem (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,23 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+
+- The Splunk Distribution of OpenTelemetry Java Instrumentation now uses the OTLP span exporter as the default. The OTLP
+  exporter supports the `splunk.access.token` configuration option and can be used to send spans directly to Splunk
+  cloud. The default OTLP exporter endpoint is `http://localhost:4317`. You can still use the Jaeger exporter by
+  setting `OTEL_TRACES_EXPORTER=jaeger-thrift-splunk`.
+- We have also changed the default endpoint of the SignalFx metrics exporter: it now points to `http://localhost:9943`,
+  which is the default endpoint of [Splunk OpenTelemetry Connector](https://github.com/signalfx/splunk-otel-collector)
+  deployed on `localhost`.
+- The agent now uses W3C `tracecontext` as the default trace propagation
+  mechanism; [OpenTelemetry Baggage propagation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#propagation)
```
jkwatson, line 20: It's odd to point at the baggage spec when referring to tracecontext. We should point at the tracecontext spec, and perhaps mention that W3C baggage will also be propagated by default.

mateuszrzeszutek: Done - I pointed to both tracecontext and baggage spec.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
mateuszrzeszutek: That was the only major reason that I had.

PR #239, opened 2021-05-19
Disable experimental components by default

flands:
- Disable metric collection
- Add experimental banner to relevant pages

Also alphabetize table env vars


jkwatson:
I'm curious why we would want to disable the server-timing header. Is it harmful to have it enabled by default? I would think it has near zero impact, in general.

iNikem:
+1 to John's question. This was part of epic's acceptance criteria to have it enabled by default, if I am not mistaken

docs/advanced-config.md
```diff
@@ -11,13 +11,13 @@ Below you will find all the configuration options supported by this distribution
...
+| System property                        | Environment variable                   | Default value           | Support | Purpose                                                                                                                                                                                                                                                          |
+| -------------------------------------- | -------------------------------------- | ----------------------- | ------- | -------                                                                                                                                                                                                                                                          |
+| `splunk.access.token`                  | `SPLUNK_ACCESS_TOKEN`                  | unset                   | Beta  | (Optional) Auth token allowing exporters to communicate directly with the Splunk cloud, passed as `X-SF-TOKEN` header. Currently the [both Jaeger and OTLP trace exporters](#trace-exporters) and [SignalFx metrics exporter](metrics.md) support this property.
```
mateuszrzeszutek, line 16: Is `splunk.access.token` beta too? I thought that it was a stable feature.

Also, should we use word "beta" or "experimental"? I think we use "experimental" for all other features, the whole project is in "beta".

mateuszrzeszutek:
> I'm curious why we would want to disable the server-timing header. Is it harmful to have it enabled by default?

It's also required to be true by the GDI spec. Should we change the spec first?

docs/advanced-config.md
flands: Idea is to mark stable with 1.0 release. Experimental is likely a better term -- will update.

flands:
> I'm curious why we would want to disable the server-timing header. Is it harmful to have it enabled by default? I would think it has near zero impact, in general.

With goal to move this to OTel, thought was to not mark stable so easy to remove. Not marking stable means off by default. Sounds like this may take longer than expected so will opt to mark stable/on and we can deprecate when OTel is ready (per GDI spec)

mateuszrzeszutek (approved):

PR #242, opened 2021-05-20
Add splunk.distro.version resource attribute

mateuszrzeszutek:
Implements https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md#splunk-resource-attributes

I didn't touch the agent version (and `telemetry.auto.version`) in this PR, it's still

```groovy
attributes.put("Implementation-Version", "splunk-${project.version}-otel-${versions["opentelemetryJavaagent"]}")
```

I think that first the upstream agent has to stop using `Implementation-Version` to determine the agent version; any distro can (and probably will) override it.

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/SplunkDistroVersionResourceProvider.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/SplunkDistroVersionResourceProvider.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry;
+
+import com.google.auto.service.AutoService;
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.sdk.autoconfigure.ConfigProperties;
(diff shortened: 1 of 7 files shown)


custom/src/main/java/com/splunk/opentelemetry/SplunkDistroVersionResourceProvider.java
```diff
@@ -0,0 +1,57 @@
...
+  static final AttributeKey<String> SPLUNK_DISTRO_VERSION =
+      AttributeKey.stringKey("splunk.distro.version");
+
+  private static final Resource DISTRO_VERSION_RESOURCE = initialize();
+
+  private static Resource initialize() {
+    InputStream in =
+        Thread.currentThread().getContextClassLoader().getResourceAsStream("splunk.properties");
+    if (in == null) {
+      return Resource.empty();
```
jkwatson, line 40: would it be better to populate this with something like "unknown" or "dev" or something that signals that it wasn't just missed somehow?

mateuszrzeszutek: If it's missed somehow and the attribute isn't added our smoke tests will detect that and fail.
Anyway, since the file is baked in to the javaagent jar this scenario is not be possible unless someone tampers with our jar and removes the file.

jkwatson (approved):

custom/src/main/java/com/splunk/opentelemetry/SplunkDistroVersionResourceProvider.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+    try (InputStream ignored = in) {
```
breedx-splk, line 42: This double stream thing reads weird to me.  Does try-with-resources internally NPE on the null case?  If not, why not just put the try-with-resources at the top of the method and check for null inside?  I think it would read cleaner (if it works).

breedx-splk (approved):
Looks good.  Had one thought about reading the stream from the classloader, if it's viable.

custom/src/main/java/com/splunk/opentelemetry/SplunkDistroVersionResourceProvider.java
mateuszrzeszutek: > Does try-with-resources internally NPE on the null case?

It does! I didn't know that before 😄 
Thanks!

PR #248, opened 2021-05-21
Profiler module scaffolding

breedx-splk:
Set up the scaffolding for adding profiling code. The activation doesn't do anything yet, but it will check for `splunk.profiler.enabled=true`, which defaults to false. 

I also thought we'd want to exclude the profiling code from builds for a while (since we're close to 1.0), so I made that conditional in the agent with a gradle property.  We can remove that after 1.0.


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -0,0 +1,42 @@
...
+import org.slf4j.LoggerFactory;
+
+@AutoService(ComponentInstaller.class)
+public class JfrActivator implements ComponentInstaller {
+
+  private static final Logger logger = LoggerFactory.getLogger(JfrActivator.class.getName());
+
+  @Override
+  public void afterByteBuddyAgent(Config config) {
+    if ("false".equals(config.getProperty(CONFIG_KEY_ENABLE_PROFILER))) {
```
mateuszrzeszutek, line 34: Why not use the typed method?

```suggestion
    if (config.getBooleanProperty(CONFIG_KEY_ENABLE_PROFILER, true)) {
```

Also if you just use `false` as the default here you won't need a custom `PropertySource` implementation

profiler/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+}
+
+apply from: "$rootDir/gradle/shadow.gradle"
+
+def relocatePackages = ext.relocatePackages
+
+dependencies {
+    compileOnly("io.opentelemetry:opentelemetry-sdk:${versions.opentelemetry}")
+    compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:${versions.opentelemetryAlpha}")
+    compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-spi:${versions.opentelemetryJavaagentAlpha}")
```
mateuszrzeszutek, line 12: `javaagent-spi` does not exist anymore:

```suggestion
    compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api:${versions.opentelemetryJavaagentAlpha}")
```

profiler/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+    compileOnly("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:${versions.opentelemetryJavaagentAlpha}")
+    compileOnly("io.opentelemetry:opentelemetry-semconv:1.0.1-alpha")
+    compileOnly deps.slf4j
+
+    annotationProcessor("com.google.auto.service:auto-service:1.0-rc7")
+    annotationProcessor("com.google.auto:auto-common:0.8")
+    implementation("com.google.auto.service:auto-service:1.0-rc7")
+    implementation("com.google.auto:auto-common:0.8")
```
mateuszrzeszutek, line 20: You can use the `deps` helper map:
```suggestion
    annotationProcessor deps.autoservice
    implementation deps.autoservice
```

profiler/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+apply from: "$rootDir/gradle/shadow.gradle"
+
+def relocatePackages = ext.relocatePackages
+
+dependencies {
+    compileOnly("io.opentelemetry:opentelemetry-sdk:${versions.opentelemetry}")
+    compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:${versions.opentelemetryAlpha}")
+    compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-spi:${versions.opentelemetryJavaagentAlpha}")
+    compileOnly("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:${versions.opentelemetryJavaagentAlpha}")
+    compileOnly("io.opentelemetry:opentelemetry-semconv:1.0.1-alpha")
```
mateuszrzeszutek, line 14:
```suggestion
    compileOnly("io.opentelemetry:opentelemetry-semconv:${versions.opentelemetryAlpha}")
```

agent/build.gradle
```diff
@@ -20,6 +20,9 @@ configurations {
 dependencies {
     customShadow project(path: ":custom", configuration: "shadow")
     customShadow project(path: ":instrumentation", configuration: "shadow")
+    if(project.hasProperty('profiler')) {
+        customShadow project(path: ":profiler", configuration: "shadow")
+    }
```
mateuszrzeszutek, line 25: 👍

mateuszrzeszutek (commented):
🎉

profiler/build.gradle
breedx-splk: Yeah, I caught that the hard way.  :). Thanks.  I've updated it.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
breedx-splk: Nice.  I think I had just forgotten that there were typed methods, thanks for the suggestion!

There will be other properties that justify the property source.  :)

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -0,0 +1,42 @@
...
+import org.slf4j.LoggerFactory;
+
+@AutoService(ComponentInstaller.class)
+public class JfrActivator implements ComponentInstaller {
+
+  private static final Logger logger = LoggerFactory.getLogger(JfrActivator.class.getName());
+
+  @Override
+  public void afterByteBuddyAgent(Config config) {
+    if (config.getBooleanProperty(CONFIG_KEY_ENABLE_PROFILER, true)) {
```
iNikem, line 34: If config returns true, then print DISABLED, otherwise print "active"? Does not make sense :D

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
breedx-splk: Dangit, I flopped the logic since I wanted to start disabled...but clearly I screwed that up.  Thanks, I'll fix it!

breedx-splk:
Ugh, let me get out the mop.....

PR #252, opened 2021-05-24
Add GDI spec badge

mateuszrzeszutek:


jkwatson (approved):
cool! Didn't know there were arbitrary badge-makers like this.

PR #253, opened 2021-05-24
One more docs update after GDI spec review

mateuszrzeszutek:
* Add supported library versions to metrics doc
* Add Support=Stable column to all config options that did not have it
in advanced-config.md


jkwatson (approved):

breedx-splk (approved):
Great!

PR #256, opened 2021-05-24
Profiler module scaffolding

breedx-splk:
(sorry for recreating, I had made a mess of the #248)

Set up the scaffolding for adding profiling code. The activation doesn't do anything yet, but it will check for splunk.profiler.enabled=true, which defaults to false.

I also thought we'd want to exclude the profiling code from builds for a while (since we're close to 1.0), so I made that conditional in the agent with a gradle property. We can remove that after 1.0.


mateuszrzeszutek (approved):

mateuszrzeszutek:
@breedx-splk please rebase to fix PR checks.

PR #261, opened 2021-05-25
Add pre- and post-release helper scripts for updating versions

mateuszrzeszutek:
Those pre- and post-release scripts should help with updating READMEs in this repo.

TODO: CHANGELOG.md

### DIFF
--- a/pre-release.sh
+++ b/pre-release.sh
@@ -2,19 +2,59 @@
 
 set -e
 
-if [[ $# < 4 ]]
+print_usage() {
+  cat <<EOF
+Usage: $(basename $0) splunk_old_version splunk_new_version
+    [--otel-instrumentation otel_instrumentation_old_version otel_instrumentation_new_version]
+    [--otel otel_old_version otel_new_version]
+
+This projects old and new versions are mandatory arguments; OTel instrumentation
+and API versions are optional.
+All versions MUST NOT begin with 'v'. Example: 1.2.3".
+EOF
+}
+
+if [[ $# < 2 ]]
 then
-  echo "Usage $(basename $0) splunk_old_version splunk_new_version otel_instrumentation_old_version otel_instrumentation_new_version [otel_old_version] [otel_new_version]"
-  echo "  All versions MUST NOT begin with 'v'. Example: 1.2.3"
+  print_usage
   exit 1
 fi
 
 splunk_old_version=$1
 splunk_new_version=$2
-otel_instrumentation_old_version=$3
-otel_instrumentation_new_version=$4
-otel_old_version=${5:-${otel_instrumentation_old_version}}
-otel_new_version=${6:-${otel_instrumentation_new_version}}
+shift 2
+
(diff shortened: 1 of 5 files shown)


pre-release.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [[ $# < 4 ]]
+then
+  echo "Usage $(basename $0) splunk_old_version splunk_new_version otel_instrumentation_old_version otel_instrumentation_new_version [otel_old_version] [otel_new_version]"
```
breedx-splk, line 7: I feel like this is a lot of similarly named arguments to read horizontally.  Maybe split to new lines for readability?  Alternatively, if we think this is always manual we could prompt for them...

breedx-splk: Or we could use getopts.

pre-release.sh
```diff
@@ -0,0 +1,46 @@
...
+  -e "s/<!--OTEL_VERSION-->${otel_old_version}<!--OTEL_VERSION-->/<!--OTEL_VERSION-->${otel_new_version}<!--OTEL_VERSION-->/g"
+
+  # update upstream otel link & badge
+  -e "s https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v${otel_instrumentation_old_version} https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v${otel_instrumentation_new_version} g"
+  -e "s https://img.shields.io/badge/otel-${otel_instrumentation_old_version} https://img.shields.io/badge/otel-${otel_instrumentation_new_version} g"
+
+  # remove development version docs warning
+  -e '/<!--DEV_DOCS_WARNING_START-->/,/<!--DEV_DOCS_WARNING_END-->/d'
+)
+sed ${sed_flag} "${readme_sed_args[@]}" README.md
```
breedx-splk, line 41: This feels a little more complicated than using line continuations, but I guess that works.  😁

breedx-splk (approved):
Thanks for taking this on and helping future us save keystrokes and reduce mistakes.  👍🏻 

I do wonder if later we might benefit from using an actual template language (like handlebars or jinja or something) if this gets much more involved.

pre-release.sh
mateuszrzeszutek: It allows me to add comments in the middle of the array - and you can't do that with line continuations 😄

pre-release.sh
mateuszrzeszutek: I thought about using getops or case/shift, but then I decided to keep it simple...
But you're right, it would be more readable that way -- will refactor.

> Alternatively, if we think this is always manual we could prompt for them...

I'm hoping that sometime in the future this script (or its future version) will be executed by an automatic process, so I wanted to avoid prompts. Although I don't know how distant that future is 😅

pre-release.sh
breedx-splk: True!  :)

PR #262, opened 2021-05-25
Silence confusing signalfx metrics exporter INFO log

mateuszrzeszutek:
Closes https://github.com/signalfx/splunk-otel-java/issues/206

PR #263, opened 2021-05-26
Correctly silence confusing signalfx metrics exporter INFO log

mateuszrzeszutek:
Fixes https://github.com/signalfx/splunk-otel-java/pull/262

This morning I realized that config is initialized after logging, so the previous solution couldn't function properly. This one does (I tested it).


kubawach (approved):
Nice.

PR #264, opened 2021-05-26
Add a profile recording sequencer.

breedx-splk:
This sequencer will periodically and conditionally flush recording data to disk.  The predicate and recorder are simply stubbed for now and will be fleshed out in subsequent/following PRs.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -50,12 +50,12 @@ public class JfrActivator implements ComponentInstaller {
   private void activateJfrAndRunForever(Config config) {
     String recordingDurationStr = config.getProperty(CONFIG_KEY_RECORDING_DURATION_SECONDS, "20");
     Duration recordingDuration = Duration.ofSeconds(Integer.parseInt(recordingDurationStr));
-    RecordingStartPredicate recordingStartPredicate = new RecordingStartPredicate();
+    RecordingEscapeHatch recordingEscapeHatch = new RecordingEscapeHatch();
     JfrRecorder recorder = new JfrRecorder();
     RecordingSequencer sequencer =
         RecordingSequencer.builder()
             .recordingDuration(recordingDuration)
-            .recordingStartPredicate(recordingStartPredicate)
+            .recordingEscapeHatch(recordingEscapeHatch)
             .recorder(recorder)
             .build();
     sequencer.start();
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -17,10 +17,15 @@
 package com.splunk.opentelemetry.profiler;
 
 import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_ENABLE_PROFILER;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_RECORDING_DURATION_SECONDS;
+import static com.splunk.opentelemetry.profiler.util.HelpfulExecutors.logUncaught;
 
 import com.google.auto.service.AutoService;
+import com.splunk.opentelemetry.profiler.util.HelpfulExecutors;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.javaagent.spi.ComponentInstaller;
+import java.time.Duration;
+import java.util.concurrent.ExecutorService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -28,6 +33,7 @@ import org.slf4j.LoggerFactory;
 public class JfrActivator implements ComponentInstaller {
 
   private static final Logger logger = LoggerFactory.getLogger(JfrActivator.class.getName());
+  private final ExecutorService executor = HelpfulExecutors.newSingleThreadExecutor("JFR Profiler");
 
   @Override
   public void afterByteBuddyAgent(Config config) {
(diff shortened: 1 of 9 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -38,5 +44,20 @@ public void afterByteBuddyAgent(Config config) {
       return;
     }
     logger.info("JFR profiler is active.");
+    executor.submit(logUncaught(() -> activateJfrAndRunForever(config)));
+  }
+
+  private void activateJfrAndRunForever(Config config) {
+    String recordingDurationStr = config.getProperty(CONFIG_KEY_RECORDING_DURATION_SECONDS, "20");
```
jkwatson, line 51: any particular reason 20s is a good default?

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencer.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public void start() {
+    int period = (int) (recordingDuration.toMillis() * OVERLAP_FACTOR);
+    executor.scheduleAtFixedRate(this::handleInterval, 0, period, TimeUnit.MILLISECONDS);
+  }
+
+  @VisibleForTesting
+  void handleInterval() {
+    if (!recordingEscapeHatch.jfrCanContinue()) {
+      logger.warn("JFR recordings cannot proceed.");
```
jkwatson, line 57: will this end up logging every 20s? Should we keep a reference to the Future from the schedule and cancel it when we see this?

jkwatson (approved):
Seems fine to me. Just a few questions.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
breedx-splk: Based on the events we plan on capturing and the frequency of those events, 20s seems like a reasonable time to keep the data volume lowish while also being timely.  This can be squishy I think.

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencer.java
breedx-splk: It will. The hope was to make it very obvious that something bad has happened and that it's broken.  Think disk filling up or completely backlogged (too many files bunching up).  It should be common, but we wanted to guard against it.

I don't want to cancel the schedule, because we want it to recover if things clear up.

If we hate that repeated logging, I supposed we could keep additional state and only log a max `n` times or something.

jkwatson: ah, ok. you expect it to be a transient issue (although disk filling up rarely is!). That makes sense.

jkwatson: It was hard to tell from this code whether the escape hatch might be something that the user pulls during runtime to turn off JFR.

breedx-splk: Nah, the request was for it to pause itself when things are bad.

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencer.java
```diff
@@ -0,0 +1,98 @@
...
+  // This forces overlap between recordings and ensures data is continuous when deduplicated.
+  public static final double OVERLAP_FACTOR = 0.8;
+
+  private final ScheduledExecutorService executor =
+      HelpfulExecutors.newSingleThreadedScheduledExecutor("JFR Recording Sequencer");
+  private final Duration recordingDuration;
+  private final RecordingEscapeHatch recordingEscapeHatch;
+  private final JfrRecorder recorder;
+
+  public RecordingSequencer(Builder builder) {
```
mateuszrzeszutek, line 43: Nit
```suggestion
 private RecordingSequencer(Builder builder) {
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/HelpfulExecutors.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public static ScheduledExecutorService newSingleThreadedScheduledExecutor(String threadName) {
+    return Executors.newSingleThreadScheduledExecutor(new HelpfulThreadFactory(threadName));
+  }
+
+  public static ExecutorService newSingleThreadExecutor(String name) {
+    return Executors.newSingleThreadExecutor(new HelpfulThreadFactory(name));
+  }
+
+  public static Runnable logUncaught(Runnable delegate) {
```
mateuszrzeszutek, line 40: This method does not seem to be used anywhere outside of tests. Will it be used in some later PR?

profiler/src/test/java/com/splunk/opentelemetry/profiler/RecordingSequencerTest.java
```diff
@@ -0,0 +1,181 @@
...
+    @Override
+    public void stop() {
+      if (!started) {
+        fail("not started");
+      }
+      started = false;
+      stopLatch.countDown();
+    }
+
+    public void setFlushLatch(CountDownLatch flushLatch) {
```
mateuszrzeszutek, line 177: Consider making this method `synchronized` (or the `flushLatch` field `volatile`), as it's called from another thread (test thread vs scheduled daemon thread)

mateuszrzeszutek (approved):

profiler/src/test/java/com/splunk/opentelemetry/profiler/RecordingSequencerTest.java
breedx-splk: Very good catch!  It's this kind of stuff that results in the random sporadic 1 in 100 test failure scenario.  Ugh.  I will definitely make it volatile. Thanks!

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/HelpfulExecutors.java
breedx-splk: It's called in the `JfrActivator` I believe.

PR #266, opened 2021-05-28
JfrRecorder core

breedx-splk:
This builds on #264 and will be rebased when that is merged.

This includes a settings file that lives in the jar and the settings are applied from that file.  The JfrRecorder is started and periodically snapshotted to disk with a timestamped filename.  For now, the files just accumulate on disk.  A following PR will include components to handle these files.


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
```diff
@@ -16,24 +16,96 @@
...
+  Recording newRecording() {
+    return new Recording();
   }
 
   public void flushSnapshot() {
-    logger.debug("Flushing a snapshot");
+    try {
+      Recording snap = jfr.takeSnapshot();
+      Path path = Path.of(Instant.now().toString() + ".jfr");
+      logger.debug("Flushing a JFR snapshot: " + path);
```
iNikem, line 66: Please, don't use string concatenation in logging.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
```diff
@@ -16,24 +16,96 @@
...
+      snap.dump(path);
+      snap.close();
```
iNikem, line 68: try-with-resources?

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrSettingsReader.java
```diff
@@ -0,0 +1,72 @@
...
+    return read(DEFAULT_JFR_SETTINGS);
+  }
+
+  public Map<String, String> read(String resourceName) {
+    Map<String, String> result = new HashMap<>();
+    try (BufferedReader reader = openResource(resourceName)) {
+      if (reader == null) {
+        return emptyMap();
+      }
+      reader
```
iNikem, line 46: Are you sure you don't want to use `java.util.Properties`?

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
breedx-splk: I think I missed a boat. 

Why?  I've seen Is templating elsewhere, but didn't imagine that it would be lower cost.  Happy to change it, but want to understand the intention.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrSettingsReader.java
breedx-splk: You might think that, huh....but at the end of the day this has to go into [Recording.setSettings()](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/jdk/jfr/Recording.html#setSettings(java.util.Map)) which takes a `Map<String,String>()`.  I don't know of a trivial way to convert props -> Map<String,String> because of the type differences...so I have to touch each thing somewhere.  

I just preferred to keep the model the same as JFR expects.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
jkwatson: I think that main reason is that if the logger is set to INFO, you'll still pay the (admittedly small) cost of the string concat even if nothing gets logged.

breedx-splk: OH!  Derp, yes, that is a very good reason.  🌹

breedx-splk: Thank you.

mateuszrzeszutek (approved):

PR #269, opened 2021-06-02
Use git fetch depth 0 in all GH workflows

mateuszrzeszutek:


breedx-splk (approved):
This feels like maybe a reaction to a problem that stung a little bit?  😄   It definitely seems like the easiest to understand, but I do wonder how it'll scale down the road.  It's probably fine.

mateuszrzeszutek:
Honestly I'm just copy-pasting this from the upstream repo 😄 
I don't think that this has ever become a problem in this repo, but hey, it's an almost free change.

jkwatson (approved):

PR #273, opened 2021-06-02
include profiling code in jar, even though disabled.

breedx-splk:
Now that 1.0.0 is released, let's go ahead and include the profiling code. The activator will be invoked early in the lifecycle, but the config default is still false and it should return early with a message logged at debug.

(this helps with the testing effort as well)


jkwatson (approved):

breedx-splk:
I am concerned that these failures might have been, in part, due to the java 8 incompat that I discovered and fixed in #274, so I'm going to just close this and let #274 take over from here (it builds on this one).

PR #274, opened 2021-06-02
Profiling smoke test

breedx-splk:
This builds on top of #273.

This makes a smoke test that fires up spring petclinic with the profiling agent enabled and ensures that JFR files are generated. It should be the foundational work for more thorough testing down the road.

In order to test, we also needed a way to configure the output directory for JFR files, so this changeset also includes that.

The smoketest has to build a spring petclinic rest image because the one in dockerhub is 2 years old and doesn't have a java 8 version with JFR.  Because of that, it's slower than I would prefer...so I'm inclined to file an issue to address that at a later time (maybe we can publish a base image like we do elsewhere).

### DIFF
--- a/profiler/build.gradle
+++ b/profiler/build.gradle
@@ -6,6 +6,11 @@ apply from: "$rootDir/gradle/shadow.gradle"
 
 def relocatePackages = ext.relocatePackages
 
+compileJava {
+    sourceCompatibility = 8
+    targetCompatibility = 8
+}
+
 dependencies {
     compileOnly("io.opentelemetry:opentelemetry-sdk:${versions.opentelemetry}")
     compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:${versions.opentelemetryAlpha}")
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -23,7 +23,6 @@ import com.google.auto.service.AutoService;
 import com.splunk.opentelemetry.profiler.util.HelpfulExecutors;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.javaagent.spi.ComponentInstaller;
-
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -16,7 +16,9 @@
 
 package com.splunk.opentelemetry.profiler;
 
-import static com.splunk.opentelemetry.profiler.Configuration.*;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_ENABLE_PROFILER;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_PROFILER_DIRECTORY;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_RECORDING_DURATION_SECONDS;
 import static com.splunk.opentelemetry.profiler.util.HelpfulExecutors.logUncaught;
 
 import com.google.auto.service.AutoService;
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -16,14 +16,16 @@
 
 package com.splunk.opentelemetry.profiler;
 
-import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_ENABLE_PROFILER;
-import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_RECORDING_DURATION_SECONDS;
+import static com.splunk.opentelemetry.profiler.Configuration.*;
 import static com.splunk.opentelemetry.profiler.util.HelpfulExecutors.logUncaught;
 
 import com.google.auto.service.AutoService;
 import com.splunk.opentelemetry.profiler.util.HelpfulExecutors;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.javaagent.spi.ComponentInstaller;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.concurrent.ExecutorService;
 import org.slf4j.Logger;
@@ -56,11 +58,13 @@ public class JfrActivator implements ComponentInstaller {
     Duration recordingDuration = Duration.ofSeconds(Integer.parseInt(recordingDurationStr));
     RecordingEscapeHatch recordingEscapeHatch = new RecordingEscapeHatch();
     JfrSettingsReader settingsReader = new JfrSettingsReader();
+    Path outputDir = Paths.get(config.getProperty(CONFIG_KEY_PROFILER_DIRECTORY, "."));
     JfrRecorder recorder =
         JfrRecorder.builder()
             .settingsReader(settingsReader)
             .maxAgeDuration(recordingDuration.multipliedBy(10))
             .jfr(JFR.instance)
+            .outputDir(outputDir)
             .build();
     RecordingSequencer sequencer =
         RecordingSequencer.builder()
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
@@ -23,7 +23,9 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
-import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
 import jdk.jfr.Recording;
 import jdk.jfr.RecordingState;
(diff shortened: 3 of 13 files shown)


profiler/build.gradle
```diff
@@ -6,6 +6,11 @@ apply from: "$rootDir/gradle/shadow.gradle"
 
 def relocatePackages = ext.relocatePackages
 
+compileJava {
+    sourceCompatibility = 8
+    targetCompatibility = 8
```
breedx-splk, line 11: I know that we use `options.release.set(8)` everywhere else, but I could not get gradle to respect the jfr APIs in java 8 with that.  This seems to work, though.

testing/profiler-tests/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+}
+
+tasks.register('copyProfilingAgentJar', Copy) {
+  from "../../agent/build/libs/"
+  include "splunk-otel-javaagent-*-all.jar"
+  into "build"
+}
+
+copyProfilingAgentJar.dependsOn(':agent:assemble')
+test.dependsOn(copyProfilingAgentJar)
```
mateuszrzeszutek, line 16: WDYT about using the smoke-tests way of getting the agent jar?

```groovy
tasks.test {
  def shadowTask = project(":agent").tasks.named("shadowJar").get()
  inputs.files(layout.files(shadowTask))

  doFirst {
    jvmArgs("-Dio.opentelemetry.smoketest.agent.shadowJar.path=${shadowTask.archiveFile.get()}")
  }
}
```

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
```diff
@@ -0,0 +1,105 @@
...
+            .withCopyFileToContainer(agentJar, "/app/javaagent.jar")
+            .withCreateContainerCmdModifier(cmd -> cmd.withEntrypoint("java"))
+            .withCommand(
+                "-javaagent:/app/javaagent.jar",
+                "-Dotel.javaagent.debug=true",
+                "-Dsplunk.profiler.enabled=true",
+                "-Dsplunk.profiler.directory=/app/jfr",
+                "-jar",
+                "/app/spring-petclinic-rest.jar")
+            .withFileSystemBind("build/test/output", "/app/jfr", BindMode.READ_WRITE)
```
mateuszrzeszutek, line 56: Hmm, if you put this code in a `@BeforeAll` you'd be able to use JUnit's `@TempDir` instead

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
```diff
@@ -0,0 +1,105 @@
...
+      return null;
+    }
+  }
+
+  @Test
+  void ensureJfrFilesCreated() throws Exception {
+    Path outputDir = Path.of("build/test/output");
+    Files.createDirectories(outputDir);
+    petclinic.start();
+    Wait.forHttp("/petclinic/api/vets");
```
mateuszrzeszutek, line 81: Is it needed? You're already waiting for it in the initialization method

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
```diff
@@ -0,0 +1,105 @@
...
+    System.out.println("Petclinic has been started.");
+
+    Instant start = Instant.now();
+    AtomicBoolean done = new AtomicBoolean(false);
+    while (!done.get()) {
+
+      System.out.println("Opening dir to look for jfr files...");
+      DirectoryStream<Path> dirStream = Files.newDirectoryStream(outputDir);
```
mateuszrzeszutek, line 89: `DirectoryStream` is `Closeable` -- I learned this the hard way with file descriptor OOMs on prod 😂

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+      dirStream.forEach(
+          item -> {
+            System.out.println("Found " + item);
+            if (item.toFile().isFile() && item.getFileName().toString().endsWith(".jfr")) {
```
mateuszrzeszutek, line 94:
```suggestion
            if (Files.isRegularFile(item) && item.getFileName().toString().endsWith(".jfr")) {
```

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
```diff
@@ -0,0 +1,105 @@
...
+              done.set(true);
+            }
+          });
+      if (Duration.between(start, Instant.now()).toSeconds() > 60) {
+        petclinic.stop();
+        fail("No output within time.");
+      }
+      TimeUnit.SECONDS.sleep(1);
```
mateuszrzeszutek, line 102: You could use `Awaitility` instead of polling manually; check out `CommonsDbcp2InstrumentationTest` for an example.

mateuszrzeszutek:
> The smoketest has to build a spring petclinic rest image because the one in dockerhub is 2 years old and doesn't have a java 8 version with JFR. Because of that, it's slower than I would prefer...so I'm inclined to file an issue to address that at a later time (maybe we can publish a base image like we do elsewhere).

You could add a separate project that published the image to GHCR, similar to smoke-tests in the upstream project.

Or, alternatively, just add a separate GHA job for profiler tests and exclude them from the main build - it won't make them run faster but should mitigate the issue a bit.

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
breedx-splk: Ah yeah, good catch.  I wasn't sure which one worked, but just confirmed this isn't needed here.  👍🏻

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
breedx-splk: THANK YOU!  😬

testing/profiler-tests/build.gradle
breedx-splk: That's cool, I can use that for consistency.  How do you ensure that the agent jar is built first tho?

breedx-splk: nm, separate concern.  I love it, ended up working out great!

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
breedx-splk: Yeah that's really nice, thanks for the tip!

breedx-splk:
Ready for another look.  I think since this is not yet hooked up to automation I will create an issue to track that and also an issue to build/publish a base image.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
```diff
@@ -63,7 +66,8 @@ Recording newRecording() {
 
   public void flushSnapshot() {
     try (Recording snap = jfr.takeSnapshot()) {
-      Path path = Path.of(Instant.now().toString() + ".jfr");
+      Path file = Paths.get(Instant.now().toString() + ".jfr");
```
jkwatson, line 69: What does `Instant.now().toString()` look like, anyway? Feels like it would be better to not use a `toString()` method for generating filenames, since there's no guarantee that it will generate usable filenames on misc. filesystems. For instance, what if the `toString()` changes to look like `06/03/2021`? Seems like that would be bad.

laurit: Could use `File.createTempFile`

breedx-splk: That's a good call.  The name isn't that important, but it's nice to have them be sortable and also human readable.  Lemme find something better.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -16,14 +16,15 @@
 
 package com.splunk.opentelemetry.profiler;
 
-import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_ENABLE_PROFILER;
-import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_RECORDING_DURATION_SECONDS;
+import static com.splunk.opentelemetry.profiler.Configuration.*;
```
laurit, line 19: at least in otel wildcard imports seem to be forbidden

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
```diff
@@ -66,7 +68,10 @@ Recording newRecording() {
...
   public void flushSnapshot() {
     try (Recording snap = jfr.takeSnapshot()) {
-      Path file = Paths.get(Instant.now().toString() + ".jfr");
+      String prefix =
+          DateTimeFormatter.ISO_DATE_TIME.format(
+              LocalDateTime.now().truncatedTo(ChronoUnit.SECONDS));
+      Path file = Paths.get(prefix + ".jfr");
       Path path = outputDir.resolve(file);
       logger.debug("Flushing a JFR snapshot: {}", path);
       snap.dump(path);
```
jkwatson, line 77: do we worry about the filesystem filling up ? Or not even being writable?

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
breedx-splk: 🥱 The IDE did that.  I can change it back.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
breedx-splk: Not yet, no.  There will be a component soon that deletes the files after they are read and there will also be a cleanup shutdown hook.

jkwatson (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrRecorder.java
breedx-splk: If the filesystem isn't writable, this won't work.  I suppose a future effort would be to use the streaming APIs for later versions of java if available.

PR #277, opened 2021-06-04
Introduce JfrPathHandler and clean up files

breedx-splk:
This establishes the foundation for the processing chain via the `JfrPathHandler`.  It also tracks files as they are created and deletes them after processing.  We also register a shutdown hook to not leave files behind.

A new config option `splunk.profiler.keepfiles` is available to leave the JFR files behind on disk if desired (this is useful for later development/comparison/analysis).

### DIFF
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -10,3 +10,4 @@ It should be considered experimental and is completely unsupported.
 |`splunk.profiler.enabled`            | false   | set to true to enable the profiler   |
 |`splunk.profiler.directory`          | "."     | location of jfr files                |
 |`splunk.profiler.recording.duration` | 20      | number of seconds per recording unit |
+|`splunk.profiler.keepfiles`          | false   | leave JFR files on disk id `true`    |
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -10,4 +10,4 @@ It should be considered experimental and is completely unsupported.
 |`splunk.profiler.enabled`            | false   | set to true to enable the profiler   |
 |`splunk.profiler.directory`          | "."     | location of jfr files                |
 |`splunk.profiler.recording.duration` | 20      | number of seconds per recording unit |
-|`splunk.profiler.keepfiles`          | false   | leave JFR files on disk id `true`    |
+|`splunk.profiler.keep-files`         | false   | leave JFR files on disk id `true`    |
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrDirCleanup.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrDirCleanup.java
@@ -30,7 +30,7 @@ import org.slf4j.LoggerFactory;
  * This class is responsible for cleaning up jfr files that might not have otherwise been cleaned
  * up. It can register a VM shutdown hook to remove its list of pending files at shutdown.
  */
-public class JfrDirCleanup {
+class JfrDirCleanup {
 
   private static final Logger logger = LoggerFactory.getLogger(JfrDirCleanup.class);
   private final Consumer<Path> fileDeleter;
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrDirCleanup.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrDirCleanup.java
@@ -53,14 +53,13 @@ public class JfrDirCleanup {
         .keySet()
         .forEach(
             path -> {
-              logger.warn("Shutdown :: JfrDirCleanup deleting " + path);
+              logger.warn("Shutdown :: JfrDirCleanup deleting {}", path);
               fileDeleter.accept(path);
             });
   }
 
   public void registerShutdownHook() {
     getRuntime().addShutdownHook(new Thread(logUncaught(this::cleanUp)));
-    ;
   }
 
   @VisibleForTesting
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrDirCleanup.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrDirCleanup.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler;
+
+import static com.splunk.opentelemetry.profiler.util.HelpfulExecutors.logUncaught;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is responsible for cleaning up jfr files that might not have otherwise been cleaned
+ * up. It can register a VM shutdown hook to remove its list of pending files at shutdown.
+ */
+public class JfrDirCleanup {
+
+  private static final Logger logger = LoggerFactory.getLogger(JfrDirCleanup.class);
+  private final Consumer<Path> fileDeleter;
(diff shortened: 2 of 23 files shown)


profiler/README.md
```diff
@@ -10,3 +10,4 @@ It should be considered experimental and is completely unsupported.
 |`splunk.profiler.enabled`            | false   | set to true to enable the profiler   |
 |`splunk.profiler.directory`          | "."     | location of jfr files                |
 |`splunk.profiler.recording.duration` | 20      | number of seconds per recording unit |
+|`splunk.profiler.keepfiles`          | false   | leave JFR files on disk id `true`    |
```
mateuszrzeszutek, line 13: Nit: why not `keep-files`?

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/FileDeleter.java
```diff
@@ -0,0 +1,45 @@
...
+        Files.delete(path);
+      } catch (IOException e) {
+        logger.warn("Could not delete: " + path, e);
+      }
+    };
+  }
+
+  static FileDeleter noopFileDeleter() {
+    return path -> {
+      logger.warn("Leaving " + path + " on disk.");
```
mateuszrzeszutek, line 42: Consider using slf4j placeholders:

```suggestion
      logger.warn("Leaving {} on disk.", path);
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/FileDeleter.java
```diff
@@ -0,0 +1,45 @@
...
+public interface FileDeleter extends Consumer<Path> {
+
+  Logger logger = LoggerFactory.getLogger(FileDeleter.class);
+
+  static FileDeleter newDeleter() {
+    return path -> {
+      try {
+        Files.delete(path);
+      } catch (IOException e) {
+        logger.warn("Could not delete: " + path, e);
```
mateuszrzeszutek, line 35:
```suggestion
        logger.warn("Could not delete: {}", path, e);
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrDirCleanup.java
```diff
@@ -0,0 +1,70 @@
...
+        .forEach(
+            path -> {
+              logger.warn("Shutdown :: JfrDirCleanup deleting " + path);
+              fileDeleter.accept(path);
+            });
+  }
+
+  public void registerShutdownHook() {
+    getRuntime().addShutdownHook(new Thread(logUncaught(this::cleanUp)));
+    ;
```
mateuszrzeszutek, line 63:
```suggestion
```
?

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrDirCleanup.java
```diff
@@ -0,0 +1,70 @@
...
+  public void recordingDeleted(Path path) {
+    pending.remove(path);
+  }
+
+  public void cleanUp() {
+    pending
+        .keySet()
+        .forEach(
+            path -> {
+              logger.warn("Shutdown :: JfrDirCleanup deleting " + path);
```
mateuszrzeszutek, line 56:
```suggestion
              logger.warn("Shutdown :: JfrDirCleanup deleting {}", path);
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrPathHandler.java
```diff
@@ -0,0 +1,81 @@
...
+import java.nio.file.Path;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+import jdk.jfr.consumer.RecordedEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Responsible for processing a single jfr file snapshot. It streams events from the
+ * RecordedEventStream into the processing chain and, once complete, calls the onComplete callback.
```
mateuszrzeszutek, line 28: Shouldn't it be `onFileFinished` instead?

```suggestion
 * RecordedEventStream into the processing chain and, once complete, calls the onFileFinished callback.
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrPathHandler.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public JfrPathHandler(Builder builder) {
+    this.eventProcessingChain = builder.eventProcessingChain;
+    this.onFileFinished = builder.onFileFinished;
+    this.recordedEventStreamFactory = builder.recordedEventStreamFactory;
+  }
+
+  @Override
+  public void accept(Path path) {
+    logger.info("New jfr file detected: " + path);
```
mateuszrzeszutek, line 45:
```suggestion
    logger.info("New jfr file detected: {}", path);
```

profiler/src/test/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFileTest.java
```diff
@@ -0,0 +1,59 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler;
+
+import static org.junit.jupiter.api.Assertions.*;
```
mateuszrzeszutek, line 19: Looks like 66 wasn't enough 😂

jkwatson:
Do all these classes need to be `public`? Seems like it's all contained to a single package, so it would probably be better to restrict access until we need to expose things to users.

profiler/README.md
breedx-splk: I dunno, I feel like the hyphen introduces a small parsing risk, but I'm not too paranoid about it.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrPathHandler.java
breedx-splk: Yup, I think a refactor didn't propagate into the comments.  Good catch.

profiler/src/test/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFileTest.java
breedx-splk: OMG!

profiler/README.md
mateuszrzeszutek: Not really, the upstream agent fully supports this: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1414#issuecomment-733181398
We've basically copied the property naming scheme from spring

breedx-splk:
> Do all these classes need to be `public`? Seems like it's all contained to a single package, so it would probably be better to restrict access until we need to expose things to users.

As the package grows, it can make reorganizing a bit more frustrating down the road...but I can try it out in the name of public safety.

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/FileDeleter.java
breedx-splk: Hmmm....I made this change, but I'm not confident that it will work.  The IDE says there are these overloads, none matching treats the exception as a Throwable, so I worry it would just error out trying to find a placeholder for that object?  Do you know if it works?
image

jkwatson: IIRC, there's an overload with the throwable in the first slot to account for this case. But, I might be remembering some other library with a similar pattern.

breedx-splk: That would make sense...but I'm not seeing it here.

profiler/README.md
breedx-splk: Ok, I can change it.  👍🏻

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/FileDeleter.java
mateuszrzeszutek: It will work -- it's kinda non-obvious when you look at the signature, but slf4j checks the type of the last argument and if it's a `Throwable` then it adds the stack trace.

breedx-splk: Hah, ok cool then I'll leave it. I'm glad you knew that,  heh.

mateuszrzeszutek: I've checked it at least few times over the last few years 😂

mateuszrzeszutek (approved):

PR #280, opened 2021-06-07
Build out the recording escape hatch

breedx-splk:
The desire is to no longer record to files when the disk is low on space or if the number of files are backing up too much.  The `keep-files` configuration is considered a rare override that allows recording to continue even when files are backed up (also useful for development).

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingEscapeHatch.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingEscapeHatch.java
@@ -16,9 +16,148 @@
 
 package com.splunk.opentelemetry.profiler;
 
+import java.io.IOException;
+import java.nio.file.FileStore;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.Duration;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 class RecordingEscapeHatch {
 
+  private static final long MIN_FREE_SPACE_BYTES = 100 * 1024 * 1024; // 100MiB
+  private static final Logger logger = LoggerFactory.getLogger(RecordingEscapeHatch.class);
+  private static final Duration MAX_PENDING_DURATION = Duration.ofMinutes(5);
+
+  private final RecordingFileNamingConvention namingConvention;
+  private final boolean configKeepsFilesOnDisk;
+  private final long maxFileCount;
+  private final FilesShim filesShim;
+
+  RecordingEscapeHatch(Builder builder) {
+    this.namingConvention = builder.namingConvention;
+    this.configKeepsFilesOnDisk = builder.configKeepsFilesOnDisk;
+    this.filesShim = builder.filesShim;
+    this.maxFileCount = MAX_PENDING_DURATION.toMillis() / builder.recordingDuration.toMillis();
+  }
+
   public boolean jfrCanContinue() {
-    return true;
+    boolean result = freeSpaceIsOk() && notFileBacklogged();
+    if (!result) {
+      logger.warn("** THIS WILL RESULT IN LOSS OF PROFILING DATA **");
+    }
+    return result;
+  }
+
+  private boolean freeSpaceIsOk() {
+    try {
+      Path outputPath = namingConvention.getOutputPath();
+      FileStore store = filesShim.getFileStore(outputPath);
+      long usableSpace = store.getUsableSpace();
+      boolean result = usableSpace > MIN_FREE_SPACE_BYTES;
+      if (!result) {
+        logger.warn(
+            "** NOT STARTING RECORDING, only "
+                + usableSpace
+                + " bytes free, require "
+                + MIN_FREE_SPACE_BYTES);
+      }
+      return result;
+    } catch (IOException e) {
+      logger.error("Could not check free space, assuming everything is bad", e);
+      return false;
+    }
+  }
+
+  /**
+   * Backlogged means that there are too many recent jfr files sitting on disk, and we think we are
+   * falling behind. We will never be backlogged if the user has specified
+   * -Dsplunk.profiler.keep-files=true
+   */
+  private boolean notFileBacklogged() {
+    if (configKeepsFilesOnDisk) {
+      return true;
+    }
+    try {
+      return pendingFileCount() < maxFileCount;
+    } catch (IOException e) {
+      logger.warn("Error listing files in output directory, assuming everything is bad");
+      return false;
+    }
+  }
+
+  /** Returns the number of jfr files in the output directory that match our pattern */
+  private long pendingFileCount() throws IOException {
+    Path outputPath = namingConvention.getOutputPath();
+    return filesShim
+        .list(outputPath)
+        .filter(filesShim::isRegularFile)
+        .filter(namingConvention::matches)
+        .count();
+  }
+
+  static Builder builder() {
+    return new Builder();
+  }
+
+  // To help with testing
+  interface FilesShim {
+    FilesShim DEFAULT =
+        new FilesShim() {
+          @Override
+          public FileStore getFileStore(Path path) throws IOException {
+            return Files.getFileStore(path);
+          }
+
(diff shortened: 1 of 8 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingEscapeHatch.java
```diff
@@ -16,9 +16,148 @@
...
+      logger.warn("** THIS WILL RESULT IN LOSS OF PROFILING DATA **");
+    }
+    return result;
+  }
+
+  private boolean freeSpaceIsOk() {
+    try {
+      Path outputPath = namingConvention.getOutputPath();
+      FileStore store = filesShim.getFileStore(outputPath);
+      long usableSpace = store.getUsableSpace();
```
jkwatson, line 58: Any idea what these APIs do on docker? Is it going to be returning something from the host, or is it for-sure just the container?

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingFileNamingConvention.java
```diff
@@ -0,0 +1,68 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
+
+class RecordingFileNamingConvention {
+
+  private static final String PREFIX = "otel-profiler";
+  private final Path outputDir;
+
+  public RecordingFileNamingConvention(Path outputDir) {
```
mateuszrzeszutek, line 30: Nit: the class is package-private, public modifiers on methods don't mean anything

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingFileNamingConvention.java
```diff
@@ -0,0 +1,68 @@
...
+  /** Determines if the path represents a file that we would have recorded to. */
+  public boolean matches(Path path) {
+    return outputDir.equals(path.getParent()) && filenameMatches(path);
+  }
+
+  private boolean filenameMatches(Path path) {
+    String filename = path.getFileName().toFile().getName();
+    // ISO_DATE_TIME format is like 2021-12-03T10:15:30
+    return filename.startsWith(PREFIX)
+        && filename.matches("^" + PREFIX + "-\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.jfr$");
```
mateuszrzeszutek, line 62: Consider using a precompiled `Pattern`

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingFileNamingConvention.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /** Determines if the path represents a file that we would have recorded to. */
+  public boolean matches(Path path) {
+    return outputDir.equals(path.getParent()) && filenameMatches(path);
+  }
+
+  private boolean filenameMatches(Path path) {
+    String filename = path.getFileName().toFile().getName();
+    // ISO_DATE_TIME format is like 2021-12-03T10:15:30
+    return filename.startsWith(PREFIX)
```
mateuszrzeszutek, line 61: The regex pattern already checks that the file name starts with `PREFIX`

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingFileNamingConvention.java
```diff
@@ -0,0 +1,68 @@
...
+    return Paths.get(PREFIX + "-" + timestamp + ".jfr");
+  }
+
+  /** Determines if the path represents a file that we would have recorded to. */
+  public boolean matches(Path path) {
+    return outputDir.equals(path.getParent()) && filenameMatches(path);
+  }
+
+  private boolean filenameMatches(Path path) {
+    String filename = path.getFileName().toFile().getName();
```
mateuszrzeszutek, line 59: Nit: `toString()` would probably work:
```suggestion
    String filename = path.getFileName().toString();
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingEscapeHatch.java
```diff
@@ -16,9 +16,148 @@
...
+      logger.warn("Error listing files in output directory, assuming everything is bad");
+      return false;
+    }
+  }
+
+  /** Returns the number of jfr files in the output directory that match our pattern */
+  private long pendingFileCount() throws IOException {
+    Path outputPath = namingConvention.getOutputPath();
+    return filesShim
+        .list(outputPath)
```
mateuszrzeszutek, line 95: Stream returned by `Files.list()` should be closed

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingEscapeHatch.java
```diff
@@ -16,9 +16,148 @@
...
+      Path outputPath = namingConvention.getOutputPath();
+      FileStore store = filesShim.getFileStore(outputPath);
+      long usableSpace = store.getUsableSpace();
+      boolean result = usableSpace > MIN_FREE_SPACE_BYTES;
+      if (!result) {
+        logger.warn(
+            "** NOT STARTING RECORDING, only "
+                + usableSpace
+                + " bytes free, require "
+                + MIN_FREE_SPACE_BYTES);
```
mateuszrzeszutek, line 65:
```suggestion
        logger.warn(
            "** NOT STARTING RECORDING, only {} bytes free, require {}", usableSpace,  MIN_FREE_SPACE_BYTES);
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingFileNamingConvention.java
breedx-splk: I hear ya.  I guess in general I'm not a huge fan of restricting access, and I think that pulling `public` off of methods muddies the object shape / user contract.  It's much harder to quickly look and see "what does this object provide" if everything is just package access, and it's made worse when some methods are package access just to facilitate testing.

I understand that this is the convention in this repo (and upstream) so I'll try and be better about keeping consistent.

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingEscapeHatch.java
breedx-splk: I would expect them to do the same thing in a container.  Containers have a certain amount of space available to them, and volume mounts could have been created to map onto the host system.

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingFileNamingConvention.java
jkwatson: IMO, we should avoid using `toString()` unless it is documented to be stable on the API we are calling. Is this one?

mateuszrzeszutek: Good point - it isn't documented to be unstable in the JDK, the Javadoc just says that it `Returns the string representation of this path`. It is used to convert to `File` in `toFile()` though.

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingFileNamingConvention.java
```diff
@@ -0,0 +1,70 @@
...
+  boolean matches(Path path) {
+    return outputDir.equals(path.getParent()) && filenameMatches(path);
+  }
+
+  private boolean filenameMatches(Path path) {
+    String filename = path.getFileName().toString();
+    return filenamePattern.matcher(filename).matches();
+  }
+
+  Path getOutputPath() {
```
jkwatson, line 67: since this just the directory, would it make more sense to name the method `getOutputDir` ?

PR #281, opened 2021-06-08
add context storage to generate jfr scope tracking events

breedx-splk:
If the profiler is enabled, we configure the SDK with a context storage that will generate custom JFR events for scope start/stops.  This allows us to later associate stack with spans.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrContextStorage.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrContextStorage.java
@@ -37,18 +37,18 @@ class JfrContextStorage implements ContextStorage {
     this(delegate, Span::fromContext, JfrContextStorage::newEvent);
   }
 
-  static ContextAttached newEvent(SpanContext spanContext, byte direction) {
-    return new ContextAttached(spanContext.getTraceId(), spanContext.getSpanId(), direction);
-  }
-
   @VisibleForTesting
   JfrContextStorage(
-      ContextStorage delegate,
-      Function<Context, Span> spanFromContext,
-      BiFunction<SpanContext, Byte, ContextAttached> newEvent) {
+          ContextStorage delegate,
+          Function<Context, Span> spanFromContext,
+          BiFunction<SpanContext, Byte, ContextAttached> newEvent) {
     this.delegate = delegate;
     this.spanFromContext = spanFromContext;
     this.newEvent = newEvent;
+  }
+
+  static ContextAttached newEvent(SpanContext spanContext, byte direction) {
+    return new ContextAttached(spanContext.getTraceId(), spanContext.getSpanId(), direction);
   }
 
   @Override
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrContextStorage.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrContextStorage.java
@@ -39,9 +39,9 @@ class JfrContextStorage implements ContextStorage {
 
   @VisibleForTesting
   JfrContextStorage(
-          ContextStorage delegate,
-          Function<Context, Span> spanFromContext,
-          BiFunction<SpanContext, Byte, ContextAttached> newEvent) {
+      ContextStorage delegate,
+      Function<Context, Span> spanFromContext,
+      BiFunction<SpanContext, Byte, ContextAttached> newEvent) {
     this.delegate = delegate;
     this.spanFromContext = spanFromContext;
     this.newEvent = newEvent;
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrContextStorage.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrContextStorage.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.splunk.opentelemetry.profiler.events.ContextAttached;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanContext;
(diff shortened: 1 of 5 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrContextStorage.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private final ContextStorage delegate;
+  private final Function<Context, Span> spanFromContext;
+  private final BiFunction<SpanContext, Byte, ContextAttached> newEvent;
+
+  JfrContextStorage(ContextStorage delegate) {
+    this(delegate, Span::fromContext, JfrContextStorage::newEvent);
+  }
+
+  static ContextAttached newEvent(SpanContext spanContext, byte direction) {
```
iNikem, line 40: Please have all constructors together

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+@ExtendWith(MockitoExtension.class)
+class JfrContextStorageTest {
+
+  @Mock Span span;
+  @Mock ContextStorage delegate;
+  @Mock Context newContext;
+  @Mock Scope delegatedScope;
+  @Mock Function<Context, Span> spanFromContext;
+  @Mock SpanContext spanContext;
```
iNikem, line 49: Do you actually need all these mocks? Context API is totally usable by itself, no?

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
```diff
@@ -0,0 +1,127 @@
...
+    verify(delegatedScope, never()).close();
+
+    resultScope.close();
+    verify(outEvent).begin();
+    verify(outEvent).commit();
+    verify(delegatedScope).close();
+  }
+
+  @Test
+  void testAttachInvalidContext() {
```
iNikem, line 99: This test definitely can have more descriptive name :)

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
```diff
@@ -0,0 +1,127 @@
...
+    when(span.getSpanContext()).thenReturn(spanContext);
+    when(spanContext.isValid()).thenReturn(false);
+
+    JfrContextStorage contextStorage = new JfrContextStorage(delegate, spanFromContext, newEvent);
+
+    contextStorage.attach(newContext);
+  }
+
+  @Test
+  void testCurrent() {
```
iNikem, line 117: This test definitely can have more descriptive name :)

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
breedx-splk: * span - mocked in order to control whether or not the span context is in/valid
* delegate - mocked in order to return the delegated scope (below) so that we can assert that close is/isn't called
* newContext - the implementation of the Context interface is the ArrayBasedContext which has package access and cannot be instantiated, hence the mock
* delegatedScope - mocked so that we can verify that close() is/isn't called
* spanFromContext - this could actually be a real implementation but would probably add some code.  is there a specific problem with this being a mock?
* spanContext - mocked so that we can control whether or not the context is/isn't valid.

I don't know how to use the existing Context API by itself to drive this behavior without mocks.

jkwatson: As long as you're not doing `context.makeCurrent()` or anything to do with the scope/storage, you should just be able to work with the Context objects as POJOs. So, you can just do `Context.root().with(span)` etc.

jkwatson: And `SpanContext` has a `getInvalid()` method on it to get an invalid one without a mock, as well.

breedx-splk: Ok, I was able to simplify things a little. Thanks.

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
```diff
@@ -0,0 +1,141 @@
...
+    span = Span.wrap(spanContext);
+    newContext = Context.root().with(span);
+  }
+
+  @Test
+  void testNewEvent() {
+    String traceId = "nter108y";
+    String spanId = "abc123";
+
+    SpanContext context = mock(SpanContext.class);
```
iNikem, line 71: Why still using mocks here?

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
```diff
@@ -0,0 +1,141 @@
...
+    assertEquals(traceId, result.getTraceId());
+    assertEquals(spanId, result.getSpanId());
+    assertEquals(ContextAttached.OUT, result.getDirection());
+  }
+
+  @Test
+  void testAttachLifecycle() {
+    ContextAttached inEvent = mock(ContextAttached.class);
+    ContextAttached outEvent = mock(ContextAttached.class);
+    BiFunction<SpanContext, Byte, ContextAttached> newEvent = mock(BiFunction.class);
```
iNikem, line 86: Do you actually need to mock `BiFunction`? You can just implement it however you like, cannot you?

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
breedx-splk: I missed it....will change.  👍🏻

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
breedx-splk: You're right..._however_ I think that the fluent mocking API actually reads cleaner in this case.  The function is called twice and returns different events, so it would require putting logic in the body.

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrContextStorageTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    JfrContextStorage contextStorage = new JfrContextStorage(delegate, newEvent);
+
+    contextStorage.attach(newContext);
+  }
+
+  @Test
+  void testCurrentSimplyDelegates() {
+    Context expected = mock(Context.class);
+    ContextStorage delegate = mock(JfrContextStorage.class);
```
mateuszrzeszutek, line 125: Nit:
```suggestion
    ContextStorage delegate = mock(ContextStorage.class);
```

mateuszrzeszutek (approved):

PR #282, opened 2021-06-08
Ensure that the file is closed at stream end

breedx-splk:
Could possibly have been holding open resources.  Not sure how this could impact the file deletion, but on some platforms maybe this could be a problem and it just seems better to close it.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
@@ -42,24 +42,26 @@ class BasicJfrRecordingFile implements RecordedEventStream {
   public Stream<RecordedEvent> open(Path path) {
     RecordingFile file = jfr.openRecordingFile(path);
     return StreamSupport.stream(
-        new Spliterators.AbstractSpliterator<RecordedEvent>(Long.MAX_VALUE, Spliterator.ORDERED) {
-          public boolean tryAdvance(Consumer<? super RecordedEvent> action) {
-            if (file.hasMoreEvents()) {
-              action.accept(jfr.readEvent(file, path));
-              return true;
-            }
-            closeSafely(file);
-            return false;
-          }
+            new Spliterators.AbstractSpliterator<RecordedEvent>(
+                Long.MAX_VALUE, Spliterator.ORDERED) {
+              public boolean tryAdvance(Consumer<? super RecordedEvent> action) {
+                if (file.hasMoreEvents()) {
+                  action.accept(jfr.readEvent(file, path));
+                  return true;
+                }
+                closeSafely(file);
+                return false;
+              }
 
-          public void forEachRemaining(Consumer<? super RecordedEvent> action) {
-            while (file.hasMoreEvents()) {
-              action.accept(jfr.readEvent(file, path));
-            }
-            closeSafely(file);
-          }
-        },
-        false);
+              public void forEachRemaining(Consumer<? super RecordedEvent> action) {
+                while (file.hasMoreEvents()) {
+                  action.accept(jfr.readEvent(file, path));
+                }
+                closeSafely(file);
+              }
+            },
+            false)
+        .onClose(() -> closeSafely(file));
   }
 
   private void closeSafely(RecordingFile file) {
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
@@ -16,6 +16,7 @@
 
 package com.splunk.opentelemetry.profiler;
 
+import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Spliterator;
 import java.util.Spliterators;
@@ -24,9 +25,12 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import jdk.jfr.consumer.RecordedEvent;
 import jdk.jfr.consumer.RecordingFile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Simple/basic abstraction around a recording file. Can open and get a stream of events. */
 class BasicJfrRecordingFile implements RecordedEventStream {
+  private static final Logger logger = LoggerFactory.getLogger(BasicJfrRecordingFile.class);
 
   private final JFR jfr;
 
@@ -44,6 +48,7 @@ class BasicJfrRecordingFile implements RecordedEventStream {
               action.accept(jfr.readEvent(file, path));
               return true;
             }
+            closeSafely(file);
             return false;
           }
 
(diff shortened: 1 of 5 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
```diff
@@ -44,15 +48,25 @@ public boolean tryAdvance(Consumer<? super RecordedEvent> action) {
...
           }
         },
         false);
   }
+
+  private void closeSafely(RecordingFile file) {
+    try {
+      file.close();
+    } catch (IOException e) {
+      logger.warn("Error closing JFR file", e);
```
iNikem, line 69: Will an end-user see this WARN log? What should they do about it? Because they will be alarmed by that.

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
```diff
@@ -44,15 +48,25 @@ public boolean tryAdvance(Consumer<? super RecordedEvent> action) {
               action.accept(jfr.readEvent(file, path));
               return true;
             }
+            closeSafely(file);
```
iNikem, line 51: This will not be closed if this stream is never consumed. Should we open jfr file lazily? Should we require actually closing the stream and then close the file?

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
breedx-splk: If they look at the logs and have it configured for WARN or below, then yes.  What should they do about it? I dunno. It could help inform when looking into other problems? There are some wild failure modes that cause close to fail. Would you prefer that we silently eat it?

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
breedx-splk: I don't think there's really a use case for opening the stream and not consuming it fully.  But yes, if a user did chose to do that then they would be responsible for closing it. Should I make a comment to that end?

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
iNikem: For me "WARN" always means "not a problem right now but you should take a look and fix it anyway". User cannot do anything to actually remedy this WARN, can they?

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
iNikem: What about actually changing the API/implementation to prevent misuse?

mateuszrzeszutek: How about you add the close handler to the stream (`Stream#onClose(Runnable)`) and use try-with-resources on it?

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
breedx-splk: For me, WARN usually means "Hey something kinda bad happened that you probably should know about, but your app didn't necessarily hit an error".  If their filesystem got turned read only, for example, they could change it back to read/write to fix the problem.  So I I dunno, I think maybe yeah they can do something....but are you suggesting we make it a lower level then?

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
breedx-splk: I like the idea of adding an `Stream.onClose()` callback to close the underlying file, but that still pushes the responsibility out and requires the user to do more.  I guess I'll try it out and see how it feels...

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
```diff
@@ -42,24 +42,26 @@ public BasicJfrRecordingFile(JFR jfr) {
...
-        false);
+              public void forEachRemaining(Consumer<? super RecordedEvent> action) {
+                while (file.hasMoreEvents()) {
+                  action.accept(jfr.readEvent(file, path));
+                }
+                closeSafely(file);
+              }
+            },
+            false)
+        .onClose(() -> closeSafely(file));
```
iNikem, line 64: Second call to `closeSafely` will not generate `IOException`, will it?

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
iNikem: You convinced me :)

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
```diff
@@ -38,21 +42,33 @@ public BasicJfrRecordingFile(JFR jfr) {
...
+                }
+                closeSafely(file);
+                return false;
+              }
+
+              public void forEachRemaining(Consumer<? super RecordedEvent> action) {
+                while (file.hasMoreEvents()) {
+                  action.accept(jfr.readEvent(file, path));
+                }
+                closeSafely(file);
```
mateuszrzeszutek, line 60: Nit: you probably don't need to call `closeSafely()` here and in `tryAdvance()` now that you've added it as the close handler.

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
mateuszrzeszutek: > that still pushes the responsibility out and requires the user to do more.

Yeah, that's one of the drawbacks.
The worst part is that you don't really know if you're supposed to close the stream or not unless it's described in the Javadoc... I wish there was a `CloseableStream` instead

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
breedx-splk: Nope, I peeked at the implementation to make sure.

profiler/src/main/java/com/splunk/opentelemetry/profiler/BasicJfrRecordingFile.java
breedx-splk: The intention was to be helpful in the event that the user doesn't actually call close.  Best effort close() is called for you if you consume the entire stream.

Do you think it's confusing to have close in here and ask the caller to to also call close?  Want me to remove it?

mateuszrzeszutek: Hmm, I guess it's probably fine to leave it if doesn't throw on second call

mateuszrzeszutek (approved):

PR #288, opened 2021-06-11
look for specific events in the jfr file output

breedx-splk:
Adds some additional details in the profiling smoke test. We look for the synthetic `otel.ContextAttached` events in the JFR files now (instead of only checking for their existence).

### DIFF
--- a/testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
+++ b/testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
@@ -18,7 +18,9 @@ package com.splunk.opentelemetry;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import java.io.IOException;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
```diff
@@ -74,13 +82,57 @@ static void teardown() {
...
+        }
+      }
+    } catch (IOException e) {
+      return false;
+    }
+    return false;
+  }
+
+  private void generateSomeSpans() throws Exception {
+    System.out.println("Generating some spans...");
```
jkwatson, line 116: probably don't want printlns in here. log, maybe?

breedx-splk: I had hoped to see them in the build stdout but I think gradle prevents that from actually working.  Does using a logger help with that?  I just want some indication of how far the thing progressed before the monolithic pass/fail happens.  Open to ideas.

jkwatson: should matter logger or no. You have to run at --info to see either, correct?

breedx-splk: 🤷🏻 I dunno, probably.  I'm flexible, so if we prefer logs then I can change it.

mateuszrzeszutek (approved):

testing/profiler-tests/src/test/java/com/splunk/opentelemetry/ProfilerSmokeTest.java
jkwatson: I do prefer using a logger here, so if we can get that to work, then 👍🏼 . Otherwise, whatever gets things up and running if fine with me in test code. (do we expect to actually need this output, or was it just useful when getting things running initially?)

breedx-splk: It's super helpful if/when things break. Honestly, without the output it's just an opaque "test failed, good luck buddy".  At least with some output you can get an indication of how far it got before it failed.

PR #293, opened 2021-06-16
Don't set multiple copies of the same header

iNikem:


mateuszrzeszutek (approved):

kubawach (approved):
Yeah, I even recall similar changes in the upstream repo some time ago (set vs add).

PR #294, opened 2021-06-16
Logs processor

breedx-splk:
This builds up the foundation of very rudimentary logs support in otel. It includes a very simple data model and a LogsProcessor and LogsExporter interfaces. A simple `BatchingLogsProcessor` implements `LogsProcessor` and will pass batches of `LogEntry` to an exporter when a size or time is reached.

This also includes some adapters in order to convert from the otel data model (`LogEntry`) to the protobuf model.

### DIFF
--- a/profiler/build.gradle
+++ b/profiler/build.gradle
@@ -12,11 +12,13 @@ compileJava {
 }
 
 dependencies {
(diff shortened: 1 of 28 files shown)


breedx-splk:
Going to pull the protobuf stuff out to make this nicer/smaller.

PR #295, opened 2021-06-16
Add LogsProcessor

breedx-splk:
This builds up the foundation of very rudimentary logs support in otel. It includes a very simple data model and a LogsProcessor and LogsExporter interfaces. A simple BatchingLogsProcessor implements LogsProcessor and will pass batches of LogEntry to an exporter when a size or time is reached.

The next phase will add an exporter implementation.


profiler/src/main/java/com/splunk/opentelemetry/logs/LogsExporter.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.logs;
+
+import java.util.List;
+
+/** Sends a list of LogEntry objects somewhere else, often to the OpenTelemetry collector. */
+public interface LogsExporter {
+  void export(List<LogEntry> logs);
```
jkwatson, line 23: should this match the java SDK and return `CompletableResultCode`?

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+      doAction();
+    }
+  }
+
+  @Override
+  public void log(LogEntry log) {
+    synchronized (lock) {
+      batch.add(log);
+      if (batch.size() >= maxBatchSize) {
+        doAction();
```
jkwatson, line 71: I don't think this is a good idea...to continue to block while the action runs in the same thread. At least, if this were in the SDK, I think we'd definitely need to make 100% sure that this didn't block the hot path.

profiler/src/main/java/com/splunk/opentelemetry/logs/LogsExporter.java
breedx-splk: Yeah probably.  It's all stubbed for now, so maybe when implementing the actual impl would be a good time to make that change?

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
breedx-splk: Oh yeah interesting. For this particular profiling use case, we are definitely not in a hot path and we can back up a little during export no problem. I can definitely see wanting to background the actual action/work.  I can think of a couple of options here:

* Always run the action in a background thread.  It's probably an extra/unnecessary/extra thread in the profiling use case, but should be fine.
* Keep the BatchingLogsProcessor like it is now, but clarify that action can be invoked from the calling thread if capacity is reached...and then build a backgrounding action to suppose the other use case.

I suppose I'm torn here.  The first option is probably more conventional, but I hate to fire up yet another thread when exporters themselves may be async/background. Maybe the second idea is just harder to understand as well.

breedx-splk: Ok, I took a stab at doing the first thing and just running the action on a background thread.  Lemme know what you think!  🤘🏻

profiler/src/main/java/com/splunk/opentelemetry/profiler/StackToSpanLinkageProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Glue helper to turn a linked stack into an otel LogEntry instance and pass it to the logs
+ * processor.
+ */
+public class StackToSpanLinkageProcessor implements Consumer<StackToSpanLinkage> {
+  private final LogEntryCreator logEntryCreator;
+  private final BatchingLogsProcessor processor;
+
+  public StackToSpanLinkageProcessor(
+      LogEntryCreator logEntryCreator, BatchingLogsProcessor processor) {
```
mateuszrzeszutek, line 33: Consider using the interface, not the implementation:

```suggestion
  private final LogsProcessor processor;

  public StackToSpanLinkageProcessor(
      LogEntryCreator logEntryCreator, LogsProcessor processor) {
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/SpanLinkage.java
```diff
@@ -41,11 +43,11 @@ String getTraceId() {
   }
 
   RecordedThread getRecordedThread() {
-    return recordedThread;
+    return recordedEvent == null ? null : recordedEvent.getThread();
   }
 
   Long getThreadId() {
```
mateuszrzeszutek, line 49: This method can return a primitive long now:
```suggestion
  long getThreadId() {
```

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+    private int maxBatchSize = DEFAULT_MAX_BATCH_SIZE;
+    private Duration maxTimeBetweenBatches;
+    private Consumer<List<LogEntry>> batchAction = logs -> {};
+    private ScheduledExecutorService executorService =
+        HelpfulExecutors.newSingleThreadedScheduledExecutor("BatchingLogsProcessor action");
```
mateuszrzeszutek, line 102: Nit: consider creating that executor service in `BatchingLogsProcessor` only when the user didn't pass their own - this way you'll avoid creating an unused thread pool.

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+    private int maxBatchSize = DEFAULT_MAX_BATCH_SIZE;
+    private Duration maxTimeBetweenBatches;
+    private Consumer<List<LogEntry>> batchAction = logs -> {};
```
mateuszrzeszutek, line 100: Does it make sense to have a default value for this? Isn't this param pretty much mandatory?

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+    this.executorService = builder.executorService;
+    batch = new ArrayList<>(maxBatchSize);
+  }
+
+  public void start() {
+    synchronized (lock) {
+      if (watchdog != null) {
+        throw new IllegalStateException("Already running");
+      }
+      watchdog = new WatchdogTimer(maxTimeBetweenBatches, this::doAction, executorService);
```
mateuszrzeszutek, line 55: Now the watchdog adds another task to the executor every time it executes - would it make sense to have separate `doAction` and `asyncDoAction` methods? One for the watchdog, one for exporting asynchronously when the list fills up.

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+ * which ever comes first.
+ */
+public class BatchingLogsProcessor implements LogsProcessor {
+
+  private static final int DEFAULT_MAX_BATCH_SIZE = 250;
+
+  private final int maxBatchSize;
+  private final Duration maxTimeBetweenBatches;
+  private final List<LogEntry> batch;
+  private final Consumer<List<LogEntry>> batchAction;
```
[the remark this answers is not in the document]
breedx-splk: Wait, you want the looser interface below but the stricter one here?  Hmmmm....curious.  🤔 
Even though this is serving log export, there _could_ be some additional handling prior to export, so I dunno, keeping it looser felt more flexible/extensible...but it should be fine. 

Still curious about why stricter here but looser in `StackToSpanLinkageProcessor`?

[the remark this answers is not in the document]
mateuszrzeszutek: Oh, in the case of `StackToSpanLinkageProcessor` it was just a simple matter of using the interface, not the actual implementation of it (like `List` vs `ArrayList`).

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
breedx-splk: Not even a nit, this is a good call!  👍🏻

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/SpanLinkage.java
breedx-splk: Nah, the thread can still be null.

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
breedx-splk: Sure, I think that makes sense (a little clearer than the double submit).

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+
+  @Override
+  public void log(LogEntry log) {
+    synchronized (lock) {
+      batch.add(log);
+      if (batch.size() >= maxBatchSize) {
+        asyncDoExport();
+        watchdog.reset();
```
jkwatson, line 79: I wonder if this `reset()` is really necessary. Would it be all that bad if the timed export just kept ticking along on it's normal schedule?

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+      }
+      watchdog = new WatchdogTimer(maxTimeBetweenBatches, this::syncDoExport, executorService);
+      watchdog.start();
+    }
+  }
+
+  void stop() {
+    synchronized (lock) {
+      if (watchdog == null) {
+        throw new IllegalStateException("Not running");
```
jkwatson, line 65: if it's null, could we just no-op this, rather than throwing an exception? Seems like it shouldn't really be harmful.

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
breedx-splk: Not _necessary_ per se, but it changes the behavior, and I like the "max size or timeout, whichever comes first" idea. If you remove this `reset()` then you could get an export with a tiny batch immediately after a full batch, which is a little unexpected/odd.

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
breedx-splk: It's fair, but I just wanted to keep the state correct, and the exception seemed like a way to do that while keeping parity with the `start()` state check.

PR #309, opened 2021-06-21
Smoke tests base image

### LINKED ISSUE #276: Build and publish base image for profiling smoke tests
Originates in discussion in #274.

The docker image for the profiling smoke test takes a little while to come up.  We should pre-bake an image to help reduce this time (tests run faster).  Unfortunately, we can't just use the `springcommunity/spring-petclinic-rest` image from dockerhub because it is > 2 years old and doesn't have JFR support in that VM.

We can probably just publish our image to GHCR and use it in the test as the base image.

breedx-splk:
Hopefully resolves #276.

This creates a manually triggered GH action that will publish a base image for spring-petclinic-rest to ghcr.  This base image can then be used by the profiling smoke tests. 

Having this base image should greatly speed up build/test cycle time (largely spent in mvn downloading deps).

### DIFF
--- a/.github/publish-profiling-base-petclinic.yml
+++ b/.github/publish-profiling-base-petclinic.yml
@@ -1,0 +1,25 @@
+name: publish-profiling-base-petclinic-image
+on: workflow_dispatch
+jobs:
+  push_to_registry:
+    name: publish custom spring-petclinic base image for profiling
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    steps:
+      - name: check out the repo
+        uses: actions/checkout@v2
+      - uses: docker/setup-buildx-action@v1
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: push to gh packages
+        uses: docker/build-push-action@v2
+        with:
+          push: true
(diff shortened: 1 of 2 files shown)


jkwatson (approved):

.github/publish-profiling-base-petclinic.yml
```diff
@@ -0,0 +1,25 @@
+name: publish-profiling-base-petclinic-image
+on: workflow_dispatch
```
iNikem, line 2: Why only manually? You can trigger it whenever image's source code changes

.github/publish-profiling-base-petclinic.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  push_to_registry:
+    name: publish custom spring-petclinic base image for profiling
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    steps:
+      - name: check out the repo
+        uses: actions/checkout@v2
```
mateuszrzeszutek, line 12: Nit: you can set fetch depth to 0:
```suggestion
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
```

mateuszrzeszutek (approved):

.github/publish-profiling-base-petclinic.yml
breedx-splk: I don't expect it to change often, and having it triggered by changes is likely to produce a lot of unnecessary images and take up build time, no?

PR #310, opened 2021-06-22
whoops this was supposed to be in the workflows dir

breedx-splk:


mateuszrzeszutek (approved):
Didn't even notice that the last time 😂

PR #311, opened 2021-06-22
now that we have a base image, let's use it!

breedx-splk:


breedx-splk:
Here is the image, FYI: https://github.com/signalfx/splunk-otel-java/pkgs/container/splunk-otel-java%2Fprofiling-petclinic-base

mateuszrzeszutek (approved):

jkwatson (approved):

PR #312, opened 2021-06-22
No stacks in ContextAttached events.

breedx-splk:
We are not planning on using the stack data from this event, so let's omit it.

This also adds a test that confirms no stacks and enhances test code.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/events/ContextAttached.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/events/ContextAttached.java
@@ -22,10 +22,12 @@ import jdk.jfr.Category;
 import jdk.jfr.Event;
 import jdk.jfr.Label;
 import jdk.jfr.Name;
+import jdk.jfr.StackTrace;
 
 @Name(EVENT_NAME)
 @Label("otel context attached")
 @Category("opentelemetry")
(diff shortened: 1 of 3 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/events/ContextAttached.java
```diff
@@ -22,10 +22,12 @@
 import jdk.jfr.Event;
 import jdk.jfr.Label;
 import jdk.jfr.Name;
+import jdk.jfr.StackTrace;
 
 @Name(EVENT_NAME)
 @Label("otel context attached")
 @Category("opentelemetry")
+@StackTrace(false)
```
jkwatson, line 30: without the stack, why are we creating the event at all? I feel like I'm confused about what this is used for.

breedx-splk: It's used to know what thread is working on a span at a given instant in time, so that a stack from another event (like `jdk.ThreadDump`) can be linked to a span.

jkwatson: 👍🏼

jkwatson (approved):

PR #313, opened 2021-06-22
Let's add the profiler smoke test to catch regressions.

breedx-splk:
Added to the main CI but also for PRs.


jkwatson (approved):

breedx-splk:
Oh, I didn't want to run the profiler tests in the matrix...just after in docker.  I need to change this.

PR #315, opened 2021-06-23
allow logs ingest endpoint to be configured

breedx-splk:
Adds a new config property `splunk.profiler.logs-endpoint` that allows changing the collector endpoint for logs.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -141,33 +141,6 @@ public class JfrActivator implements AgentListener {
     sequencer.start();
     dirCleanup.registerShutdownHook();
   }
-  //
-  //  private static LogsExporter buildExporter(Config config) {
-  //    ResourceLogsAdapter adapter = buildResourceLogsAdapter();
-  //
-  //    OtlpLogsExporterBuilder builder =
-  //        OtlpLogsExporter.builder()
-  //            .setAdapter(adapter)
-  //            .addHeader("Extra-Content-Type", "otel-profiling-stacktraces");
-  //
-  //    String ingestUrl = config.getProperty(CONFIG_KEY_INGEST_URL);
-  //    if (ingestUrl != null) {
-  //      builder.setEndpoint(ingestUrl);
-  //    }
-  //    return builder.build();
-  //  }
-  //
-  //  private static ResourceLogsAdapter buildResourceLogsAdapter() {
-  //    Resource resource = OpenTelemetrySdkAutoConfiguration.getResource();
-  //    LogEntryAdapter logEntryAdapter = new LogEntryAdapter();
-  //    InstrumentationLibraryLogsAdapter instLibraryLogsAdapter =
-  //        InstrumentationLibraryLogsAdapter.builder()
-  //            .logEntryAdapter(logEntryAdapter)
-  //            .instrumentationName(OTEL_INSTRUMENTATION_NAME)
-  //            .instrumentationVersion(OTEL_INSTRUMENTATION_VERSION)
-  //            .build();
-  //    return new ResourceLogsAdapter(instLibraryLogsAdapter, resource);
-  //  }
 
   private Consumer<Path> buildFileDeleter(Config config) {
     if (keepFiles(config)) {
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -17,7 +17,6 @@
 package com.splunk.opentelemetry.profiler;
 
 import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_ENABLE_PROFILER;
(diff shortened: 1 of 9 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -151,13 +153,19 @@ private void activateJfrAndRunForever(Config config) {
...
+    OtlpLogsExporterBuilder builder =
+        OtlpLogsExporter.builder()
+            .setAdapter(adapter)
+            .addHeader("Extra-Content-Type", "otel-profiling-stacktraces");
+
+    String ingestUrl = config.getProperty(CONFIG_KEY_INGEST_URL);
+    if (ingestUrl != null) {
+      builder.setEndpoint(ingestUrl);
+    }
+    return builder.build();
```
jkwatson, line 168: do we have any way to write a unit test for this, at least to make sure it doesn't blow up?

jkwatson (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
breedx-splk: I'm not really sure of a good approach. We could put a setter on the builder and ensure that in the non-null case it got set, but I'm not sure how much that helps us....do you?

jkwatson: maybe make this method test-accessible and put in null & not-null into the config and make sure you get an non-null activator out the other side?

jkwatson: or separate exporter building into its own testable class?

breedx-splk: ok, took a stab at it.  Lemme know what you think.

profiler/src/main/java/com/splunk/opentelemetry/profiler/LogsExporterBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+import com.splunk.opentelemetry.logs.LogEntryAdapter;
+import com.splunk.opentelemetry.logs.LogsExporter;
+import com.splunk.opentelemetry.logs.OtlpLogsExporter;
+import com.splunk.opentelemetry.logs.OtlpLogsExporterBuilder;
+import com.splunk.opentelemetry.logs.ResourceLogsAdapter;
+import io.opentelemetry.instrumentation.api.config.Config;
+import io.opentelemetry.sdk.autoconfigure.OpenTelemetrySdkAutoConfiguration;
+import io.opentelemetry.sdk.resources.Resource;
+
+public class LogsExporterBuilder {
```
jkwatson, line 31: needs to be public?

jkwatson: also, why put this in the profiler package, rather than the logs package? If you did that, you wouldn't need to expose the properties on the exporter, right?

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -152,33 +141,33 @@ private void activateJfrAndRunForever(Config config) {
...
+  //    Resource resource = OpenTelemetrySdkAutoConfiguration.getResource();
+  //    LogEntryAdapter logEntryAdapter = new LogEntryAdapter();
+  //    InstrumentationLibraryLogsAdapter instLibraryLogsAdapter =
+  //        InstrumentationLibraryLogsAdapter.builder()
+  //            .logEntryAdapter(logEntryAdapter)
+  //            .instrumentationName(OTEL_INSTRUMENTATION_NAME)
+  //            .instrumentationVersion(OTEL_INSTRUMENTATION_VERSION)
+  //            .build();
+  //    return new ResourceLogsAdapter(instLibraryLogsAdapter, resource);
+  //  }
```
jkwatson, line 170:
```suggestion
```

breedx-splk: whoops, thanks.

profiler/src/main/java/com/splunk/opentelemetry/profiler/LogsExporterBuilder.java
breedx-splk: nope, reduced it.

jkwatson: I'm confused by this comment. If you put the builder into the logs package (along side what it builds), then you could move the test for it into the logs package, and the exporter could have it's accessors non-public. this class would be public still, which I think would be preferable to exposing things on the exporter itself.

breedx-splk: Unfortunately, I don't think this builder really exists other than to serve the profiling specific configurations....as it has profiling specific stuff in it.  As such, I think it should stay in the profiling package. There is also the `OtlpLogsExporterBuilder`, that's in the logs package.

Sorry about my terse/confusing comment...What I meant to say was that I made the change to reduce the access from public to package. 🤘🏻

PR #316, opened 2021-06-24
Configurable JFR event periods

breedx-splk:
This allows the user to customize/configure the JFR period for events in a somewhat generic way. The defaults are typically in the `jfr.settings` file and don't need to be changed, but especially for development this is helpful.

### DIFF
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -12,4 +12,4 @@ It should be considered experimental and is completely unsupported.
 |`splunk.profiler.recording.duration` | 20                     | number of seconds per recording unit      |
 |`splunk.profiler.keep-files`         | false                  | leave JFR files on disk id `true`         |
 |`splunk.profiler.logs-endpoint`      | http://localhost:4317  | where to send OTLP logs                   |
-|`splunk.profiler.period.{eventName}` | n/a                    | customize period for a specific jfr event |
+|`splunk.profiler.period.{eventName}` | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (100ms): `-Dsplunk.profiler.period.threaddump=1000` |
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -5,10 +5,11 @@ It should be considered experimental and is completely unsupported.
 
 # configuration
 
-| name                                | default                | description                          |
-|-------------------------------------|------------------------|--------------------------------------|
-|`splunk.profiler.enabled`            | false                  | set to true to enable the profiler   |
-|`splunk.profiler.directory`          | "."                    | location of jfr files                |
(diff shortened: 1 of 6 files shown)


profiler/README.md
```diff
@@ -5,10 +5,11 @@ It should be considered experimental and is completely unsupported.
...
-|`splunk.profiler.keep-files`         | false                  | leave JFR files on disk id `true`    |
-|`splunk.profiler.logs-endpoint`      | http://localhost:4317  | where to send OTLP logs              |
+| name                                | default                | description                               |
+|-------------------------------------|------------------------|-------------------------------------------|
+|`splunk.profiler.enabled`            | false                  | set to true to enable the profiler        |
+|`splunk.profiler.directory`          | "."                    | location of jfr files                     |
+|`splunk.profiler.recording.duration` | 20                     | number of seconds per recording unit      |
+|`splunk.profiler.keep-files`         | false                  | leave JFR files on disk id `true`         |
+|`splunk.profiler.logs-endpoint`      | http://localhost:4317  | where to send OTLP logs                   |
+|`splunk.profiler.period.{eventName}` | n/a                    | customize period for a specific jfr event |
```
jkwatson, line 15: should we document the most useful event name here? I assume that's "threaddump"?

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -30,6 +30,7 @@
       "splunk.profiler.recording.duration";
   public static final String CONFIG_KEY_KEEP_FILES = "splunk.profiler.keep-files";
   public static final String CONFIG_KEY_INGEST_URL = "splunk.profiler.logs-endpoint";
+  public static final String CONFIG_KEY_PERIOD_PREFIX = "splunk.profiler.period";
```
jkwatson, line 33: why not put the `.` at the end of this, and not have to append it inline?

breedx-splk: It was intentional, and purely preference. 

I used directory/paths as a reference...and it's miserable to always guess whether or not the first part has a trailing slash or not. (My) preference is always given to omitting slash from the dir and always having the slash present when concatenating because it feels more explicit.

profiler/README.md
breedx-splk: That makes sense.  It should also mention that it's milliseconds with no suffix, so I'll change it.  👍🏻

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
jkwatson: cool. one other thing...do these values need to be public?

jkwatson (approved):

PR #319, opened 2021-06-25
fetch ingested logs and make a few assertions

breedx-splk:
This stands up a collector now along with the fake backend from upstream, just like the smoke tests. It waits for log data to be present in the fake backend and then fetches/deserializes the data in order to make some assertions about it.  For now, we just verify that the period was propagated and that the sourcetype is specified and that we captured stacks for the `main` thread and the JFR thread itself.


jkwatson (approved):
👍🏼

PR #322, opened 2021-06-29
Create issue if snapshot build fails

iNikem:


.github/workflows/snapshot.yaml
```diff
@@ -22,3 +22,24 @@ jobs:
 
       - name: Build and test
         run: ./gradlew -Psnapshot=yes build --scan --no-daemon
+
+  issue:
+    name: Open issue on failure
+    needs: [ build ]
+    runs-on: ubuntu-latest
```
mateuszrzeszutek, line 29:
```suggestion
    runs-on: ubuntu-20.04
```

.github/workflows/nightly-failed.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Nightly build #{{ env.GITHUB_RUN_NUMBER }} failed
+assignees: iNikem
```
mateuszrzeszutek, line 3: Should we put the maintainers group here instead?

.github/workflows/snapshot.yaml
```diff
@@ -22,3 +22,24 @@ jobs:
 
       - name: Build and test
         run: ./gradlew -Psnapshot=yes build --scan --no-daemon
```
mateuszrzeszutek, line 24: > `-Psnapshot=yes`

Should we change how that flag works? Perhaps we should default to `snapshot=yes` and only pass `snapshot=no` when releasing a new stable version?

iNikem: I am not sure... Let's leave that as a separate though

mateuszrzeszutek (approved):

PR #327, opened 2021-06-30
wrap the assignees in a string

breedx-splk:
I think the `@` breaks the yaml if not in a string?

image

### DIFF
--- a/.github/workflows/nightly-failed.template
+++ b/.github/workflows/nightly-failed.template
@@ -1,6 +1,6 @@
 ---
 title: Nightly build #{{ env.GITHUB_RUN_NUMBER }} failed
-assignees: @signalfx/gdi-java-maintainers
+assignees: '@signalfx/gdi-java-maintainers'
 labels: bug, area:build, priority:p1
 ---
 <a href="https://github.com/{{ env.GITHUB_REPOSITORY }}/actions/runs/{{ env.GITHUB_RUN_ID }}">


laurit (approved):

.github/workflows/nightly-failed.template
```diff
@@ -1,6 +1,6 @@
 ---
 title: Nightly build #{{ env.GITHUB_RUN_NUMBER }} failed
-assignees: @signalfx/gdi-java-maintainers
+assignees: '@signalfx/gdi-java-maintainers'
```
mateuszrzeszutek, line 3: Just thought of it: maybe you don't need that `@` here at all?

breedx-splk: No clue!  😁 Is there even a way to test this stuff?

mateuszrzeszutek: No idea - I just noticed that the same file in the instrumentation repo doesn't have `@`: https://raw.githubusercontent.com/open-telemetry/opentelemetry-java-instrumentation/main/.github/workflows/nightly-failed.md

Somehow the idea that `@` might be a part of yaml syntax didn't occur to me when I was reviewing this change for the first time 😄

breedx-splk: I assumed the `@` was needed because it's a group and not just a user. Should we try it with the `@` in the string first?

mateuszrzeszutek (approved):

.github/workflows/nightly-failed.template
mateuszrzeszutek: Sure 👍

PR #328, opened 2021-06-30
Introduce gradle pluginManagement

mateuszrzeszutek:
This PR moves all gradle plugin versions to settings.gradle file - which was also migrated to Kotlin, just to try it out


build.gradle
```diff
@@ -44,7 +44,7 @@ subprojects {
     }
   }
 
-  tasks.withType(JavaCompile) {
+  tasks.withType(JavaCompile).configureEach {
```
breedx-splk, line 47: interesting that this was working before...

breedx-splk (approved):
Looks like a good improvement to me! Thanks.

build.gradle
mateuszrzeszutek: 🤷‍♂️

jkwatson (approved):

PR #334, opened 2021-07-02
Introduce maven-like dependencyManagement

mateuszrzeszutek:
I found a cool [spring gradle plugin](https://docs.spring.io/dependency-management-plugin/docs/current/reference/html/) that allows me to use maven-like `dependencyManagement` section


build.gradle
```diff
@@ -28,27 +29,80 @@ nexusPublishing {
...
+  apply plugin: "io.spring.dependency-management"
   apply from: "$rootDir/gradle/spotless.gradle"
-  apply from: "$rootDir/gradle/dependencies.gradle"
 
-  repositories {
-    mavenLocal()
-    mavenCentral()
-    maven {
-      url = uri("https://oss.sonatype.org/content/repositories/snapshots")
+  dependencyManagement {
```
mateuszrzeszutek, line 50: In the next PR I'll move this to a convention plugin (similar to what we have in the upstream repos) and get rid of the rest of those `gradle/` scripts

jkwatson (approved):

build.gradle
```diff
@@ -28,27 +29,80 @@ nexusPublishing {
...
+        entry('opentelemetry-javaagent-servlet-2.2')
+        entry('opentelemetry-javaagent-servlet-3.0')
+        entry('opentelemetry-javaagent-servlet-common')
+      }
+    }
+
+    imports {
+      mavenBom("io.grpc:grpc-bom:1.38.0")
+      mavenBom("io.micrometer:micrometer-bom:1.7.1")
+      mavenBom("io.opentelemetry:opentelemetry-bom-alpha:${otelAlphaVersion}")
```
iNikem, line 93: Should we publish bom from otel-java-instr as well?

instrumentation/khttp/build.gradle
```diff
@@ -2,6 +2,11 @@ apply plugin: 'groovy'
 apply from: "$rootDir/gradle/instrumentation.gradle"
 
 repositories {
+  mavenLocal()
+  mavenCentral()
+  maven {
+    url = uri("https://oss.sonatype.org/content/repositories/snapshots")
+  }
```
iNikem, line 9: Why do we need these three repos here?

iNikem (approved):

build.gradle
mateuszrzeszutek: We already do - the problem is, it contains mostly library instrumentations and is pretty useless here.

instrumentation/khttp/build.gradle
mateuszrzeszutek: The repositories defined in `dependencyResolutionManagement` are lost otherwise. I think I'll revert to the previous version though, the next PR will make it a bit better once I move the repositories definition into a convention plugin.

PR #345, opened 2021-07-06
Automate creating GH releases from GitLab

mateuszrzeszutek:


scripts/release.sh
```diff
@@ -0,0 +1,76 @@
...
+    "https://uploads.github.com/repos/signalfx/splunk-otel-java/releases/${gh_release_id}/assets?name=$(basename $artifact)" > /dev/null
+}
+
+build_project
+gh_release_id=$(create_release)
+
+for f in dist/*
+do
+  upload_artifact "$gh_release_id" "$f"
+done
```
iNikem, line 76: are you sure you don't want to use existing `softprops/action-gh-release` action for that?

mateuszrzeszutek: As it is now we can't do signing in GH, so I can't...

iNikem: I don't see any mentioning of signing in that action description 🤔

scripts/release.sh
```diff
@@ -0,0 +1,74 @@
...
+  mkdir -p ~/.gnupg
+  chmod 700 ~/.gnupg
+  cat > ~/.gnupg/gpg.conf <<EOF
+no-tty
+pinentry-mode loopback
+EOF
+
+  echo ">>> Importing secret key ..."
+  echo "$GPG_SECRET_KEY" > seckey.gpg
+  trap "rm seckey.gpg" EXIT
```
iNikem, line 36: This will not remove file if script gets killed. Use more signals, at least `KILL`

scripts/release.sh
```diff
@@ -0,0 +1,74 @@
...
+  gpg --batch --allow-secret-key-import --import seckey.gpg
+}
+
+build_project() {
+  mkdir -p dist
+
+  echo ">>> Building the javaagent ..."
+  ./gradlew -Prelease.useLastTag=true build final closeAndReleaseSonatypeStagingRepository -x test --no-daemon
```
iNikem, line 44: Why no tests? How can we be sure that this given commit is not broken?

scripts/release.sh
mateuszrzeszutek: Good point, will add 👍

scripts/release.sh
mateuszrzeszutek: Test are run in the `build` stage/job - see `.gitlab-ci.yml`. This script is only run once the build passes.

mateuszrzeszutek: Well, except smoke tests - gitlab runners can't into docker containers, so we only run those on GitHub.

scripts/release.sh
```diff
@@ -33,7 +33,7 @@ EOF
 
   echo ">>> Importing secret key ..."
   echo "$GPG_SECRET_KEY" > seckey.gpg
-  trap "rm seckey.gpg" EXIT
+  trap "rm seckey.gpg" EXIT KILL
```
iNikem, line 36: INT, TERM, STOP?

mateuszrzeszutek: Done

agent/build.gradle.kts
```diff
@@ -138,7 +138,7 @@ tasks {
             developer {
               id.set("splunk")
               name.set("Splunk Instrumentation Authors")
-              email.set("support+sonatype@signalfx.com")
+              email.set("support+java@signalfx.com")
```
jkwatson, line 141: is this a thing? Who get these emails, anyway?

mateuszrzeszutek: No idea 🤷‍♂️ 
`support+java` was what was in the signalfx-java repo, so I wanted to use the same value. And I used it as the GPG key name too.

iNikem (approved):

agent/build.gradle.kts
jkwatson: I'll put in a quick PR to change Android to use this as well

mateuszrzeszutek: 👍 

I'm doing the same thing for other repos too

PR #351, opened 2021-07-09
Implement metrics for Hikari connection pool

mateuszrzeszutek:
Plus some refactorings to share common names/conventions between connection pool instrumentations

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -76,20 +76,32 @@ All memory pool metrics have the following tags:
 | `runtime.jvm.threads.peak`   | The peak live thread count since the Java virtual machine started or peak was reset.
 | `runtime.jvm.threads.states` | The current number of threads per `state` (metric tag).
 
-### Apache DBCP2 connection pool
+### Connection pool metrics
 
-| Metric name                    | Description |
-| ------------------------------ | ----------- |
-| `db.pool.connections`          | The number of open connections.
-| `db.pool.connections.active`   | The number of open connections that are currently in use.
-| `db.pool.connections.idle`     | The number of open connections that are currently idle.
-| `db.pool.connections.idle.max` | The maximum number of idle open connections allowed.
-| `db.pool.connections.idle.min` | The minimum number of idle open connections allowed.
-| `db.pool.connections.max`      | The maximum number of open connections allowed.
+Splunk Distribution of OpenTelemetry Java instruments several JDBC connection pool implementations:
 
-All Apache DBCP2 metrics have the following tags:
+* [Apache DBCP2](https://commons.apache.org/proper/commons-dbcp/)
+* [HikariCP](https://github.com/brettwooldridge/HikariCP)
+
+Each of the supported connection pools reports a subset of the following metrics:
+
+| Metric name                           | Description |
+| ------------------------------------- | ----------- |
+| `db.pool.connections`                 | The number of open connections.
+| `db.pool.connections.active`          | The number of open connections that are currently in use.
+| `db.pool.connections.idle`            | The number of open connections that are currently idle.
+| `db.pool.connections.idle.max`        | The maximum number of idle open connections allowed.
+| `db.pool.connections.idle.min`        | The minimum number of idle open connections allowed.
+| `db.pool.connections.max`             | The maximum number of open connections allowed.
+| `db.pool.connections.pending_threads` | The number of threads that are currently waiting for an open connection.
+| `db.pool.connections.timeouts`        | The number of threads that are currently waiting for an open connection.
+| `db.pool.connections.create_time`     | The time it took to create a new connection.
+| `db.pool.connections.wait_time`       | The time it took to get an open connection from the pool.
+| `db.pool.connections.use_time`        | The time between borrowing a connection and returning it to the pool.
+
+All connection pool metrics have the following tags:
 
 | Tag name    | Tag value |
 | ----------- | --------- |
 | `pool.name` | The name of the connection pool: Spring bean name if Spring is used, the JMX object name otherwise.
-| `pool.type` | `dbcp2`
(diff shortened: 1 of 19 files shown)


instrumentation/hikaricp/src/test/java/com/splunk/opentelemetry/hikaricp/HikariInstrumentationTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private static void assertConnectionPoolMetrics() {
+    var tags = Map.of("pool.name", "testPool", "pool.type", "hikari");
+
+    assertThat(TestMetricsAccess.getMeters())
+        .containsExactlyInAnyOrder(
+            new MeterData("db.pool.connections", "gauge", "connections", tags),
+            new MeterData("db.pool.connections.active", "gauge", "connections", tags),
+            new MeterData("db.pool.connections.idle", "gauge", "connections", tags),
+            new MeterData("db.pool.connections.idle.min", "gauge", "connections", tags),
```
jkwatson, line 79: missing `idle.max`?

instrumentation/hikaricp/src/main/java/com/splunk/opentelemetry/hikaricp/HikariDataSourceInstrumentation.java
```diff
@@ -0,0 +1,53 @@
...
+  public void transform(TypeTransformer transformer) {
+    transformer.applyAdviceToMethod(
+        isConstructor(), this.getClass().getName() + "$ConstructorAdvice");
+  }
+
+  public static class ConstructorAdvice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void onExit(@Advice.This HikariDataSource dataSource) {
+      try {
+        dataSource.setMetricsTrackerFactory(new MicrometerMetricsTrackerFactory());
```
iNikem, line 46: Will that override/interfere somehow with user-supplied factory?

instrumentation/hikaricp/src/test/java/com/splunk/opentelemetry/hikaricp/HikariInstrumentationTest.java
mateuszrzeszutek: Hikari does not expose anything that could be used as `idle.max`.

instrumentation/hikaricp/src/main/java/com/splunk/opentelemetry/hikaricp/HikariDataSourceInstrumentation.java
mateuszrzeszutek: Yes, it will prevent other factories from being set. Do you think it could be a problem?

iNikem: That may be unexpected and undesirable for the end-user, if we suddenly break their own metric reporting pipelines.

mateuszrzeszutek: Hmm, good point. I'll rework that.

instrumentation/hikaricp/src/main/java/com/splunk/opentelemetry/hikaricp/HikariPoolInstrumentation.java
```diff
@@ -16,37 +16,39 @@
...
   public ElementMatcher<TypeDescription> typeMatcher() {
-    return named("com.zaxxer.hikari.HikariDataSource");
+    return named("com.zaxxer.hikari.pool.HikariPool");
   }
 
   @Override
   public void transform(TypeTransformer transformer) {
     transformer.applyAdviceToMethod(
-        isConstructor(), this.getClass().getName() + "$ConstructorAdvice");
+        named("setMetricsTrackerFactory")
```
iNikem, line 41: What if this method is never called? :)

mateuszrzeszutek: It's always called by in the (only) constructor of the class, I've added a comment that explains this choice.

docs/metrics.md
```diff
@@ -76,20 +76,32 @@ All memory pool metrics have the following tags:
...
+* [HikariCP](https://github.com/brettwooldridge/HikariCP)
+
+Each of the supported connection pools reports a subset of the following metrics:
+
+| Metric name                           | Description |
+| ------------------------------------- | ----------- |
+| `db.pool.connections`                 | The number of open connections.
+| `db.pool.connections.active`          | The number of open connections that are currently in use.
+| `db.pool.connections.idle`            | The number of open connections that are currently idle.
+| `db.pool.connections.idle.max`        | The maximum number of idle open connections allowed.
```
jkwatson, line 93: since hikari doesn't support this, should we add a comment about that here?

jkwatson (approved):
Looks good! Do you have a demo ready to go to show it off?

mateuszrzeszutek:
> Do you have a demo ready to go to show it off?

Not yet - I'm planning to spend some time on an example app tomorrow.

docs/metrics.md
mateuszrzeszutek: To be honest I'm still wondering how to handle docs for those connection pools - I've started looking into another one (`tomcat-jdbc`) and it seems that it supports yet another subset of these metrics (and maybe a few new ones, who knows). I think there are 2 or 3 more connection pools to cover - after I'm done with instrumenting them I'll probably reserve some time for making them a bit more uniform, and for fixing those docs too.
Anyway, I'm thinking of either adding a column with a list of connection pools that support these metrics or having a separate table for each pool implementation.

iNikem (approved):

PR #362, opened 2021-07-21
Update upstream 1.4.0

breedx-splk:


buildSrc/src/main/kotlin/splunk.java-conventions.gradle.kts
```diff
@@ -13,9 +13,9 @@ repositories {
   }
 }
 
-val otelVersion = "1.3.0"
-val otelAlphaVersion = "1.3.0-alpha"
-val otelInstrumentationVersion = "1.4.0-SNAPSHOT"
+val otelVersion = "1.4.0"
+val otelAlphaVersion = "1.4.0-alpha"
+val otelInstrumentationVersion = "1.5.0-SNAPSHOT"
 val otelInstrumentationAlphaVersion = "1.4.0-alpha-SNAPSHOT"
```
iNikem, line 19:
```suggestion
val otelInstrumentationAlphaVersion = "1.5.0-alpha-SNAPSHOT"
```

iNikem (approved):

buildSrc/src/main/kotlin/splunk.java-conventions.gradle.kts
breedx-splk: You'd think so, wouldn't you.

It doesn't work.  There is no `opentelemetry-javaagent-api` published for that version, which breaks our build.
https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/javaagent/opentelemetry-javaagent-api/

Maybe we need to pull `opentelemetry-javaagent-api` out of that dependency set and have it use its own thing?

breedx-splk: I hadn't realized that the api was also renamed. I made this change and think it's fixed up now.

iNikem (approved):

PR #368, opened 2021-07-26
Remove all locally changed files before running ./gradlew final

mateuszrzeszutek:
This should fix the following problem:

```
> Failed to apply plugin 'nebula.release'.
   > Final and candidate builds require all changes to be committed into Git.
     Found unstaged changes:
       [+] seckey.gpg
```


iNikem:
Wait, why `trap` does not work?

mateuszrzeszutek:
It does, but it only executes on process exit 🤷‍♂️

iNikem:
A, got it.

iNikem (approved):

PR #369, opened 2021-07-27
Publish a no-classifier jar too

mateuszrzeszutek:


agent/build.gradle.kts
```diff
@@ -105,8 +105,19 @@ tasks {
     }
   }
 
+  val mainShadowJar by registering(Jar::class) {
```
mateuszrzeszutek, line 108: I wanted to just use `Copy` here but it was so painful and problematic (with the publishing stuff) that I ended up just using `Jar` - which extracts and creates a new archive.

iNikem: Why exactly is `Copy` painful?

mateuszrzeszutek: For `Copy` the output is the directory, not the copied jar file - so I have to add the copied jar file manually to the publishing config by passing the path to the file. But then, for some reason, publishing picks up the jar that was built by the `jar` task (because it has the same path I guess?), which is disabled by the way (🤯 ), and that jar file does not contain Splunk classes.

iNikem (approved):

PR #374, opened 2021-07-27
Refactor javaagent buildscripts

mateuszrzeszutek:
... so that they're more similar to what we have in the upstream repo (and easier to maintain too)

* Removed shadowing from `custom`, `instrumentation` and `profiler`
* Moved `SplunkAgent` to `bootstrap` to avoid depending on `agent:jar`, which is disabled and I have no idea why this has worked before

### DIFF
--- a/agent/build.gradle.kts
+++ b/agent/build.gradle.kts
@@ -14,17 +14,17 @@ java {
 }
 
 // this configuration collects libs that will be placed in the bootstrap classloader
-val bootstrapLibs by configurations.creating {
+val bootstrapLibs: Configuration by configurations.creating {
   isCanBeResolved = true
   isCanBeConsumed = false
 }
 // this configuration collects libs that will be placed in the agent classloader, isolated from the instrumented application code
-val javaagentLibs by configurations.creating {
+val javaagentLibs: Configuration by configurations.creating {
   isCanBeResolved = true
   isCanBeConsumed = false
 }
 // this configuration stores the upstream agent dep that's extended by this project
-val upstreamAgent by configurations.creating {
+val upstreamAgent: Configuration by configurations.creating {
   isCanBeResolved = true
   isCanBeConsumed = false
 }
@@ -46,6 +46,9 @@ tasks {
     enabled = false
   }
 
+  // building the final javaagent jar is done in 3 steps:
+
+  // 1. all Splunk-specific javaagent libs are relocated (by the splunk.shadow-conventions plugin)
   val relocateJavaagentLibs by registering(ShadowJar::class) {
     configurations = listOf(javaagentLibs)
 
@@ -54,8 +57,10 @@ tasks {
     archiveFileName.set("javaagentLibs-relocated.jar")
   }
 
-  // having a separate task for isolating javaagent libs is required to avoid duplicates
-  // duplicatesStrategy in shadowJar won't be applied when adding files with with(CopySpec)
+  // 2. the Splunk javaagent libs are then isolated - moved to the inst/ directory
+  // having a separate task for isolating javaagent libs is required to avoid duplicates with the upstream javaagent
+  // duplicatesStrategy in shadowJar won't be applied when adding files with with(CopySpec) because each CopySpec has
+  // its own duplicatesStrategy
   val isolateJavaagentLibs by registering(Copy::class) {
     dependsOn(relocateJavaagentLibs)
     isolateClasses(relocateJavaagentLibs.get().outputs.files)
@@ -63,6 +68,8 @@ tasks {
     into("$buildDir/isolated/javaagentLibs")
   }
 
+  // 3. the relocated and isolated javaagent libs are merged together with the bootstrap libs (which undergo relocation
+  // in this task) and the upstream javaagent jar; duplicates are removed
   shadowJar {
     configurations = listOf(bootstrapLibs, upstreamAgent)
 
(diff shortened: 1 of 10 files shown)


agent/build.gradle.kts
```diff
@@ -13,98 +13,81 @@ java {
...
+val bootstrapLibs by configurations.creating {
+  isCanBeResolved = true
+  isCanBeConsumed = false
+}
+// this configuration collects libs that will be placed in the agent classloader, isolated from the instrumented application code
+val javaagentLibs by configurations.creating {
+  isCanBeResolved = true
+  isCanBeConsumed = false
+}
+// this configuration stores the upstream agent dep that's extended by this project
```
breedx-splk, line 26: thanks for these helpful comments.

agent/build.gradle.kts
```diff
@@ -13,98 +13,81 @@ java {
...
-        rename("(^.*)\\.class\$", "\$1.classdata")
-      }
-    }
-  }
+  upstreamAgent("io.opentelemetry.javaagent:opentelemetry-javaagent:${otelInstrumentationVersion}:all")
 }
 
 tasks {
-  compileJava {
-    options.release.set(8)
```
breedx-splk, line 62: I think we still need to specify version 8 somewhere for the agent, don't we?  I see it being set in instrumentation and profiler, but not the agent. Am I just not seeing it here?

agent/build.gradle.kts
```diff
@@ -179,3 +161,12 @@ tasks {
 rootProject.tasks.named("release") {
   finalizedBy(tasks["publishToSonatype"])
 }
+
+fun CopySpec.isolateClasses(jars: Iterable<File>) {
+  jars.forEach {
+    from(zipTree(it)) {
+      into("inst")
+      rename("(^.*)\\.class\$", "\$1.classdata")
```
breedx-splk, line 169: In my brain I wouldn't consider the starting ^ part of the match group, but if this works it's fine with me. Not a big deal.

breedx-splk (commented):
Looks ok to me, I had a couple small comments/questions and would definitely want a 2nd opinion.  It seems like `splunk.shadow-conventions` is removed from most places, but why not all? 

I built the jar on main and compared the contents to the agent built on this branch.  Here are the differences:
https://gist.github.com/breedx-splk/6c962a28815f71f158031a7e7c349427

It looks like we were shading in the metrics api into a second package before. We previously had:
* `inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/metrics`
* in addition to `io/opentelemetry/javaagent/shaded/io/opentelemetry/api/metrics`

I'm guessing that wasn't desirable?

agent/build.gradle.kts
```diff
@@ -13,98 +13,81 @@ java {
...
   }
 
-  // includes everything except otel agent
-  val relocateAndIsolate by registering(ShadowJar::class) {
-    dependsOn(":custom:shadowJar")
-    dependsOn(":instrumentation:shadowJar")
-    dependsOn(":profiler:shadowJar")
-    dependsOn(":bootstrap:jar")
-    dependsOn(":agent:jar")
+  val relocateJavaagentLibs by registering(ShadowJar::class) {
```
iNikem, line 49: I think we need some explanatory documentation for all this. I would mention exact steps that are happening: first, this task shades, then next Copy task isolates, then last shadow task adds upstream agent and bootstrap. And also that actual shading rules come from the convention. This last piece got me confused for a couple of minutes...

buildSrc/src/main/kotlin/splunk.shadow-conventions.gradle.kts
```diff
@@ -6,6 +6,9 @@ plugins {
 
 tasks.withType<ShadowJar>().configureEach {
   mergeServiceFiles()
+  mergeServiceFiles {
```
iNikem, line 9: Why do we need 2 merge calls?

agent/build.gradle.kts
iNikem: There is no code anymore in agent module, just packaging infra. So nothing to compile.

iNikem (approved):

agent/build.gradle.kts
mateuszrzeszutek: Exactly that ^^

Although I guess we could set release=8 in the java convention script for all projects...

buildSrc/src/main/kotlin/splunk.shadow-conventions.gradle.kts
mateuszrzeszutek: We actually don't - removed. Thanks!

agent/build.gradle.kts
mateuszrzeszutek: Good point, done.

mateuszrzeszutek:
> It seems like `splunk.shadow-conventions` is removed from most places, but why not all?

Because we still need it when building javaagents: in the `agent` and `testing:agent-for-testing` modules.

> I built the jar on main and compared the contents to the agent built on this branch. Here are the differences:
> https://gist.github.com/breedx-splk/6c962a28815f71f158031a7e7c349427
> 
> It looks like we were shading in the metrics api into a second package before. We previously had:
> 
> * `inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/metrics`
> * in addition to `io/opentelemetry/javaagent/shaded/io/opentelemetry/api/metrics`
> 
> I'm guessing that wasn't desirable?

It wasn't, it was a bug actually - OTel API should only be available in the bootstrap classpath, not the agent classpath.
I've noticed that we still have some parts of OTel API in the inst/ directory, I'll remove them too. They should definitely not be there.

agent/build.gradle.kts
mateuszrzeszutek: Wow, I copy-pasted this fragment of code tens of times and didn't even notice that 😂

PR #383, opened 2021-08-04
Implement metrics for Tomcat thread pools

mateuszrzeszutek:


instrumentation/tomcat/src/main/java/com/splunk/opentelemetry/tomcat/metrics/AbstractEndpointInstrumentation.java
```diff
@@ -0,0 +1,61 @@
...
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.apache.tomcat.util.net.AbstractEndpoint;
+
+final class AbstractEndpointInstrumentation implements TypeInstrumentation {
+
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return named("org.apache.tomcat.util.net.AbstractEndpoint");
```
jkwatson, line 33: What is an `AbstractEndpoint` and why is it the thing to instrument for a thread pool? I think it would be useful for future us to have a few comments about what this tomcat class is about and why it's the right thing to hook into for thread pool metrics.

mateuszrzeszutek: Good point, added.

docs/metrics.md
```diff
@@ -114,3 +115,28 @@ All connection pool metrics have the following tags:
...
+
+Each of the supported connection pools reports a subset of the following metrics:
+
+| Metric name                | Description |
+| -------------------------- | ----------- |
+| `executor.threads`         | The current number of threads in the pool.
+| `executor.threads.active`  | The number of threads that are currently busy.
+| `executor.threads.idle`    | The number of threads that are currently idle.
+| `executor.threads.core`    | Core thread pool size - the number of threads that are always kept in the pool.
+| `executor.threads.max`     | The maximum number of threads in the pool.
```
breedx-splk, line 133: I definitely understand the need to have `core` and `max` accounted for, but they really don't change during the lifespan of a pool, right? It's an interesting idea to report the same metric repeatedly.....hmmm....not sure what to do about it, just observing.

instrumentation/tomcat/src/main/java/com/splunk/opentelemetry/tomcat/metrics/ThreadPoolMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+      return getMetric(ThreadPoolExecutor::getCompletedTaskCount, e -> Double.NaN);
+    }
+
+    private Number getMetric(
+        Function<ThreadPoolExecutor, Number> threadPoolExecutorFunc,
+        Function<ResizableExecutor, Number> resizableExecutorFunc) {
+      Executor executor = endpoint.getExecutor();
+      if (executor instanceof ThreadPoolExecutor) {
+        return threadPoolExecutorFunc.apply((ThreadPoolExecutor) executor);
+      } else if (executor instanceof ResizableExecutor) {
```
breedx-splk, line 123: I like removing the redundant `else` here (and below).

breedx-splk (changes requested):
Looks good, but I wonder about a couple of those gauges that could/should be counters?

breedx-splk (approved):

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/ThreadPoolSemanticConventions.java
```diff
@@ -0,0 +1,56 @@
...
+  public static final GaugeSemanticConvention THREADS_IDLE =
+      gauge("executor.threads.idle", THREADS);
+  /** Core thread pool size - the number of threads that are always kept in the pool. */
+  public static final GaugeSemanticConvention THREADS_CORE =
+      gauge("executor.threads.core", THREADS);
+  /** The maximum number of threads in the pool. */
+  public static final GaugeSemanticConvention THREADS_MAX = gauge("executor.threads.max", THREADS);
+  /** The total number of tasks that were submitted to this executor. */
+  public static final GaugeSemanticConvention TASKS_SUBMITTED =
+      gauge("executor.tasks.submitted", TASKS);
```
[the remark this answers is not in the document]
breedx-splk: Ah ok, Watson and I chatted about this and it looks like the pool just tracks a singular number that happens to be monotonically increasing but does not report deltas...so we'd have to keep state and do that math ourselves each time, which isn't great. I'm sure this is fine as a gauge.  Very minor (microscopic) bummer.

docs/metrics.md
mateuszrzeszutek: You actually can change those by calling `setCorePoolSize()`/`setMaximumPoolSize()` - I don't think that anybody does that though, they're effectively final.

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/ThreadPoolSemanticConventions.java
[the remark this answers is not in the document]
mateuszrzeszutek: I thought that too, but I checked the micrometer API once again and it turns out there's a `FunctionCounter` instrument there that can keep track of a monotonically increasing function - which fits the purpose of those `TASKS_*` meters. So I ended up using a `FunctionCounter` instead of a gauge after all.
(thanks for the comment by the way - I probably wouldn't have ended up scanning through the micrometer API if it weren't for that)

PR #388, opened 2021-08-06
Implement application<->agent Micrometer bridge

mateuszrzeszutek:
Part 1 - I left a couple of TODOs that'll be resolved in next PR(s)

### DIFF
--- a/agent/build.gradle.kts
+++ b/agent/build.gradle.kts
@@ -35,10 +35,21 @@ dependencies {
   bootstrapLibs(project(":bootstrap"))
 
   javaagentLibs(project(":custom"))
-  javaagentLibs(project(":instrumentation"))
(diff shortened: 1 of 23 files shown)


iNikem:
So if application uses Micrometer  and records some metrics "manually" and scrapes them into Prometheus and then user adds our agent, will they still get their metrics as before, "the old way"? In addition to them being reported by to our APM?

mateuszrzeszutek:
> So if application uses Micrometer and records some metrics "manually" and scrapes them into Prometheus and then user adds our agent, will they still get their metrics as before, "the old way"? In addition to them being reported by to our APM?

Yes, that is correct. We will not overwrite anything the user already has set up.

iNikem (approved):

PR #399, opened 2021-08-12
Add muzzle to GHA jobs

mateuszrzeszutek:


.github/workflows/nightly.yaml
```diff
@@ -1,8 +1,7 @@
-name: Build with upstream snapshot
+name: Nightly build
```
mateuszrzeszutek, line 1: Every build uses snapshots now, so I guess it' better to just call it "nightly"

iNikem (approved):

PR #402, opened 2021-08-16
Update correlating-traces-with-logs.md

somensari-splunk:
The property tag (otel.resource.enviroment) does not work. The proper tag is otel.resource.deployment.environment, I just tested on my local environment.


breedx-splk (approved):
Indeed, this relates to #139. Thanks for catching this.

PR #403, opened 2021-08-17
Implement metrics for IBM WebSphere thread pools

mateuszrzeszutek:
I couldn't really test those metrics in this PR, because the OTel Liberty image does not have the `monitor-1.0` feature enabled and it doesn't register the thread pool mbean.

### DIFF
--- a/bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxMetricsWatcher.java
+++ b/bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxMetricsWatcher.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.javaagent.bootstrap.metrics.jmx;
+
+import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toSet;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
+import java.lang.management.ManagementFactory;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+import javax.management.InstanceNotFoundException;
+import javax.management.JMException;
+import javax.management.ListenerNotFoundException;
+import javax.management.MBeanServer;
+import javax.management.MBeanServerDelegate;
+import javax.management.MBeanServerFactory;
+import javax.management.MBeanServerNotification;
+import javax.management.Notification;
+import javax.management.NotificationFilter;
+import javax.management.NotificationListener;
+import javax.management.ObjectName;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class JmxMetricsWatcher {
+
+  private static final Logger log = LoggerFactory.getLogger(JmxMetricsWatcher.class);
+
+  private static final AtomicInteger threadIds = new AtomicInteger();
+  private static final ThreadFactory daemonThreadFactory =
+      r -> {
+        Thread t = new Thread(r, "JmxMetricsWatcher-" + threadIds.incrementAndGet());
+        t.setDaemon(true);
+        t.setContextClassLoader(null);
+        return t;
+      };
+
+  private static final Set<String> handledNotificationTypes =
+      new HashSet<>(
+          asList(
+              MBeanServerNotification.REGISTRATION_NOTIFICATION,
+              MBeanServerNotification.UNREGISTRATION_NOTIFICATION));
+
+  private final MBeanServer mBeanServer;
+  private final ExecutorService executorService;
+  private final MeterRegistry meterRegistry;
+  private final JmxQuery query;
+  private final MetersFactory metersFactory;
+
+  private final Map<ObjectName, List<Meter>> meters = new ConcurrentHashMap<>();
+  private volatile NotificationListener notificationListener;
+
+  public JmxMetricsWatcher(JmxQuery query, MetersFactory metersFactory) {
+    this(
+        findMBeanServer(),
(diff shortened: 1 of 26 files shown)


bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxMetricsWatcher.java
```diff
@@ -0,0 +1,193 @@
...
+    this.executorService = executorService;
+    this.meterRegistry = meterRegistry;
+    this.query = query;
+    this.metersFactory = metersFactory;
+  }
+
+  // copied from micrometer CommonsObjectPool2Metrics class
+  private static MBeanServer findMBeanServer() {
+    List<MBeanServer> mBeanServers = MBeanServerFactory.findMBeanServer(null);
+    if (!mBeanServers.isEmpty()) {
```
breedx-splk, line 105: I'd uninvert the `if` logic here.

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxMetricsWatcher.java
```diff
@@ -0,0 +1,193 @@
...
+
+    executorService.shutdown();
+
+    meters.values().stream().flatMap(Collection::stream).forEach(meterRegistry::remove);
+    meters.clear();
+  }
+
+  // visible for tests
+  Set<Meter> getMeters() {
+    return meters.values().stream().flatMap(Collection::stream).collect(toSet());
```
breedx-splk, line 148: most of this is duplicated with the above. You could factor out `getAllMeters()` or something that just encapsulates `meters.values().stream().flatMap(Collection::stream)`

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxMetricsWatcher.java
```diff
@@ -0,0 +1,193 @@
...
+    return query.matches(objectName);
+  }
+
+  private void handleNotification(Notification notification, Object handback) {
+    // we can't get JMX object attributes in the NotificationListener, so we'll do that
+    // asynchronously on a separate thread
+    executorService.submit(() -> handleNotificationAsync(notification));
+  }
+
+  private void handleNotificationAsync(Notification notification) {
```
breedx-splk, line 178: I think the names of `handleNotification` and `handleNotificationAsync` are swapped. To me, the async method should be the non-blocking one and the other one should be the one that does the work (blocking).

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxQuery.java
```diff
@@ -0,0 +1,64 @@
...
+
+  ObjectName toObjectNameQuery() throws MalformedObjectNameException {
+    String propertiesStr =
+        properties().entrySet().stream()
+            .map(e -> e.getKey() + "=" + e.getValue())
+            .collect(Collectors.joining(","));
+    if (!propertiesStr.isEmpty()) {
+      propertiesStr += ",";
+    }
+    return new ObjectName(domain() + ":" + propertiesStr + "*");
```
breedx-splk, line 50: Since it looks like `propertiesStr` can be empty, is it safe to allow a domain wildcard query?

instrumentation/liberty/src/main/java/com/splunk/opentelemetry/instrumentation/liberty/LibertyMetricsInstaller.java
```diff
@@ -0,0 +1,72 @@
...
+public class LibertyMetricsInstaller implements AgentListener {
+
+  private final JmxMetricsWatcher jmxMetricsWatcher =
+      new JmxMetricsWatcher(
+          JmxQuery.create("WebSphere", "type", "ThreadPoolStats"), this::createMeters);
+
+  @Override
+  public void afterAgent(Config config) {
+    boolean metricsRegistryPresent = !Metrics.globalRegistry.getRegistries().isEmpty();
+    if (!config.isInstrumentationEnabled(singleton("liberty"), metricsRegistryPresent)) {
```
breedx-splk, line 50: uninvert the `if` please.

breedx-splk: Also, does this also make sure that we're running inside of Liberty? I think we would definitely want to prevent spinning up a jmx snarfing thread pool unless actually running inside of WebSphere/Liberty, right? Did I miss that?

breedx-splk (changes requested):
Good stuff. A couple questions and tweaks. 🚚 🕸️ 🌐

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxMetricsWatcher.java
mateuszrzeszutek: Good point, will do that. 👍

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxMetricsWatcher.java
mateuszrzeszutek: That's totally true 🙈  It seems I mixed them up and didn't notice - thanks!

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxQuery.java
mateuszrzeszutek: What do you mean by "allow a domain wildcard query"? Something like `JmxQuery.create("*", Map.of())`?

instrumentation/liberty/src/main/java/com/splunk/opentelemetry/instrumentation/liberty/LibertyMetricsInstaller.java
mateuszrzeszutek: Yeah it did always run without checking that... Fixed that now - I found a pretty good place to start the watcher.

laurit:
If it helps it is possible to overwrite configuration inside otel image to enable more features. For example see https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertyServletOnlySmokeTest.groovy

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxQuery.java
breedx-splk: Kinda. Like the query ends up being `<domain>:*`. Is there a risk of that matching too many objects and being heavyweight?

breedx-splk (approved):

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/metrics/jmx/JmxQuery.java
mateuszrzeszutek: There is - but I found it really useful during debugging, so I decided to leave it. There's a significant chance it'll be needed again when we pick up the next story involving JMX.

breedx-splk: 👍🏻

PR #406, opened 2021-08-17
Add profiler TLAB support.

breedx-splk:
Tracking allocations and the stacks for where they happen.

### DIFF
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -13,3 +13,4 @@ It should be considered experimental and is completely unsupported.
 |`splunk.profiler.keep-files`         | false                  | leave JFR files on disk id `true`         |
 |`splunk.profiler.logs-endpoint`      | `otel.exporter.otlp.endpoint` or http://localhost:4317  | where to send OTLP logs                   |
 |`splunk.profiler.period.{eventName}` | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (100ms): `-Dsplunk.profiler.period.threaddump=1000` |
+|`splunk.profiler.tlab.enabled`       | true                   | set to `false` to disable TLAB memory events |
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler.util;
+
+import java.util.List;
+import jdk.jfr.consumer.RecordedFrame;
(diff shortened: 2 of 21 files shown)


profiler/README.md
```diff
@@ -13,3 +13,4 @@ It should be considered experimental and is completely unsupported.
 |`splunk.profiler.keep-files`         | false                  | leave JFR files on disk id `true`         |
 |`splunk.profiler.logs-endpoint`      | http://localhost:4317  | where to send OTLP logs                   |
 |`splunk.profiler.period.{eventName}` | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (100ms): `-Dsplunk.profiler.period.threaddump=1000` |
+|`splunk.profiler.tlab.enabled`       | true                   | set to `false` to disable TLAB memory events |
```
jkwatson, line 16: don't we want this off by default, due to the cost?

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public String serialize(RecordedStackTrace stack) {
+    List<RecordedFrame> frames = stack.getFrames();
+    int skipNum = Math.max(0, frames.size() - maxDepth);
+    return frames.stream()
+        .skip(skipNum)
+        .limit(maxDepth)
+        .reduce(
+            "",
```
mateuszrzeszutek, line 44: Nit: would it be possible to use a StringBuilder here?

mateuszrzeszutek (approved):

profiler/README.md
breedx-splk: I think this is still an open question. We haven't yet really quantified the cost yet, so maybe to play it safe is a better idea. What do @iNikem and @mateuszrzeszutek and @laurit think?

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
breedx-splk: Good idea! Thanks.

profiler/README.md
jkwatson: My vote would be to leave it off by default until we've at least verified it's not going to destroy the average app's performance.

breedx-splk: It's still turned off unless `splunk.profiler.enabled=true`, which is not the default (yet?).

breedx-splk: Ok, I think you're right and that this is a better idea, so I've inverted the logic around this and defaulted tlab support to off.

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
```diff
@@ -0,0 +1,60 @@
...
+    List<RecordedFrame> frames = stack.getFrames();
+    int skipNum = Math.max(0, frames.size() - maxDepth);
+    return frames.stream()
+        .skip(skipNum)
+        .limit(maxDepth)
+        .reduce(
+            new StringBuilder(),
+            (sb, recordedFrame) -> {
+              String frameStr = serializeFrame(recordedFrame);
+              return sb.length() == 0 ? sb.append(frameStr) : sb.append("\n").append(frameStr);
```
mateuszrzeszutek, line 47: You can use the `StringBuilder` in `serializeFrame` too 😄 

```java
if (sb.length() > 0) {
  sb.append("\n");
}
serializeFrame(recordedFrame, sb);
return sb;
```

And avoid one string concatenation per each frame.

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
breedx-splk: Sounds good.

mateuszrzeszutek (approved):

PR #415, opened 2021-08-20
Disable lychee checks

mateuszrzeszutek:
A new `lycheeverse/lychee:latest` tag was pushed ~9 hours ago, and it's completely broken - it fails with the following error

```
standard_init_linux.go:228: exec user process caused: no such file or directory
```

Anyway, I find it REALLY BAD that a concrete `lycheeverse/lychee-action@v1.0.8` version depends on a "latest" tag of some other image.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,26 +11,27 @@ on:
       - main
 
 jobs:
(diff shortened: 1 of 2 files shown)


kubawach (approved):
Oooof!

PR #421, opened 2021-08-23
Changelog, metrics manual instrumentation docs, and 1.3.0 pre-release changes

mateuszrzeszutek:


docs/metrics.md
```diff
@@ -22,13 +22,63 @@ The following dimensions are automatically added to all metrics exported by the
 | `process.pid`            | The Java process identifier (PID).
 | `service`                | The value of the `service.name` resource attribute.
 
+## Manual instrumentation
```
mateuszrzeszutek, line 25: I've noticed that I forgot to document how to manually add metrics, so I added it right now - that's a feature that needs to be described in this release's docs.

CHANGELOG.md
```diff
@@ -21,9 +47,10 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 ### Enhancements
 
 - We have added instrumentation for [HikariCP](https://github.com/brettwooldridge/HikariCP)
-  and [Tomcat JDBC](https://tomcat.apache.org/tomcat-8.5-doc/jdbc-pool.html) connection pools. The agent now
-  collects and exports metrics for both JDBC connection pools.
-- You can now set the service name using the `OTEL_SERVICE_NAME` environment variable and the `otel.service.name` system property. This removes the need of using `OTEL_RESOURCE_ATTRIBUTES` to set the service name.
+  and [Tomcat JDBC](https://tomcat.apache.org/tomcat-8.5-doc/jdbc-pool.html) connection pools. The agent now collects
+  and exports metrics for both JDBC connection pools.
+- You can now set the service name using the `OTEL_SERVICE_NAME` environment variable and the `otel.service.name` system
```
breedx-splk, line 52: Would be nice to link to the spec here (https://github.com/open-telemetry/opentelemetry-specification/pull/1677 or https://github.com/open-telemetry/opentelemetry-specification/blob/dfd4039a4fc178da89dc2e9eaefd79f651604fa3/CHANGELOG.md#sdk-configuration-3 or https://github.com/open-telemetry/opentelemetry-specification/blob/c5f1a70340059953244fcc9aabb371347e0989fc/specification/sdk-environment-variables.md#general-sdk-configuration not sure which is best)

docs/faq.md
```diff
@@ -9,11 +9,6 @@
   for Splunk products, access to other open-source projects including
   Micrometer. Note, we take an upstream-first approach, Splunk Distribution of
   OpenTelemetry Java allow us to move fast.
-- **Why don't you publish the javaagent jar to a Maven repository?** It would
-  make it very easy to accidentally put the agent on the application runtime
-  classpath, which may cause all sorts of problems and confusion - and the
-  agent won't work anyway, because it has to be passed in the `-javaagent` JVM
-  parameter.
```
breedx-splk, line 16: I guess we removed this because we _do_ publish to maven now. Might be nice to swizzle this around to say something like `Q: Should I put the agent on my classpath? A: No, the agent should never be on the runtime classpath`. Down the road, we could provide some gradle/mvn snippets that show how to source it without including it in the classpath....I dunno.

breedx-splk (approved):
I had a couple small comments, but this looks good. Thanks for taking this on!

docs/faq.md
mateuszrzeszutek: Good point, will add.

PR #433, opened 2021-09-02
log the profiler config at agent start.

breedx-splk:
This helps us to do troubleshooting and provide additional support.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/ConfigurationLogger.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/ConfigurationLogger.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler;
+
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_ENABLE_PROFILER;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_INGEST_URL;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_KEEP_FILES;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_OTEL_OTLP_URL;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_PERIOD_PREFIX;
+import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_PROFILER_DIRECTORY;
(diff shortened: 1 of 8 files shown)


profiler/src/test/java/com/splunk/opentelemetry/profiler/ConfigurationLoggerTest.java
```diff
@@ -0,0 +1,61 @@
...
+        when(config.getBoolean(CONFIG_KEY_TLAB_ENABLED)).thenReturn(false);
+        when(config.getString(CONFIG_KEY_PERIOD_PREFIX + ".jdk.threaddump")).thenReturn("500ms");
+
+        ConfigurationLogger configurationLogger = new ConfigurationLogger(log);
+
+        configurationLogger.log(config);
+
+        InOrder inOrder = inOrder(log);
+        inOrder.verify(log).info("-----------------------");
+        inOrder.verify(log).info("Profiler configuration:");
```
jkwatson, line 48: any thoughts to using logunit here, rather than mocking?

profiler/src/test/java/com/splunk/opentelemetry/profiler/ConfigurationTest.java
```diff
@@ -0,0 +1,41 @@
+package com.splunk.opentelemetry.profiler;
+
+import io.opentelemetry.instrumentation.api.config.Config;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class ConfigurationTest {
```
jkwatson, line 10: this change seems unrelated to the logging of the config.

profiler/src/test/java/com/splunk/opentelemetry/profiler/ConfigurationLoggerTest.java
breedx-splk: yeah that worked out, thanks for the advice.

profiler/src/test/java/com/splunk/opentelemetry/profiler/ConfigurationTest.java
breedx-splk: It is. I had factored it out thinking that I would use it in the logging, but it was misleading in the logs, so I didn't end up using it there...but I felt like it was still a good change so I left it. Do you think it should be removed?

jkwatson: doesn't matter to me too much. I was just confused to see this here when it didn't appear to have anything to do with the rest of the PR. :)

jkwatson (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/ConfigurationLogger.java
```diff
@@ -0,0 +1,59 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** This class logs the active profiler configuration for debug/troubleshooting purposes. */
+public class ConfigurationLogger {
+
+  private static final Logger logger = LoggerFactory.getLogger(ConfigurationLogger.class);
+
+  public void log(Config config) {
+    logger.info("-----------------------");
```
mateuszrzeszutek, line 39: Just one thought: do we want to log everything to `INFO` if it's for debug/troubleshooting purposes?

mateuszrzeszutek: Or maybe we could guar this with `config.isAgentDebugEnabled()`

breedx-splk: I did intend for it to be at INFO -- because it's only at app startup and it's pretty important when troubleshooting.

PR #434, opened 2021-09-03
Safer JFR availability checks

breedx-splk:
Tested on 8u212 and sure enough, there were some exceptions in the logs about unknown JFR classes. I added some additional preventions here and confirmed that the exceptions no longer happen. Also confirmed that JFR still works on newer JVMs.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JFR.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JFR.java
@@ -29,7 +29,12 @@ class JFR {
   public static final JFR instance = new JFR();
 
   public boolean isAvailable() {
-    return FlightRecorder.isAvailable();
+    try {
+      JFR.class.getClassLoader().loadClass("jdk.jfr.FlightRecorder");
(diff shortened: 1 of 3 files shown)


jkwatson (approved):

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/JFR.java
```diff
@@ -29,7 +29,12 @@
   public static final JFR instance = new JFR();
 
   public boolean isAvailable() {
-    return FlightRecorder.isAvailable();
+    try {
+      JFR.class.getClassLoader().loadClass("jdk.jfr.FlightRecorder");
```
laurit, line 33: It might be better to do class available check in static initializer and keep the result in a field so you won't accidentally do it multiple times.

breedx-splk: There are only 2 calls and they're early in the lifecycle, but it's still a good idea...so I did that.  Thanks! 👍🏻

PR #436, opened 2021-09-07
Test that profiling survives collector restart

laurit:

PR #437, opened 2021-09-08
Fix profiler NPE

laurit:
Apparently stack frames can have null methods although javadoc https://docs.oracle.com/javase/9/docs/api/jdk/jfr/consumer/RecordedFrame.html#getMethod-- states otherwise.  I discovered this while investigating why test added in https://github.com/signalfx/splunk-otel-java/pull/436 is flaky.
This pr also adds catching all exceptions to `RecordingSequencer. handleInterval`. This method is invoked periodically by `ScheduledExecutorService` if it fails it will not be invoked again so we should not let it fail.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencer.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencer.java
@@ -54,18 +54,22 @@ class RecordingSequencer {
 
   @VisibleForTesting
   void handleInterval() {
-    if (!recordingEscapeHatch.jfrCanContinue()) {
-      logger.warn("JFR recordings cannot proceed.");
-      if (recorder.isStarted()) {
-        recorder.stop();
+    try {
+      if (!recordingEscapeHatch.jfrCanContinue()) {
+        logger.warn("JFR recordings cannot proceed.");
+        if (recorder.isStarted()) {
+          recorder.stop();
+        }
+        return;
       }
-      return;
+      if (!recorder.isStarted()) {
+        recorder.start();
+        return;
+      }
+      recorder.flushSnapshot();
+    } catch (Throwable throwable) {
+      logger.error("Profiler periodic task failed.", throwable);
     }
-    if (!recorder.isStarted()) {
-      recorder.start();
-      return;
-    }
-    recorder.flushSnapshot();
   }
 
   public static Builder builder() {
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
@@ -46,10 +46,13 @@ public class StackSerializer {
   }
 
   private StringBuilder serializeFrame(StringBuilder sb, RecordedFrame frame) {
+    RecordedMethod method = frame.getMethod();
+    if (method == null) {
+      return sb;
+    }
     if (sb.length() > 0) {
       sb.append("\n");
     }
-    RecordedMethod method = frame.getMethod();
     String className = method.getType().getName();
     String methodName = method.getName();
     int lineNumber = frame.getInt("lineNumber");
--- a/profiler/src/test/java/com/splunk/opentelemetry/profiler/util/StackSerializerTest.java
+++ b/profiler/src/test/java/com/splunk/opentelemetry/profiler/util/StackSerializerTest.java
@@ -34,6 +34,7 @@ class StackSerializerTest {
   RecordedFrame frame2 = makeFrame("io.test.MyClass", "silver", 456);
   RecordedFrame frame3 = makeFrame("io.test.Framewerk", "root", 66);
   List<RecordedFrame> frames = Arrays.asList(frame1, frame2, frame3);
+  List<RecordedFrame> framesWithNullMethod = Arrays.asList(frame1, makeFrameWithNullMethod("io.test.MyClass", 456), frame3);
 
   @Test
   void serialize() {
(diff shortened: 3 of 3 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencer.java
```diff
@@ -54,18 +54,22 @@ public void start() {
...
-    }
-    if (!recorder.isStarted()) {
-      recorder.start();
-      return;
+      if (!recorder.isStarted()) {
+        recorder.start();
+        return;
+      }
+      recorder.flushSnapshot();
+    } catch (Throwable throwable) {
```
mateuszrzeszutek, line 70: Can we catch a `RuntimeException` here? Or is there a need to catch `Error`s too?

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencer.java
laurit: I guess this is open for debate how we should handle it. Anything that gets thrown out of this method break profiling. Would it be ok for it to break when there is an `OOM`? If not then `RuntimeException` wouldn't be enough. I personally would lean towards catching everything as when this thing breaks it will be completely impossible to debug, best to let it never break and log any unexpected exceptions.

mateuszrzeszutek: I mostly thought about `ThreadDeath`, but OOM is also a valid scenario too. Honestly I'd normally expect OOMs to break absolutely everything, but in case of profiling maybe it's worth to just try to ignore them and gather as much info as possible? @breedx-splk

breedx-splk: Yeah, I thought the general rule for this was to almost never catch `Throwable` but rather catch `Exception`. Errors are almost never recoverable and trying to keep going after an `Error` has been thrown is going to be counterproductive (making a bad situation worse).

My recommendation is to catch `Exception` here.

laurit: In a regular application I have never needed to catch a `Throwable`. With javaagnet the thing is that you are running inside someone else's application which is pretty much a hostile territory. You don't know whether their app sometimes ooms. If that happens inside your code then in my opinion it is really good if it is recoverable as otherwise if it ends up as support case it will be really hard to figure out why suff sometimes breaks.

breedx-splk: Ok I think you convinced me. I'm ok with it.

breedx-splk: @johnbley

johnbley: Catching `Throwable` is sometimes OK.  For example, instrumentation might accidentally attempt to allocate a 120 GB array,  causing an "OOM" - the instrumentation can abort, leaving the application unfazed.  Now, you might say, "how in the world could we distinguish between this and a "regular" OOM that is actually the app reaching its limits due to, e.g., a memory leak in the app itself?  If we swallow the OOM the app might not get notified."  Well, if the app is actually near its limit, there will be another OOM coming soon enough.
`ThreadDeath` absolutely cannot be swallowed.  I am annoyed that the current ByteBuddy instrumentation doesn't have a clean way to do the truly paranoid thing which is
```java
try {
   instrumentationStuff()
} catch (ThreadDeath td) {
  throw td;
} catch (Throwable t) {
  try {
      logProblem(t);
  } catch (ThreadDeath td) {
     throw td;
  } catch (Throwable t2) {
        // swallow 
  }
}
```
This pattern hopefully only needs to be in a few very high-level entry points to instrumentation.

laurit: Why do you care about `ThreadDeath` that much? Isn't it only used by `Thread.stop` which has been deprecated for like 20 years. Are there really applications that try to stop random threads?

johnbley: Yes, it's deprecated; yes, developers should not use it; yes, there are not that many that do.  But!  With an extra half dozen lines of code, our instrumentation will not break the small number of apps/systems that do use it.  The first rule of instrumentation is "do no harm".

laurit: As far as I can tell it would only break if user code for some unknown reason decided to find our profiler periodic executor thread which is named `JFR Recording Sequencer` call `stop` on it and then `join` and hang because thread wasn't really stopped. I could understand if this catch Throwable was inserted somewhere in user code and could suppress death of a user thread, but this is inside our agent code. Is this truly something we should worry about? If suppressing `ThreadDeath` is bad then wouldn't swallowing of `InterruptedException` also be bad?

mateuszrzeszutek: @johnbley @laurit @breedx-splk can we merge this PR?

laurit: @mateuszrzeszutek you and @iNikem have to decide whether to catch `Throwable` or `RuntimeException` or something else

iNikem: Lauri's argument convinced me and I am OK with catching `Throwable` here.

mateuszrzeszutek: I'm fine with that too - nobody's gonna call Thread.stop() on our profiler thread.

mateuszrzeszutek: Okay, there are no objections - I'm hitting the merge button then

PR #438, opened 2021-09-09
Profiler omits some agent internal threads

breedx-splk:
Most users will find little value in these and probably won't want to pay for ingest/collection, so let's omit them.

This is a first-pass. It is expected that we will iterate on the list of things being dropped.

### DIFF
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -12,6 +12,6 @@ It should be considered experimental and is completely unsupported.
 |`splunk.profiler.recording.duration`      | 20s                    | recording unit duration                   |
 |`splunk.profiler.keep-files`              | false                  | leave JFR files on disk id `true`         |
 |`splunk.profiler.logs-endpoint`           | `otel.exporter.otlp.endpoint` or http://localhost:4317  | where to send OTLP logs                   |
-|`splunk.profiler.period.{eventName}`      | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (100ms): `-Dsplunk.profiler.period.threaddump=1000` |
+|`splunk.profiler.period.{eventName}`      | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (1000ms): `-Dsplunk.profiler.period.threaddump=1000` |
 |`splunk.profiler.tlab.enabled`            | false                  | set to `false` to disable TLAB memory events |
 |`splunk.profiler.include.agent.internals` | false                  | set to `true` to include agent internal call stacks |
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -5,12 +5,13 @@ It should be considered experimental and is completely unsupported.
 
 # configuration
 
-| name                                | default                | description                               |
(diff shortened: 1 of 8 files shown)


jkwatson (approved):

laurit (approved):

profiler/README.md
```diff
@@ -5,12 +5,13 @@ It should be considered experimental and is completely unsupported.
...
-|`splunk.profiler.tlab.enabled`       | false                  | set to `false` to disable TLAB memory events |
+| name                                     | default                | description                               |
+|------------------------------------------|------------------------|-------------------------------------------|
+|`splunk.profiler.enabled`                 | false                  | set to true to enable the profiler        |
+|`splunk.profiler.directory`               | "."                    | location of jfr files                     |
+|`splunk.profiler.recording.duration`      | 20s                    | recording unit duration                   |
+|`splunk.profiler.keep-files`              | false                  | leave JFR files on disk id `true`         |
+|`splunk.profiler.logs-endpoint`           | `otel.exporter.otlp.endpoint` or http://localhost:4317  | where to send OTLP logs                   |
+|`splunk.profiler.period.{eventName}`      | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (100ms): `-Dsplunk.profiler.period.threaddump=1000` |
```
mateuszrzeszutek, line 15: Unrelated, but I've just noticed that it's wrong:
```suggestion
|`splunk.profiler.period.{eventName}`      | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (1000ms): `-Dsplunk.profiler.period.threaddump=1000` |
```

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/AgentInternalsFilter.java
```diff
@@ -0,0 +1,44 @@
...
+        "\"BatchSpanProcessor_WorkerThread-",
+        "\"JFR Recorder Thread\"",
+        "\"JFR Periodic Tasks\"",
+        "\"JFR Recording Scheduler\"",
+        "\"JFR Recording Sequencer\""
+      };
+
+  @Override
+  public boolean test(String stack) {
+    if (Arrays.stream(UNWANTED_PREFIXES).anyMatch(stack::startsWith)) {
```
iNikem, line 36: I am concerned about the performance of that as compared to common `for` loop.

mateuszrzeszutek: Nit: you can use a `Trie` here 😄

laurit: Is this worth the effort? If I understand correctly this test is applied once a second to the trace form each thread which doesn't seem like a lot.

breedx-splk: Prefer to do the simple thing first and optimize/microbenchmark later if deemed important.

PR #448, opened 2021-09-16
Add FAQ entry for exporter failure

laurit:
@theletterf Could you review :)

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -14,3 +14,7 @@
 - **How often do you release?** We strive to release the Splunk distribution within 2 working days after
   the [upstream project](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases)
   releases. OpenTelemetry Java generally releases a new version every 4 weeks.
+- **What happens when exporter endpoint is temporarily unreachable?** Sending telemetry will resume when endpoint
+  becomes available again. Telemetry collected meanwhile will get dropped. To avoid dropping data it is recommended to
+  use [Splunk OpenTelemetry Connector](https://github.com/signalfx/splunk-otel-collector) that supports retrying failed
+  transmissions.
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -14,7 +14,5 @@
 - **How often do you release?** We strive to release the Splunk distribution within 2 working days after
   the [upstream project](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases)
   releases. OpenTelemetry Java generally releases a new version every 4 weeks.
-- **What happens when exporter endpoint is temporarily unreachable?** Sending telemetry will resume when endpoint
-  becomes available again. Telemetry collected meanwhile will get dropped. To avoid dropping data it is recommended to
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

docs/faq.md
```diff
@@ -14,3 +14,7 @@
 - **How often do you release?** We strive to release the Splunk distribution within 2 working days after
   the [upstream project](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases)
   releases. OpenTelemetry Java generally releases a new version every 4 weeks.
+- **What happens when exporter endpoint is temporarily unreachable?** Sending telemetry will resume when endpoint
+  becomes available again. Telemetry collected meanwhile will get dropped. To avoid dropping data it is recommended to
+  use [Splunk OpenTelemetry Connector](https://github.com/signalfx/splunk-otel-collector) that supports retrying failed
+  transmissions.
```
theletterf, line 20:
```suggestion
- **What happens when the exporter endpoint is temporarily unreachable?** Sending telemetry resumes when the endpoint becomes available again. Telemetry collected in the meantime is dropped. To avoid dropping data, 
  use the [Splunk OpenTelemetry Connector](https://github.com/signalfx/splunk-otel-collector), which supports retrying failed transmissions.
```

theletterf (changes requested):

docs/faq.md
```diff
@@ -14,3 +14,7 @@
 - **How often do you release?** We strive to release the Splunk distribution within 2 working days after
   the [upstream project](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases)
   releases. OpenTelemetry Java generally releases a new version every 4 weeks.
+- **What happens when the exporter endpoint is temporarily unreachable?** Sending telemetry resumes when the endpoint
+  becomes available again. Telemetry collected in the meantime is dropped. To avoid dropping data, use the
```
breedx-splk, line 18: The `BatchSpanProcessor` will buffer spans for a little bit, so maybe this is a bit vague. Not sure that it matters to capture much more detail, but it's not completely accurate, right?

breedx-splk (approved):
Just want to make sure we give some consideration to the accuracy of the claim and don't set up false expectations. This plays it pretty safe, so I'm good with it...sounds like other people are too.

docs/faq.md
```diff
@@ -14,3 +14,7 @@
...
+  [Splunk OpenTelemetry Connector](https://github.com/signalfx/splunk-otel-collector), which supports retrying failed
```
iNikem, line 19: Wait, this claim about Connector is totally misleading. We have "JVM -> Collector -> Backend".  This FAQ item speaks about "what happens if Collector is down", meaning broken "JVM -> Collector" leg. Connector/Collector retrying works for broken "Collector -> Backend" leg.

iNikem (changes requested):
Connector claim is wrong

docs/faq.md
laurit: Removed collector part, is it ok now?

mateuszrzeszutek (approved):

iNikem (approved):

PR #457, opened 2021-09-22
Fix Collector config

iNikem:


laurit:
Profiling test also has collector conf that has the same issue. The bigger issue is what to do with windows, I believe it runs a different version of collector that probably does not accept the new style of conf.

mateuszrzeszutek (approved):

laurit:
Upstream completely removed collector from smoke tests and sends data directly to fakebackend. I believe we should do the same to simplify things. For example the windows collector version we used was built by upstream and that also got removed.

mateuszrzeszutek:
The agent exports metrics in the signalfx format, the fake backend can't accept them - we have to use the collector as a translation layer.

iNikem:
Should we update https://github.com/signalfx/splunk-otel-java/pull/190 to contrib and use it?

mateuszrzeszutek:
~~Is there a contrib windows image?~~ Looks like there's no contrib Windows image: https://github.com/open-telemetry/opentelemetry-collector/issues/2715 😢 
We'd have to build it ourselves

iNikem:
> ~Is there a contrib windows image?~ Looks like there's no contrib Windows image: [open-telemetry/opentelemetry-collector#2715](https://github.com/open-telemetry/opentelemetry-collector/issues/2715) 😢
> We'd have to build it ourselves

@mateuszrzeszutek isn't your #190 exactly about that?

mateuszrzeszutek:
Right... that's strange; I was almost certain that we're using splunk-otel-collector in linux tests and that's why I chose it for Windows too - but we're actually using contrib everywhere.
Anyway, I'll update that PR so that it builds a contrib-based image - let's use the same collector everywhere.

PR #458, opened 2021-09-22
Set collector version

laurit:
The collector configuration that we use does not work with the latest version.


mateuszrzeszutek:
Is this PR superseded by https://github.com/signalfx/splunk-otel-java/pull/457 ?

laurit:
@mateuszrzeszutek this is an alternative. If #457 works then lets use that, if it doesn't then this one has a better chance of just working so we could merge this one while #457 is fixed.

mateuszrzeszutek (approved):

PR #463, opened 2021-09-23
WebLogic thread pool metrics

laurit:


instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/instrumentation/weblogic/metrics/ThreadPoolMetrics.java
```diff
@@ -0,0 +1,75 @@
...
+        THREADS_IDLE.create(
+            tags,
+            () ->
+                getNumberAttribute(objectName, "ExecuteThreadIdleCount").applyAsDouble(mBeanServer)
+                    + getNumberAttribute(objectName, "StandbyThreadCount")
+                        .applyAsDouble(mBeanServer)));
+  }
+
+  private static Tags executorTags(ObjectName objectName) {
+    return Tags.of(Tag.of(EXECUTOR_TYPE, "weblogic"), Tag.of(EXECUTOR_NAME, objectName.toString()));
```
laurit, line 71: Is using `objectName.toString()` as executor name ok?

smoke-tests/src/test/java/com/splunk/opentelemetry/WebLogicSmokeTest.java
```diff
@@ -56,6 +59,11 @@ public void webLogicSmokeTest(TestImage image, ExpectedServerAttributes serverAt
     // creates spans
     assertWebAppTrace(serverAttributes);
 
+    // TODO: Windows collector image cannot accept signalfx metrics right now, don't assert them
+    if (!image.isWindows()) {
```
laurit, line 63: will remove it when windows collector changes is merged

instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/instrumentation/weblogic/metrics/ThreadPoolMetrics.java
mateuszrzeszutek: If there's no better candidate, sure. The main requirement here is that it should be unique -- so that it's possible to distinguish between different pools.

instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/instrumentation/weblogic/metrics/ThreadPoolMetrics.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private static List<Meter> createMeters(MBeanServer mBeanServer, ObjectName objectName) {
+    Tags tags = executorTags(objectName);
+    return asList(
+        THREADS_CURRENT.create(
+            tags, mBeanServer, getNumberAttribute(objectName, "ExecuteThreadTotalCount")),
+        TASKS_COMPLETED.create(
+            tags,
+            () ->
```
mateuszrzeszutek, line 57: How about adding a
```java
  public <T> FunctionCounter create(Tags additionalTags, T obj, ToDoubleFunction<T> function)
```
function to `CounterSemanticConvention`? Similar to what `GaugeSemanticConvention` has

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

PR #465, opened 2021-09-23
TLAB events are only relevant if enabled in config

breedx-splk:
If TLAB events end up in the file (through interference from other recordings, for example), we still don't want to treat them as relevant. This guards it with config.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/events/RelevantEvents.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/events/RelevantEvents.java
@@ -16,20 +16,37 @@
 
 package com.splunk.opentelemetry.profiler.events;
 
+import com.splunk.opentelemetry.profiler.Configuration;
 import com.splunk.opentelemetry.profiler.TLABProcessor;
 import com.splunk.opentelemetry.profiler.ThreadDumpProcessor;
+import io.opentelemetry.instrumentation.api.config.Config;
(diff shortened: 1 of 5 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/events/RelevantEvents.java
```diff
@@ -16,20 +16,34 @@
...
-              Arrays.asList(
-                  ThreadDumpProcessor.EVENT_NAME,
-                  ContextAttached.EVENT_NAME,
-                  TLABProcessor.NEW_TLAB_EVENT_NAME,
-                  TLABProcessor.OUTSIDE_TLAB_EVENT_NAME)));
+
+  private final Set<String> eventNames;
+
+  private RelevantEvents(Set<String> eventNames) {
+    this.eventNames = eventNames;
```
jkwatson, line 33: any need for defensive copying here?

jkwatson: nm. it's private.

jkwatson:
Next step will be to not bother sorting TLABs at all, correct?

breedx-splk:
> Next step will be to not bother sorting TLABs at all, correct?

Yup, follow up PR to keep things small and easily reviewable!  🤘🏻

jkwatson (approved):
:shipit:

PR #467, opened 2021-09-24
Exclude TLAB events from the sorting pipeline

breedx-splk:
TLAB events are high volume, and to reduce resource usage we think we should exclude them from the sorting process. This implies a change in lifecycle, so that we know when we have all non-tlab events so that we can sort. The `EventProcessingChain` uses a `PriorityQueue` to keep the events ordered (by start time) as they are encountered (O(log(n)).


profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
```diff
@@ -39,15 +44,36 @@ void accept(RecordedEvent event) {
...
         break;
     }
   }
+
+  /**
+   * Tells the processing chain that a work unit (JFR file) is complete and it can process what's in
+   * the buffer. After flushing, the buffer will be empty.
+   */
+  public void flushBuffer() {
+    buffer.forEach(dispatchContextualizedThreadDumps());
```
jkwatson, line 62: interesting that we do these one at a time. Could any efficiencies be gained by dumping the whole buffer at once?

breedx-splk: I'm not sure I follow. Interesting that we do _what_ one at a time? File handling? Event handling? 

The buffer has to be replayed sequentially in order to know what threads where handling which spans at the time of a thread dump (the `SpanContextualizer`)....I don't think that can be done in bulk or parallelized.

jkwatson: There's a `removeAll` on the PriorityQueue. If you do that, I wonder if the queue becomes available more quickly, than if you pull the items one at a time. That's all I meant.

breedx-splk: Ah, ok. Seems like a micro optimization if anything.

jkwatson: yeah... it's just a thought. Might not apply to anything here, but there might be optimizations internally that we're missing out on if we pull them externally one at a time. Should be benchmarked before being changed, obviously.

jkwatson (approved):

PR #471, opened 2021-09-27
Add muzzle dependency

laurit:
Fixes nightly build


buildSrc/src/main/kotlin/splunk.muzzle-conventions.gradle.kts
```diff
@@ -14,4 +14,6 @@ dependencies {
   add("muzzleTooling", "ch.qos.logback:logback-classic:1.2.5")
   add("muzzleBootstrap", project(":bootstrap"))
   add("muzzleBootstrap", "io.opentelemetry:opentelemetry-api")
+
+  implementation("io.opentelemetry.javaagent:opentelemetry-muzzle")
```
iNikem, line 18: I think we should add this to `codegen` configuration, not `implementation`: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/335d1cd784754b4541bbc11f21252ef6ddba17f1/conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts#L19

iNikem (approved):

PR #478, opened 2021-09-30
Improve jdk.ThreadDump content parsing

breedx-splk:
The single large string instance obtained from the `jdk.ThreadDump` event has to be parsed so that we can send separate stack traces to ingest. Previously, we just simply called `.split("\n\n")` on this. The regex might be computationally expensive, whatever, but we definitely know that the split is allocating a lot of strings that get thrown away immediately when agent internal threads are omitted (default).

So, this creates a stateful parser to help handle this. We only create those smaller substrings in the event that we can parse a stack and pass the filters. This also adds a few more thread names to the filter list.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/AgentInternalsFilter.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/AgentInternalsFilter.java
@@ -16,29 +16,51 @@
 
 package com.splunk.opentelemetry.profiler;
 
-import java.util.Arrays;
-import java.util.function.Predicate;
+import java.util.stream.Stream;
 
-public class AgentInternalsFilter implements Predicate<String> {
+public class AgentInternalsFilter {
 
-  private static final String[] UNWANTED_PREFIXES =
+  static final String[] UNWANTED_PREFIXES =
       new String[] {
         "\"Batched Logs Exporter\"",
         "\"BatchSpanProcessor_WorkerThread-",
         "\"JFR Recorder Thread\"",
         "\"JFR Periodic Tasks\"",
         "\"JFR Recording Scheduler\"",
-        "\"JFR Recording Sequencer\""
+        "\"JFR Recording Sequencer\"",
+        "\"Reference Handler\"",
+        "\"Finalizer\"",
+        "\"C1 CompilerThread",
+        "\"Common-Cleaner\""
       };
+  private final boolean includeAgentInternals;
 
-  @Override
-  public boolean test(String stack) {
-    if (Arrays.stream(UNWANTED_PREFIXES).anyMatch(stack::startsWith)) {
+  public AgentInternalsFilter(boolean includeAgentInternals) {
+    this.includeAgentInternals = includeAgentInternals;
+  }
+
+  public boolean test(String wallOfStacks, int startIndex, int lastIndex) {
+    if (lastIndex == -1) {
       return false;
     }
-    if (stack.indexOf('\n') == -1) {
+    // Must start with a quote for the thread name
+    if (wallOfStacks.charAt(startIndex) != '"') {
       return false;
     }
-    return true;
+    // If the last newline before next is before the start, that means we have one line, so skip
+    // that
+    int previousNewlineIndex = wallOfStacks.lastIndexOf('\n', lastIndex - 1);
+    if (previousNewlineIndex <= startIndex) {
+      return false;
+    }
+    // two line cases
+    if (wallOfStacks.lastIndexOf('\n', previousNewlineIndex - 1) <= startIndex) {
+      return false;
+    }
+    if (includeAgentInternals) {
+      return true;
+    }
+    return Stream.of(AgentInternalsFilter.UNWANTED_PREFIXES)
+        .noneMatch(prefix -> wallOfStacks.regionMatches(startIndex, prefix, 0, prefix.length()));
   }
 }
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -108,7 +108,7 @@ public class JfrActivator implements AgentListener {
         new StackToSpanLinkageProcessor(logEntryCreator, batchingLogsProcessor);
 
     ThreadDumpToStacks threadDumpToStacks =
-        new ThreadDumpToStacks(buildAgentInternalsFilter(config));
+        new ThreadDumpToStacks(buildStackTraceFilter(config));
 
     ThreadDumpProcessor threadDumpProcessor =
         buildThreadDumpProcessor(spanContextualizer, processor, threadDumpToStacks);
@@ -162,7 +162,7 @@ public class JfrActivator implements AgentListener {
   }
 
   /** May filter out agent internal call stacks based on the config. */
-  private StackTraceFilter buildAgentInternalsFilter(Config config) {
+  private StackTraceFilter buildStackTraceFilter(Config config) {
     return new StackTraceFilter(config.getBoolean(CONFIG_KEY_INCLUDE_INTERNALS, false));
   }
 
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -41,7 +41,6 @@ import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Consumer;
-import java.util.function.Predicate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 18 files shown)


iNikem:
I think that replacing simple regex split with such complicated algorithm requires fact-based justification. Can you please add a jmh benchmark demonstrating that we actually have a noticeable improvement here?

Also, please verify and confirm that existing tests cover all branches of `AgentInternalsFilter.test` and `ThreadDumpToStacks`.

profiler/src/test/java/com/splunk/opentelemetry/profiler/AgentInternalsFilterTest.java
```diff
@@ -24,10 +24,10 @@
```
mateuszrzeszutek, line 24: This class seems to be missing some test cases:
* `includeAgentInternals = true`
* the two line case
* wrong arguments, i.e. string not starting with `"` or last index == -1
* start/end different from `0, stack.length()`

profiler/src/main/java/com/splunk/opentelemetry/profiler/AgentInternalsFilter.java
```diff
@@ -16,29 +16,51 @@
 
 package com.splunk.opentelemetry.profiler;
 
-import java.util.Arrays;
-import java.util.function.Predicate;
+import java.util.stream.Stream;
 
-public class AgentInternalsFilter implements Predicate<String> {
+public class AgentInternalsFilter {
```
mateuszrzeszutek, line 21: It no longer just "agent internals" filter, this class is not responsible for general stack trace filtering. WDYT about `StackTraceFilter`?

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpToStacks.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+      }
+    }
+
+    private String advanceNextStack() {
+      String result = wallOfStacks.substring(start, next);
+      if (next >= wallOfStacks.length() - 2) {
+        done = true;
+      }
+      start = next + 2;
```
mateuszrzeszutek, line 94: This 2 here is because of `"\n\n".length()`?

breedx-splk:
adding jmh and hitting edges.

profiler/src/main/java/com/splunk/opentelemetry/profiler/AgentInternalsFilter.java
breedx-splk: Sounds good.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpToStacks.java
breedx-splk: Yeah, the next chunk has to start looking after the `\n\n`.

breedx-splk:
> I think that replacing simple regex split with such complicated algorithm requires fact-based justification. Can you please add a jmh benchmark demonstrating that we actually have a noticeable improvement here?
> 
> Also, please verify and confirm that existing tests cover all branches of `AgentInternalsFilter.test` and `ThreadDumpToStacks`.

Ok, it took me quite some time to get a benchmark running with jmh, and now that I have data, I either don't know how to read the results or I've just made things worse (no actual improvement).

I ran `./gradlew testing:jmh-benchmarks:clean testing:jmh-benchmarks:jmh` and got this output:

```
Benchmark                                                                          Mode  Cnt      Score      Error   Units
ThreadDumpProcessorBenchmark.newThreadDumpProcessor                               thrpt   20  47654.113 ± 1266.723   ops/s
ThreadDumpProcessorBenchmark.newThreadDumpProcessor:·gc.alloc.rate                thrpt   20   2415.975 ±   86.159  MB/sec
ThreadDumpProcessorBenchmark.newThreadDumpProcessor:·gc.alloc.rate.norm           thrpt   20  55820.110 ±  571.208    B/op
ThreadDumpProcessorBenchmark.newThreadDumpProcessor:·gc.churn.G1_Eden_Space       thrpt   20   2418.887 ±   87.175  MB/sec
ThreadDumpProcessorBenchmark.newThreadDumpProcessor:·gc.churn.G1_Eden_Space.norm  thrpt   20  55887.646 ±  654.869    B/op
ThreadDumpProcessorBenchmark.newThreadDumpProcessor:·gc.churn.G1_Old_Gen          thrpt   20      0.007 ±    0.001  MB/sec
ThreadDumpProcessorBenchmark.newThreadDumpProcessor:·gc.churn.G1_Old_Gen.norm     thrpt   20      0.152 ±    0.027    B/op
ThreadDumpProcessorBenchmark.newThreadDumpProcessor:·gc.count                     thrpt   20   1672.000             counts
ThreadDumpProcessorBenchmark.newThreadDumpProcessor:·gc.time                      thrpt   20    818.000                 ms
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor                               thrpt   20   7104.011 ± 1856.952   ops/s
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor:·gc.alloc.rate                thrpt   20    433.509 ±  112.051  MB/sec
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor:·gc.alloc.rate.norm           thrpt   20  67276.673 ±  217.422    B/op
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor:·gc.churn.G1_Eden_Space       thrpt   20    434.048 ±  111.781  MB/sec
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor:·gc.churn.G1_Eden_Space.norm  thrpt   20  67417.861 ± 2075.981    B/op
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor:·gc.churn.G1_Old_Gen          thrpt   20      0.011 ±    0.005  MB/sec
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor:·gc.churn.G1_Old_Gen.norm     thrpt   20      1.531 ±    0.483    B/op
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor:·gc.count                     thrpt   20    300.000             counts
ThreadDumpProcessorBenchmark.oldThreadDumpProcessor:·gc.time                      thrpt   20    157.000                 ms
```

Maybe somebody can look at this and confirm what I'm seeing...and even better, maybe somebody can reproduce these results (they took about 15mins on my laptop).

iNikem:
```
Benchmark                                                                          Mode  Cnt      Score      Error   Units
newThreadDumpProcessor                               thrpt   20  47654.113 ± 1266.723   ops/s
oldThreadDumpProcessor                               thrpt   20   7104.011 ± 1856.952   ops/s

newThreadDumpProcessor:·gc.alloc.rate.norm           thrpt   20  55820.110 ±  571.208    B/op
oldThreadDumpProcessor:·gc.alloc.rate.norm           thrpt   20  67276.673 ±  217.422    B/op
```

new processor is almost 7 times faster and allocations 18% less memory :) Looks like success to me?

breedx-splk:
The non-normalized `gc.alloc.rate` is the other way around though. 

The speed does indeed look like an improvement, but not the gc count?

I think I probably don't know how to read this output probably.

mateuszrzeszutek:
It's MB/sec, so I'm guessing that since the new algorithm allows much higher throughput more operations (= more allocations) execute in a second.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -148,17 +150,19 @@ private void activateJfrAndRunForever(Config config) {
...
+        .build();
   }
 
   /** May filter out agent internal call stacks based on the config. */
-  private Predicate<String> buildAgentInternalsFilter(Config config) {
-    if (config.getBoolean(CONFIG_KEY_INCLUDE_INTERNALS, false)) {
-      return x -> true;
-    }
-    return new AgentInternalsFilter();
+  private StackTraceFilter buildAgentInternalsFilter(Config config) {
```
mateuszrzeszutek, line 164:
```suggestion
  private StackTraceFilter buildStackTraceFilter(Config config) {
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpToStacks.java
mateuszrzeszutek: Can you add a comment?

testing/jmh-benchmarks/src/main/java/com/splunk/opentelemetry/profiler/old/AgentInternalsFilter.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.splunk.opentelemetry.profiler;
+package com.splunk.opentelemetry.profiler.old;
```
mateuszrzeszutek, line 17: To be removed?

breedx-splk: The benchmark is intended to run the old version with the new version. We can't get rid of it unless we want to get rid of the benchmark too.

mateuszrzeszutek: Oh, I see - I didn't notice that it was in the benchmarks module.
👍

mateuszrzeszutek (approved):

PR #488, opened 2021-10-05
Ignore TLAB events if we didn't enable them

laurit:
JFR sends all enabled events to all recordings so we'll get TLAB events event when we didn't enable them but some other recoding did.

### DIFF
--- a/profiler/src/test/java/com/splunk/opentelemetry/profiler/TLABProcessorTest.java
+++ b/profiler/src/test/java/com/splunk/opentelemetry/profiler/TLABProcessorTest.java
@@ -28,6 +28,7 @@ import com.splunk.opentelemetry.logs.LogEntry;
 import com.splunk.opentelemetry.logs.LogsProcessor;
 import com.splunk.opentelemetry.profiler.events.EventPeriods;
 import com.splunk.opentelemetry.profiler.util.StackSerializer;
+import io.opentelemetry.instrumentation.api.config.Config;
 import java.time.Instant;
 import java.util.concurrent.atomic.AtomicReference;
 import jdk.jfr.EventType;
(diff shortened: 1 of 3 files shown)


breedx-splk:
They should be filtered out in the event pipeline. See https://github.com/signalfx/splunk-otel-java/blob/main/profiler/src/main/java/com/splunk/opentelemetry/profiler/events/RelevantEvents.java#L39 and #465

laurit:
@breedx-splk that is not enough. When there are multiple recordings and some other recording, say from overhead test harness, enables TLAB events then profiler recoding will also get TLAB events.

> 3.1 Note on Running Multiple Recordings
Java Flight Recorder allows many recordings to run concurrently. You can configure each recording by using different settings; in particular, you can configure different recordings to capture different sets of events. However, in order to make the internal logic of Java Flight Recorder as streamlined as possible, the resulting recording always contains the union of all events for all recordings active at that time. This means that if more than one recording is running, you might end up with more information in the recording than you wanted. This can be a little bit confusing but has no other negative implications.

breedx-splk:
> @breedx-splk that is not enough. When there are multiple recordings and some other recording, say from overhead test harness, enables TLAB events then profiler recoding will also get TLAB events.

OH you are right...I believe that I broke this in #467. They were separated from the sorting pipeline but I overlooked the filtering in this case. Thanks for recognizing this!

profiler/src/test/java/com/splunk/opentelemetry/profiler/TLABProcessorTest.java
```diff
@@ -87,7 +92,10 @@ private void tlabProcessTest(Long tlabSize) {
     when(eventType.getName()).thenReturn("tee-lab");
     when(serializer.serialize(stack)).thenReturn(stackAsString);
 
-    TLABProcessor processor = new TLABProcessor(serializer, consumer, commonAttrs);
+    Config config = mock(Config.class);
+    when(config.getBoolean(Configuration.CONFIG_KEY_TLAB_ENABLED)).thenReturn(true);
```
breedx-splk, line 96: Please add coverage for the disabled case.

breedx-splk (changes requested):
Looks good, thanks again. Just a small ask for test coverage in the disabled case. 👍🏻

breedx-splk (approved):
Thanks again!

PR #495, opened 2021-10-08
Flush events after each chunk

laurit:


laurit:
@breedx-splk without tlab events I get 3 chunks.

PR #516, opened 2021-10-29
Support older Micrometer versions (<1.5)

mateuszrzeszutek:


mateuszrzeszutek:
> do we have any tests to make sure this works?

Not yet -- I haven't managed to successfully write a test for that, only tested it manually. I still haven't given up though

mateuszrzeszutek:
Ouch, looks like micrometer changed the default constructor of `LongTaskTimer$Sample` in 1.7:

```
java.lang.NoSuchMethodError: io.micrometer.core.instrument.LongTaskTimer$Sample: method 'void <init>()' not found
```

This will require a separate 1.0 and 1.7 instrumentations...

At least I figured out how to test that.

mateuszrzeszutek:
Oh well, separate 1.3 and 1.5 instrumentations -- 1.0 doesn't even compile, and 1.1 and 1.2 are missing a few methods.

PR #518, opened 2021-11-03
Update project to GDI spec version 1.1.0

mateuszrzeszutek:
Turns out it was just the max attribute value length, we already had implemented everything else that was introduced in 1.1.0.

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
@@ -25,7 +25,7 @@ import io.opentelemetry.sdk.trace.SpanLimits;
 @AutoService(SdkTracerProviderConfigurer.class)
 public class SplunkTracerProviderConfigurer implements SdkTracerProviderConfigurer {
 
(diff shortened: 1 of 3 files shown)


custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
```diff
@@ -25,6 +25,8 @@
 @AutoService(SdkTracerProviderConfigurer.class)
 public class SplunkTracerProviderConfigurer implements SdkTracerProviderConfigurer {
 
+  static final int SPLUNK_DEFAULT_ATTRIBUTE_VALUE_LENGTH = 12_000;
```
breedx-splk, line 28:
```suggestion
  static final int SPLUNK_MAX_ATTRIBUTE_VALUE_LENGTH = 12_000;
```

breedx-splk (changes requested):
Small naming ask.

theletterf (approved):

custom/src/main/java/com/splunk/opentelemetry/SplunkTracerProviderConfigurer.java
mateuszrzeszutek: 👍 
Done

breedx-splk (approved):

PR #529, opened 2021-11-12
Add websphere smoke test, update other smoke test images

laurit:


smoke-tests/src/test/java/com/splunk/opentelemetry/SmokeTest.java
```diff
@@ -45,6 +45,11 @@
 
   private TelemetryRetriever telemetryRetriever;
 
+  /** Subclasses can override this method to pass jvm arguments in another environment variable */
```
breedx-splk, line 48: I would greatly prefer to pass this in via constructor than using inheritance for configuration.

breedx-splk (approved):
Looks good to me, but I do think that using polymorphic inheritance for configuration is gross.

smoke-tests/src/test/java/com/splunk/opentelemetry/SmokeTest.java
iNikem: Constructors are called by testing framework, no?

iNikem (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/SmokeTest.java
laurit: I guess passing it via constructor could work, but as there are already 3 methods for altering configuration we might need to have a separate object for holding configuration options. Or instead of passing it via constructor we could have a configure method like in https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/7c038f5c715e055d6a24e7de6de1f6fb9af4bf87/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java#L1074 which subclasses could override.
Anyway I don't really share your disgust with doing configuration this way, as long as there aren't too many configuration options it will work out nicely. As otel is using the same approach following it makes it easier to copy improvements like this from otel.

mateuszrzeszutek (approved):

smoke-tests/src/test/java/com/splunk/opentelemetry/SmokeTest.java
mateuszrzeszutek: I really like the [`HttpClientTestOptions`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/HttpClientTestOptions.java) idea that Anuraag has implemented in the upstream. I'm fine with the current version though, we don't have a very complex class hierarchy and it's rather managable.

PR #530, opened 2021-11-12
Profiler: Log the config for the correct key, and do it as duration

breedx-splk:
The jdk prefix should have been omitted, so we were logging the wrong value even when the user followed the docs.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/ConfigurationLogger.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/ConfigurationLogger.java
@@ -45,7 +45,7 @@ public class ConfigurationLogger {
     log(CONFIG_KEY_INGEST_URL, config::getString);
     log(CONFIG_KEY_OTEL_OTLP_URL, config::getString);
     log(CONFIG_KEY_TLAB_ENABLED, config::getBoolean);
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek (approved):

iNikem (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/ConfigurationLogger.java
```diff
@@ -45,7 +45,7 @@ public void log(Config config) {
     log(CONFIG_KEY_INGEST_URL, config::getString);
     log(CONFIG_KEY_OTEL_OTLP_URL, config::getString);
     log(CONFIG_KEY_TLAB_ENABLED, config::getBoolean);
-    log(CONFIG_KEY_PERIOD_PREFIX + "." + "jdk.threaddump", config::getString);
+    log(CONFIG_KEY_PERIOD_PREFIX + "." + "threaddump", config::getDuration);
     logger.info("-----------------------");
```
laurit, line 49: Actually I'm not sure at all we should be doing this. What makes profiling so special that its configuration is logged to stdout while nothing else does this?

breedx-splk: Would you be amenable to agreeing to this change and filing a new issue for us to discuss?

laurit (approved):

PR #533, opened 2021-11-15
Add configuration instructions for websphere

laurit:
@theletterf please review

### DIFF
--- a/docs/server-instructions.md
+++ b/docs/server-instructions.md
@@ -4,10 +4,11 @@ The following sections show how to add the path to the Splunk OTel agent for Jav
 
 * [JBoss EAP / WildFly](#jboss-eap--wildfly)
 * [Jetty](#jetty)
-* [Glassfish / Payara](#glassfish--payara)
+* [GlassFish / Payara](#glassfish--payara)
 * [Tomcat / TomEE](#tomcat--tomee)
-* [Weblogic](#weblogic)
-* [Websphere Liberty Profile](#websphere-liberty-profile)
+* [WebLogic](#weblogic)
+* [WebSphere Liberty Profile](#websphere-liberty-profile)
+* [WebSphere Traditional](#websphere-traditional)
 
 ## JBoss EAP / WildFly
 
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

docs/server-instructions.md
```diff
@@ -111,3 +112,16 @@ Add the path to the JVM agent to the `jvm.options` file:
    -javaagent:/path/to/splunk-otel-javaagent.jar
    ```
 - Save the file and restart the server.
+
+## WebSphere Traditional
```
theletterf, line 116: Is this the commercial name, @laurit ?

docs/server-instructions.md
```diff
@@ -111,3 +112,16 @@ Add the path to the JVM agent to the `jvm.options` file:
...
+
+1. Open the WebSphere Admin Console
```
theletterf, line 118:
```suggestion
1. Open the WebSphere Admin Console.
```

docs/server-instructions.md
```diff
@@ -111,3 +112,16 @@ Add the path to the JVM agent to the `jvm.options` file:
...
+2. Select **Servers > Server type > WebSphere application servers**.
+3. Click on the desired server.
+4. Select **Java and Process Management > Process Definition**.
+5. Click on **Java Virtual Machine**.
+6. In the **Generic JVM arguments**, enter the path to `splunk-otel-javaagent.jar`:
+   ```
+   -javaagent:/path/to/splunk-otel-javaagent.jar
+   ```
+7. Press **OK**. When asked, save the master configuration and restart the server.
```
theletterf, line 127:
```suggestion
7. Click **OK**. When asked, save the master configuration and restart the server.
```

theletterf: This is the preferred path? Any programmatic equivalent users might prefer?

theletterf (changes requested):

docs/server-instructions.md
laurit: At least docker images use this name. Wikipedia page also mentions "traditional" https://en.wikipedia.org/wiki/IBM_WebSphere_Application_Server

docs/server-instructions.md
laurit: Websphere can be programmatically configured using the `wsadmin` tool that accepts jython and jacl scripts. I would prefer not to describe how exactly this is done because it isn't that simple. Anyway websphere users should be accustomed with hardship. Those that already know about `wsadmin` are able to figure this out themselves.

theletterf (approved):

PR #540, opened 2021-11-18
Begin using upstream logs sdk

breedx-splk:
This will be the first in a series of incremental enhancements that move us in the direction of using logs from upstream (rather than our own implementation).

Two main things in this PR:
* Our core log model object is called `LogEntry` and it is a concrete class with no interfaces. This PR lets that class implement the upstream `LogData` class. In doing this, it made sense to change the individual `spanId` and `traceId` fields to `spanContext`, which also made sense to make the other change....
* Add a new byte field for `traceFlags` into the `ContextAttached` events in order to more accurately reinflate the span context.

Following PRs can migrate usages of `LogEntry` to the interface, and once the direct references are gone we can migrate to the upstream data object.

### DIFF
--- a/profiler/build.gradle.kts
+++ b/profiler/build.gradle.kts
@@ -7,7 +7,7 @@ dependencies {
   compileOnly("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api")
   compileOnly("io.opentelemetry:opentelemetry-semconv")
   implementation("io.opentelemetry.proto:opentelemetry-proto")
-  implementation("io.opentelemetry:opentelemetry-sdk-logs:1.9.0-alpha")
+  implementation("io.opentelemetry:opentelemetry-sdk-logs")
 
   compileOnly("org.slf4j:slf4j-api")
   compileOnly("io.grpc:grpc-netty")
--- a/profiler/build.gradle.kts
+++ b/profiler/build.gradle.kts
@@ -7,6 +7,8 @@ dependencies {
   compileOnly("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api")
   compileOnly("io.opentelemetry:opentelemetry-semconv")
   implementation("io.opentelemetry.proto:opentelemetry-proto")
+  implementation("io.opentelemetry:opentelemetry-sdk-logs:1.9.0-alpha")
+
   compileOnly("org.slf4j:slf4j-api")
   compileOnly("io.grpc:grpc-netty")
   implementation("io.grpc:grpc-netty-shaded")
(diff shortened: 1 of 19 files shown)


iNikem:
Does https://github.com/open-telemetry/opentelemetry-java/pull/3880 impact this PR somehow? Should we wait?

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/SpanContextualizer.java
```diff
@@ -56,7 +59,10 @@ public void updateContext(RecordedEvent event) {
     if (traceId == null || spanId == null) {
       threadSpans.remove(javaThreadId);
     } else {
-      SpanLinkage linkage = new SpanLinkage(traceId, spanId, javaThreadId);
+      TraceFlags traceFlags = TraceFlags.fromByte(event.getByte("traceFlags"));
```
laurit, line 62: you don't need to pass trace flags with event as it can only be `TraceFlags.getSampled()`

profiler/build.gradle.kts
```diff
@@ -7,6 +7,8 @@ dependencies {
   compileOnly("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api")
   compileOnly("io.opentelemetry:opentelemetry-semconv")
   implementation("io.opentelemetry.proto:opentelemetry-proto")
+  implementation("io.opentelemetry:opentelemetry-sdk-logs:1.9.0-alpha")
```
mateuszrzeszutek, line 10: Isn't this present in the otel alpha bom? Do you need to specify version explicitly?

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/SpanContextualizer.java
mateuszrzeszutek: Isn't it safer to pass the byte? Users of this distro can configure a custom sampler

profiler/build.gradle.kts
laurit: This pr was created before the agent was update to 1.9 sdk, perhaps now after rebase the version can be removed.

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/SpanContextualizer.java
breedx-splk: I'm not sure I follow what's being asked. The JFR event has a byte from the original span. The `SpanContext` requires a `TraceFlags` instance. If the user has a custom sampler, we still want to respect the original value and propagate it accordingly, right?

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/SpanContextualizer.java
mateuszrzeszutek: I think the current version is perfectly fine - we should pass that byte, cause it's possible for our users to configure a custom sampler that will sample out some spans.

PR #554, opened 2021-11-30
Remove deprecated resource auto configuration usages

laurit:


iNikem (approved):

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
```diff
@@ -44,6 +41,10 @@ public void beforeAgent(Config config) {
     }
   }
 
+  public int order() {
+    return 1;
```
mateuszrzeszutek, line 45: Can you add a comment explaining the value? E.g. `needs to run after OpenTelemetryInstaller`

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/LogsExporterBuilder.java
```diff
@@ -47,10 +46,8 @@ static LogsExporter fromConfig(Config config) {
     return builder.build();
   }
 
-  private static ResourceLogsAdapter buildResourceLogsAdapter(Config config) {
-    Resource resource =
-        OpenTelemetryResourceAutoConfiguration.configureResource(
-            new ConfigPropertiesAdapter(config));
+  private static ResourceLogsAdapter buildResourceLogsAdapter() {
+    Resource resource = ResourceHolder.getResource();
```
breedx-splk, line 50: It's a bummer to have this static call here that's dependent on external prior dependencies. Can we pass the resource in and have the `JfrActivator` do the static work (and also override `order()` to do the right thing)?

breedx-splk (changes requested):
I like the idea of taking some action here to prevent future breakage when the deprecated calls are removed...and I'm actually ok with the static holder as a stop-gap measure.

However, I do want to make sure that we're aware of [this issue](https://github.com/open-telemetry/opentelemetry-java/issues/3924) and how it relates to this. I think we can expect to get rid of the holder once that work drops -- maybe a comment should be added to clarify that to our future selves. 

Thanks!

breedx-splk (approved):

breedx-splk:
Given the other approvals, I don't want to be the one to hold this up....so I'm going to approve and merge and follow up with a small PR instead.  👍🏻

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
laurit: @mateuszrzeszutek what do you think if I changed `beforeAgent` -> `afterAgent` then `order` wouldn't be needed?

profiler/src/main/java/com/splunk/opentelemetry/profiler/LogsExporterBuilder.java
laurit: I wasn't aware of the issue. I just mimicked what Anuraag did to  x-ray sampler in otel contrib. In `JfrActivator` you won't need to worry about `order()` as it runs in `afterAgent`.

custom/src/main/java/com/splunk/opentelemetry/micrometer/MicrometerInstaller.java
mateuszrzeszutek: I think it's safer to initialize metrics before any instrumentation is applied and tries to use it, I'd probably prefer using `beforeAgent()` and `order()` to be extra safe and avoid any potential timing/race issues.
I guess we can just leave a TODO mentioning https://github.com/open-telemetry/opentelemetry-java/issues/3924 somewhere in the code so that it doesn't get lost.

laurit: Added comment to `order()`. I believe that instead of adding a TODO we can rely on @breedx-splk to keep track of that issue.

profiler/src/main/java/com/splunk/opentelemetry/profiler/LogsExporterBuilder.java
laurit: moved static call to `JfrActivator`

mateuszrzeszutek (approved):

mateuszrzeszutek:
No longer needed because of #575

PR #556, opened 2021-12-01
Use upstream LogData

breedx-splk:
Remove usages of the `LogEntry` class and turn them into `LogData`. This allows us to use `LogData` model interface from upstream. One step closer to reducing our reliance on a bespoke logs implementation.


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -103,15 +107,23 @@ private void activateJfrAndRunForever(Config config) {
             .executorService(exportExecutorService)
             .build();
     batchingLogsProcessor.start();
+
+    LogDataBuilder logDataBuilder = createLogDataBuilder(config);
```
mateuszrzeszutek, line 111: Is it safe to reuse a `LogDataBuilder` this way? It's a mutable object.

breedx-splk: Ugh, no, it's not...good call, and THANK YOU!

mateuszrzeszutek (approved):

PR #562, opened 2021-12-01
introduce MetricsDependentInstrumentationModule

breedx-splk:
I'm torn -- on the one hand we keep seeing this pattern repeat itself and this is a cheesy way to button up some of that duplication...but on the other hand it makes me pretty uncomfortable to add another class to the inheritance tree.

Curious what others think. Good idea or just leave well enough alone?

### DIFF
--- a/instrumentation/c3p0/build.gradle.kts
+++ b/instrumentation/c3p0/build.gradle.kts
@@ -16,6 +16,7 @@ muzzle {
 
 dependencies {
   compileOnly("com.mchange:c3p0:0.9.2")
+  implementation(project(":instrumentation:common"))
 
   testImplementation("com.mchange:c3p0:0.9.2")
 }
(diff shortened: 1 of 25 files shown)


mateuszrzeszutek:
Hmm, we don't seem to save that much code on that.
Maybe, as an alternative, a `SplunkMetrics` class in `:bootstrap` with a single `static boolean enabled()` method would work?

breedx-splk:
> Hmm, we don't seem to save that much code on that. Maybe, as an alternative, a `SplunkMetrics` class in `:bootstrap` with a single `static boolean enabled()` method would work?

Yeah, it's a bunch of boilerplate to still implement those methods tho, even if delegated. 🤷🏻

mateuszrzeszutek:
True that. Okay, I don't have any better idea then - I'm fine with merging this PR, and I won't be sad if you decide to close it without merging.

breedx-splk:
@laurit @iNikem @agoallikmaa @jkwatson any thoughts?

jkwatson:
"prefer delegation to inheritance"?

breedx-splk:
> "prefer delegation to inheritance"?

That would be an exciting change.  :)

laurit:
I'd rename `MetricsDependentInstrumentationModule` -> `MetricsInstrumentationModule`, otherwise looks fine. It does remove a copy-pasted method and as it isn't a public api, we have all the options to change it in the future if we so desire. I think we can merge this.

mateuszrzeszutek (approved):

laurit (approved):

PR #563, opened 2021-12-01
Add config to include jvm internal stacks

breedx-splk:
We need a config item here so that we can compare data volume and overhead with/without jvm internal stacks.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/StackTraceFilter.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/StackTraceFilter.java
@@ -34,9 +34,15 @@ public class StackTraceFilter {
         "\"Common-Cleaner\""
       };
   private final boolean includeAgentInternals;
+  private final boolean includeJVMInternals;
 
   public StackTraceFilter(boolean includeAgentInternals) {
+    this(includeAgentInternals, false);
+  }
+
+  public StackTraceFilter(boolean includeAgentInternals, boolean includeJVMInternals) {
     this.includeAgentInternals = includeAgentInternals;
+    this.includeJVMInternals = includeJVMInternals;
   }
 
   public boolean test(String wallOfStacks, int startIndex, int lastIndex) {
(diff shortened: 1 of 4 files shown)


iNikem (approved):

mateuszrzeszutek (approved):

profiler/src/test/java/com/splunk/opentelemetry/profiler/StackTraceFilterTest.java
```diff
@@ -22,6 +22,20 @@
 
 class StackTraceFilterTest {
 
+  public static final String JVM_INTERNAL_STACK =
```
laurit, line 25: It took me a while to figure out that stack trace is considered to be jvm internal when all frames are in jdk code. Imo this is a bit confusing, perhaps it should be clarified in `Configuration` where the property is declared. From the property name I assumed that it would exclude threads like finalizer thread so seeing a thread named logback-2 here made no sense.
Another question is how well does it serve the purpose of throwing away useless stack, perhaps ignoring threads that are parked or in socket accept would give better results.

breedx-splk: Ok, yeah, I can definitely see how `logback-2` here could be confusing -- I clearly just copied it from elsewhere. I'll replace it in the test here and also add a note in the config and the implementation to clarify.  

The Finalizer (and Reference Handler etc) can be simply matched by the left portion of the thread name, so we do those first (it's cheaper than walking the stack). There will be a follow-up item to make a clearer separation between "jvm internal" and "agent internal" threads.

We're definitely interested in hearing other approaches to removing unhelpful threads, feel free to bring it up with the profiler folks!

profiler/src/main/java/com/splunk/opentelemetry/profiler/StackTraceFilter.java
```diff
@@ -34,9 +34,15 @@
         "\"Common-Cleaner\""
       };
   private final boolean includeAgentInternals;
+  private final boolean includeJVMInternals;
```
mateuszrzeszutek, line 37: Nit: I think we've been strictly following camelCase everywhere, abbreviations included:
```suggestion
  private final boolean includeJvmInternals;
```

breedx-splk: Ok, cool.

mateuszrzeszutek (approved):

PR #571, opened 2021-12-08
[profiler] Log export errors at most once per hour

### LINKED ISSUE #566: Profiler only logs collector connectivity issues as DEBUG
This came up in a customer troubleshooting session. No data was coming to the collector, but no errors were displayed in the agent logs. After debug logs were turned on via `-Dotel.javaagent.debug=true`, the problem became clear. As it turned out, the endpoint was specified as `https://localhost:4317` (note the protocol)

```
Caused by: io.grpc.netty.shaded.io.netty.handler.ssl.NotSslRecordException: not an SSL/TLS record: 000006040000000000000500004000
```

To make future troubleshooting faster, this sort of errors should be logged at the WARN level. To avoid polluting the logs too much, consider logging it just once per run or per hour. We should also very non-ambiguously mark the log messages as coming from the Splunk Java Agent, not from the instrumented application itself.

breedx-splk:
Resolves #566

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/logs/OtlpLogsExporter.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/logs/OtlpLogsExporter.java
@@ -26,11 +26,11 @@ import io.opentelemetry.proto.collector.logs.v1.ExportLogsServiceResponse;
 import io.opentelemetry.proto.collector.logs.v1.LogsServiceGrpc;
 import io.opentelemetry.proto.logs.v1.ResourceLogs;
 import io.opentelemetry.sdk.logs.data.LogData;
+import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


profiler/src/main/java/com/splunk/opentelemetry/logs/OtlpLogsExporter.java
```diff
@@ -79,7 +87,31 @@ public void onSuccess(@Nullable ExportLogsServiceResponse result) {
...
     public void onFailure(Throwable th) {
-      logger.debug("Failed to export logs", th);
+      errorLogger.log("Splunk profiling agent failed to export logs", th);
+    }
+  }
+
+  @VisibleForTesting
+  static class OncePerHourLogger {
+    private final AtomicReference<Instant> lastErrorTime = new AtomicReference<>(Instant.EPOCH);
+    private final Supplier<Instant> clock;
```
mateuszrzeszutek, line 97: You can just use `Clock` instead - and call `clock.instant()`

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/logs/OtlpLogsExporter.java
```diff
@@ -79,7 +87,31 @@ public void onSuccess(@Nullable ExportLogsServiceResponse result) {
 
     @Override
     public void onFailure(Throwable th) {
-      logger.debug("Failed to export logs", th);
+      errorLogger.log("Splunk profiling agent failed to export logs", th);
```
laurit, line 90: In my opinion logging this with warn or error is enough, no need to bother with logging once an hour. Firstly isn't this class going to be replaced with logs exporter from otel? If so then all changes here would be lost when the switch happens. Secondly exporting logs should not be special cased. If exporting traces and metrics logs failures on every export then, for consistency, exporting logs should behave the same way or the same changes should be made to all exporters. Thirdly if there are intermittent failures this will allow logging such failure only once an hour, which would be confusing.

breedx-splk: Yeah, this is a valid point. I tried to point this out this same thing in #566. 
@gsmirnov-splk what do you think?

gsmirnov-splk: I do not have a strong opinion here. I only suggested logging once per hour/per run because there were prior verbosity concerns cited.

breedx-splk: Ok, well I'm in favor of simpler/less code, so I'll just remove all the time-based guarding and just log at warn. We can relax it when folks complain.

laurit (approved):

PR #573, opened 2021-12-10
Add profiler memory.enabled option

agoallikmaa:
Added `splunk.profiler.memory.enabled` option. Currently the only thing this option does is change the default value of `splunk.profiler.tlab.enabled` option.

Changed the behavior of `ConfigurationLogger` to get value of `splunk.profiler.logs-endpoint` from `getConfigUrl` to reflect that the config option value is inherited in case it is `null`. Same is done for the `splunk.profiler.tlab.enabled` and `splunk.profiler.memory.enabled`.

### DIFF
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -49,7 +49,8 @@ property or as environment variables (by making them uppercase and replacing dot
 |`splunk.profiler.keep-files`              | false                  | leave JFR files on disk id `true`         |
 |`splunk.profiler.logs-endpoint`           | `otel.exporter.otlp.endpoint` or http://localhost:4317  | where to send OTLP logs                   |
 |`splunk.profiler.period.{eventName}`      | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (1000ms): `-Dsplunk.profiler.period.threaddump=1000` |
-|`splunk.profiler.tlab.enabled`            | false                  | set to `false` to disable TLAB memory events |
+|`splunk.profiler.memory.enabled`          | false                  | set to `true` to enable all other memory profiling options unless explicitly disabled |
+|`splunk.profiler.tlab.enabled`            | `splunk.profiler.memory.enabled` | set to `true` to enable TLAB events even if `splunk.profiler.memory.enabled` is `false` |
 |`splunk.profiler.include.agent.internals` | false                  | set to `true` to include agent internal call stacks |
 
 # Escape hatch
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
@@ -50,7 +50,6 @@ public class Configuration implements ConfigPropertySource {
     config.put(CONFIG_KEY_RECORDING_DURATION, DEFAULT_RECORDING_DURATION);
     config.put(CONFIG_KEY_KEEP_FILES, "false");
     config.put(CONFIG_KEY_MEMORY_ENABLED, "false");
-    config.put(CONFIG_KEY_TLAB_ENABLED, "false");
     return config;
   }
 
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
@@ -34,6 +34,7 @@ public class Configuration implements ConfigPropertySource {
   public static final String CONFIG_KEY_INGEST_URL = "splunk.profiler.logs-endpoint";
   public static final String CONFIG_KEY_OTEL_OTLP_URL = "otel.exporter.otlp.endpoint";
   public static final String CONFIG_KEY_PERIOD_PREFIX = "splunk.profiler.period";
+  public static final String CONFIG_KEY_MEMORY_ENABLED = "splunk.profiler.memory.enabled";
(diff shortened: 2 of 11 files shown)


breedx-splk (approved):
Looks good to me. Nice work. 👍🏻

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -59,4 +60,12 @@ public static String getConfigUrl(Config config) {
     }
     return ingestUrl;
   }
+
+  public static boolean getTLABEnabled(Config config) {
```
iNikem, line 64: Why static? Having this as instance method will simplify `ConfigurationLogger` as well

agoallikmaa: Implemented it the same way as `getConfigUrl` for consistency. Should I change both?

theletterf (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
iNikem: It certainly does not make sense to me to use a static method and to pass as a parameter an instance of this very same class

profiler/README.md
```diff
@@ -49,7 +49,8 @@ property or as environment variables (by making them uppercase and replacing dot
 |`splunk.profiler.keep-files`              | false                  | leave JFR files on disk id `true`         |
 |`splunk.profiler.logs-endpoint`           | `otel.exporter.otlp.endpoint` or http://localhost:4317  | where to send OTLP logs                   |
 |`splunk.profiler.period.{eventName}`      | n/a                    | customize period (in ms) for a specific jfr event. For example, to set the ThreadDump frequency to 1s (1000ms): `-Dsplunk.profiler.period.threaddump=1000` |
-|`splunk.profiler.tlab.enabled`            | false                  | set to `false` to disable TLAB memory events |
+|`splunk.profiler.memory.enabled`          | false                  | set to `true` to enable all other memory profiling options unless explicitly disabled |
+|`splunk.profiler.tlab.enabled`            | `splunk.profiler.memory.enabled` | set to `true` to enable TLAB events even if `splunk.profiler.memory.enabled` is `false` |
```
theletterf, line 53: @agoallikmaa Table layout seems to be broken here. `splunk.profiler.memory.enabled` is in the `Default` column (lin 53). Could you check?

theletterf (changes requested):

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
breedx-splk: It's easy to miss the fact that `Configuration` is not `Config`. This is a convenience method. Because it's confusing, maybe if the call sites are limited we can just move it closer to the call site(s).

profiler/README.md
agoallikmaa: This is intended, it shows it inherits the value of that (same as for `splunk.profiler.logs-endpoint` a few lines above).

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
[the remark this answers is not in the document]
agoallikmaa: I went for consistency within this repository - since it has `TLABProcessor`.

agoallikmaa: Would not really call this a convenience method since it contains logic that the callsite should not need to know about (which configuration option inherits its value from which). A convenience method by definition should exist for the sole reason of convenience, not encapsulation as this method does.

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -59,4 +60,12 @@ public static String getConfigUrl(Config config) {
...
+    Boolean enabled = config.getBoolean(CONFIG_KEY_TLAB_ENABLED);
+    if (enabled == null) {
+      return config.getBoolean(CONFIG_KEY_MEMORY_ENABLED, false);
+    }
+    return enabled;
```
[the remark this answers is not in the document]
agoallikmaa: Same change for `getConfigUrl` then, I presume?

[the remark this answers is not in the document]
mateuszrzeszutek: Yep!

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
iNikem: > It's easy to miss the fact that `Configuration` is not `Config`. This is a convenience method. Because it's confusing, maybe if the call sites are limited we can just move it closer to the call site(s).

Ou, I totally missed that

iNikem (approved):

theletterf (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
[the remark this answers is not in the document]
agoallikmaa: Done, tests needed some adjusting for this as well.

mateuszrzeszutek (approved):

breedx-splk:
@agoallikmaa Did you see the note about signing the commits?

PR #575, opened 2021-12-13
Don't use deprecated AgentListener methods

mateuszrzeszutek:
Supersedes https://github.com/signalfx/splunk-otel-java/pull/554


breedx-splk (approved):
This is killer. Thanks for getting us updated so quickly. 🍰

PR #581, opened 2021-12-14
[COPY] Add profiler memory.enabled option

breedx-splk:
This is identical to @agoallikmaa's #573 in hopes that we can get it merged sooner rather than later. Feel free to close and accept #573 if the commit signing gets worked out first.


breedx-splk:
Actually, this probably won't work as well since the commits got brought over instead of recreated. Ugh.

PR #586, opened 2021-12-16
Exclude unwanted dependency

laurit:


custom/build.gradle.kts
```diff
@@ -12,8 +12,15 @@ dependencies {
   implementation("io.opentelemetry.contrib:opentelemetry-contrib-samplers")
   implementation("io.opentelemetry:opentelemetry-exporter-jaeger-thrift") {
     exclude("io.opentelemetry", "opentelemetry-sdk")
+    // exclude unwanted dependency
+    // https://issues.apache.org/jira/browse/THRIFT-5375
+    exclude("org.apache.tomcat.embed", "tomcat-embed-core")
```
mateuszrzeszutek, line 17: 🤯

mateuszrzeszutek (approved):

custom/build.gradle.kts
iNikem: Should we/sdk update thrift version to the one where this problem is fixed?

iNikem (approved):

custom/build.gradle.kts
laurit: Tomcat is pulled in by `org.apache.thrift:libthrift` that is used by `io.jaegertracing:jaeger-thrift` that is used by `io.jaegertracing:jaeger-client`. The problem is that although that is fixed in `libthrift` there isn't a version of `jaeger-thrift` that uses the fixed version.

PR #594, opened 2021-12-20
Publish agent as docker image

owais:
This is needed by auto-instrumentation in environments like kubernetes.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -74,6 +74,24 @@ create_gh_release() {
     --title "Release $release_tag"
 }
 
+build_docker_image() {
+  docker build -t splunk-otel-instrumentation-java .
+  docker tag splunk-otel-instrumentation-java quay.io/signalfx/splunk-otel-instrumentation-java:$release_tag
+  docker tag splunk-otel-instrumentation-java quay.io/signalfx/splunk-otel-instrumentation-java:latest
+}
+
+publish_docker_image() {
+  docker push quay.io/signalfx/splunk-otel-instrumentation-java:latest
+  docker push quay.io/signalfx/splunk-otel-instrumentation-java:$release_tag
+}
+
+login_to_quay_io() {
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -74,6 +74,18 @@ create_gh_release() {
...
 }
 
+build_docker_image() {
+  docker build -t splunk-otel-instrumentation-java .
+  docker tag splunk-otel-instrumentation-java quay.io/signalfx/splunk-otel-instrumentation-java:$release_tag
+  docker tag splunk-otel-instrumentation-java quay.io/signalfx/splunk-otel-instrumentation-java:latest
+}
+
+publish_docker_image() {
+  docker push quay.io/signalfx/splunk-otel-instrumentation-java:latest
```
iNikem, line 84: First, don't you need to authenticate to quay before pushing?
Second, you don't push `$release_tag`, is this intentional?

owais: Missed the release tag. I thought all repos are pre-configured for quay access (because of a conversation in Jira) but that doesn't seem to be the case. Will add that.

iNikem:
@owais similar idea was proposed and rejected in the upstream repo as well: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4481 Are you SURE this is a right approach?

breedx-splk (approved):

owais:
This is ready for another review. I'm working on adding the required env vars to CI. @iNikem

breedx-splk (approved):
This seems good to me. I kind of wish the docker image name more closely matched the fact that this is built exclusively for the agent (like `quay.io/signalfx/splunk-otel-javaagent:latest` or something)...but we do tend to use "instrumentation" in quite a few places.

If we have some future separation of agent from instrumentation, this could be confusing. I'm not sure it's worth holding this up tho. 👍🏻

iNikem (approved):

PR #605, opened 2022-01-06
Use upstream LogProcessor interface and exporter

breedx-splk:
This removes a lot of our bespoke log code for the profiler in favor of using the upstream sdk exporter. 

The smoke tests failed locally due to the "main" thread missing, but it also failed on `main` and I even used jstack on the container and also couldn't find a thread named "main" so maybe something has changed with petclinic or it's some other local problem.


breedx-splk:
Hmmm, the smoke test failed here too.

breedx-splk:
I committed a debug/troubleshooting change. Let's try again.

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
```diff
@@ -102,6 +104,7 @@ private void syncDoExport() {
     List<LogData> batchCopy = copyClearBatch();
     if (batchCopy != null) {
       logger.debug("Sync export of batch with size {}.", batchCopy.size());
+      // TODO: Look at responses from this call.
```
laurit, line 107: is this TODO something you wish to keep?

profiler/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ dependencies {
   compileOnly("io.opentelemetry:opentelemetry-semconv")
   implementation("io.opentelemetry.proto:opentelemetry-proto")
   implementation("io.opentelemetry:opentelemetry-sdk-logs")
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp-logs")
```
laurit, line 10: can we now remove `opentelemetry-proto` and `grpc` dependencies or are they still needed?

profiler/src/main/java/com/splunk/opentelemetry/logs/BatchingLogsProcessor.java
breedx-splk: Yeah, I think we should be using those return values....but I didn't want to increase scope in this PR. 
I could create a tracking issue if you'd like.

profiler/build.gradle.kts
breedx-splk: Good call. We don't need those, but there was a hidden transitive dep in the test code on `com.google.common.io.CharStreams` but I was able to remove that and allow the tests to run with java 11, which is a win-win.

PR #608, opened 2022-01-07
Profiler - Fix TLAB stack format

breedx-splk:
The format of stacks coming in for TLAB events didn't conform to GDI spec, so this fixes that up. They should look much more like the other stack traces now. Also added a few more checks in the smoke tests around the data contained in the tlab logs.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/TLABProcessor.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/TLABProcessor.java
@@ -30,15 +30,11 @@ import io.opentelemetry.sdk.logs.LogProcessor;
 import io.opentelemetry.sdk.logs.data.LogData;
 import io.opentelemetry.sdk.logs.data.LogDataBuilder;
 import io.opentelemetry.sdk.resources.Resource;
-
-import java.nio.file.FileSystems;
-import java.nio.file.Path;
 import java.time.Instant;
 import java.util.function.Consumer;
 import jdk.jfr.consumer.RecordedEvent;
 import jdk.jfr.consumer.RecordedStackTrace;
 import jdk.jfr.consumer.RecordedThread;
-import jdk.jfr.consumer.RecordingFile;
 
 public class TLABProcessor implements Consumer<RecordedEvent> {
   public static final String NEW_TLAB_EVENT_NAME = "jdk.ObjectAllocationInNewTLAB";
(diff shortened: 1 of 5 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/TLABProcessor.java
```diff
@@ -86,13 +88,26 @@ public void accept(RecordedEvent event) {
...
+  private String buildThreadLine(RecordedThread thread) {
+    if (thread == null) {
+      return buildThreadLine("unknown", 0);
+    }
+    return buildThreadLine(thread.getJavaName(), thread.getJavaThreadId());
+  }
+
+  private String buildThreadLine(String name, long id) {
+    String sb =
+        "\"" + name + "\"" + " #" + id + " prio=0 os_prio=0 cpu=0ms elapsed=0s tid=0 nid=0 unknown";
```
laurit, line 107: could you just omit `prio=0 os_prio=0 cpu=0ms elapsed=0s tid=0 nid=0 unknown` In my opinion it is best not to send these if we can't provide real values. If this is required by gdi spec then consider relaxing it.

profiler/src/main/java/com/splunk/opentelemetry/profiler/TLABProcessor.java
```diff
@@ -86,13 +88,26 @@ public void accept(RecordedEvent event) {
         LogDataBuilder.create(resource, INSTRUMENTATION_LIBRARY_INFO)
             .setName(PROFILING_SOURCE)
             .setEpoch(time)
-            .setBody(stack)
+            .setBody(threadLine + "\n" + "   java.lang.Thread.State: UNKNOWN\n" + stack)
```
mateuszrzeszutek, line 91: WDYT about moving this concatenation to `buildThreadLine()` (possibly renaming it) to avoid using that additional `StringBuilder`?

breedx-splk: Sure, lemme have a look.

breedx-splk: ok, tried that out. LMK what you think.

profiler/src/main/java/com/splunk/opentelemetry/profiler/TLABProcessor.java
breedx-splk: I've created https://github.com/signalfx/gdi-specification/pull/162 to address this.

mateuszrzeszutek (approved):

PR #618, opened 2022-01-12
Update the upstream instrumentation version and prepare for patch rel…

mateuszrzeszutek:
…ease

Upstream v1.9.2 contains a reactor-netty fix that we need to release.


breedx-splk (approved):
LGTM

iNikem (approved):

PR #621, opened 2022-01-13
Bump micrometer version to 1.8.2

mateuszrzeszutek:


breedx-splk:
Seems ok to me, but we might have caught something in that test breakage (it's a micrometer related test that's failing).

mateuszrzeszutek:
Yep, they had fixed one `LongTaskTimer` bug that affected our tests. Fixed that

breedx-splk (approved):

iNikem (approved):

PR #622, opened 2022-01-14
Disable some unnecessary instrumentations

mateuszrzeszutek:
We don't need the upstream micrometer instrumentation, we're using our own; I also disabled the logging appender instrumentations - we're not exporting the logs signal yet, and they get called for every log statement so it's probably better to turn them off right now.


laurit (approved):

iNikem:
Do we plan to remove our own micrometer instrumentation?

mateuszrzeszutek:
Yep, at some point in time -- when we migrate everything over to the upstream repo.

PR #626, opened 2022-01-17
Remove javaagent-all

laurit:
`splunk-otel-javaagent.jar` and `splunk-otel-javaagent-all.jar` are the same.

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -28,10 +28,10 @@ against the data you used to receive from the SignalFx Java Agent.
 Follow these steps to migrate from the SignalFx Java Agent to the Splunk
 Distribution of OpenTelemetry Java Instrumentation:
 
-1. Download the [latest release](https://github.com/signalfx/splunk-otel-java/releases/latest/download/splunk-otel-javaagent-all.jar)
+1. Download the [latest release](https://github.com/signalfx/splunk-otel-java/releases/latest/download/splunk-otel-javaagent.jar)
    of the Splunk Distribution of OpenTelemetry Java. For example use:
    ```bash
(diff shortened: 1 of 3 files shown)


theletterf (approved):

theletterf:
@mateuszrzeszutek @laurit This will require a docs and GDI UI edit, as we're currently linking to the `-all` version for some reason. I'm concerned there might be integrations or scripts out there that use it. I recommend keeping both files for now until we know about usage (if possible).

mateuszrzeszutek:
Let's just switch to `splunk-otel-javaagent.jar` without removing `splunk-otel-javaagent-all.jar` -- we can deprecate the latter in this release and remove it later. WDYT?

theletterf:
Fine by me—the change in the docs is relatively trivial. It's the scripts and automations I'm more concerned about, though there's no way of getting around that breaking change. Could we run a search in support tickets to peek usage?

mateuszrzeszutek:
> Could we run a search in support tickets to peek usage?

No idea 🤷  Probably we can, but I've no idea if anything will turn up - and if something will, it probably won't be all of -all usages anyway.
It's probably best to deprecate this and just wait for a couple of releases.

Dockerfile
```diff
@@ -1,3 +1,3 @@
 FROM busybox 
 
-ADD dist/splunk-otel-javaagent-all.jar /
+ADD dist/splunk-otel-javaagent.jar /
```
mateuszrzeszutek, line 3: @owais FYI

laurit:
> Let's just switch to `splunk-otel-javaagent.jar` without removing `splunk-otel-javaagent-all.jar` -- we can deprecate the latter in this release and remove it later. WDYT?

I reverted all non doc changes besides cloudfoundlry buildpack where just the download url changes

mateuszrzeszutek (approved):

PR #627, opened 2022-01-17
Enable memory sampler in production code

mateuszrzeszutek:
I found out it was only enabled in tests


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -172,6 +172,25 @@ private ThreadDumpProcessor buildThreadDumpProcessor(
...
+      Resource resource,
+      LogDataCommonAttributes commonAttributes,
+      BatchingLogsProcessor batchingLogsProcessor) {
+
+    int samplerInterval =
+        config.getInt(CONFIG_KEY_MEMORY_SAMPLER_INTERVAL, DEFAULT_MEMORY_SAMPLING_INTERVAL);
+    AllocationEventSampler sampler =
+        samplerInterval > 1 ? new SystematicAllocationEventSampler(samplerInterval) : null;
+
+    return TLABProcessor.builder(config)
```
laurit, line 186: `TLABProcessor.builder` also tries to set up sampling, should it be removed from there?

mateuszrzeszutek: 🤦  nope, I missed that; I'll close this PR.
Thanks!

PR #637, opened 2022-01-19
Sanity check TLAB stacks for span linkage

mateuszrzeszutek:
... and fix it

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
@@ -190,51 +190,22 @@ class EventProcessingChain {
    * change is detected process buffered events and clear the buffer.
    */
   private static class ChunkTracker {
-    private final EventTypeHolder contextAttached = new EventTypeHolder();
-    private final EventTypeHolder threadDump = new EventTypeHolder();
-    private final EventTypeHolder newTlab = new EventTypeHolder();
-    private final EventTypeHolder outsideTlab = new EventTypeHolder();
+
+    private final Map<String, EventType> eventTypes = new HashMap<>();
 
     /** @return true when event belongs to a new chunk */
     boolean isNewChunk(RecordedEvent event) {
       EventType currentEventType = event.getEventType();
       String eventName = currentEventType.getName();
 
-      switch (eventName) {
-        case ContextAttached.EVENT_NAME:
-          return contextAttached.isNewEventType(currentEventType);
-        case ThreadDumpProcessor.EVENT_NAME:
-          return threadDump.isNewEventType(currentEventType);
-        case TLABProcessor.NEW_TLAB_EVENT_NAME:
-          return newTlab.isNewEventType(currentEventType);
-        case TLABProcessor.OUTSIDE_TLAB_EVENT_NAME:
-          return outsideTlab.isNewEventType(currentEventType);
-      }
-      return false;
-    }
-
-    void reset() {
-      contextAttached.reset();
-      threadDump.reset();
-      newTlab.reset();
-      outsideTlab.reset();
-    }
-  }
-
-  private static class EventTypeHolder {
-    private EventType value = null;
-
-    /** @return true when event belongs to a new chunk */
-    boolean isNewEventType(EventType currentEventType) {
-      if (value == null) {
-        value = currentEventType;
-      }
+      eventTypes.putIfAbsent(eventName, currentEventType);
+      EventType oldEventType = eventTypes.get(eventName);
       // each chunk is parsed by a new parser that recreates event type
-      return value != currentEventType;
+      return oldEventType != currentEventType;
     }
 
     void reset() {
-      value = null;
+      eventTypes.clear();
     }
   }
 }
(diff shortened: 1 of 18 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
```diff
@@ -58,12 +58,12 @@ void accept(RecordedEvent event) {
...
-        if (chunkTracker.contextAttached(event)) {
-          flushBuffer();
+        try (EventTimer eventTimer = eventStats.time(eventName)) {
+          spanContextualizer.updateContext(event);
         }
-        buffer.add(event);
         break;
       case ThreadDumpProcessor.EVENT_NAME:
+        // TODO: is that correct?
         if (chunkTracker.threadDump(event)) {
```
mateuszrzeszutek, line 67: @breedx-splk @laurit I noticed that chunking was completely breaking span linkage in TLAB events (after I added it there); we should probably do an all-or-nothing approach here and either include TLABs or remove chunking altogether.

profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
```diff
@@ -204,32 +190,51 @@ public void close() {
    * change is detected process buffered events and clear the buffer.
    */
   private static class ChunkTracker {
-    private EventType contextAttachedEventType = null;
-    private EventType threadDumpEventType = null;
+    private final EventTypeHolder contextAttached = new EventTypeHolder();
```
laurit, line 193: perhaps it would be better to place these in a map where key is event name so we wouldn't need to touch this again when we start consuming more events.

smoke-tests/src/test/java/com/splunk/opentelemetry/ProfilerTlabSanityTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.MountableFile;
+
+// TODO: use SmokeTest as base class
+class ProfilerTlabSanityTest {
```
laurit, line 40: fails on windows

laurit (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
laurit: this got resolved by changing the chunking logic

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/SpanContextualizer.java
```diff
@@ -101,6 +101,10 @@ StackToSpanLinkage link(Instant time, String rawStack, long threadId, RecordedEv
     return new StackToSpanLinkage(time, rawStack, eventName, spanLinkage);
   }
 
+  public SpanLinkage link(long threadId) {
```
breedx-splk, line 104: Bummer now that one of the `link()` methods can return null and the other doesn't. I think it would be nice if they were consistent. ..but I understand that this is probably slightly more efficient (and saves an allocation in the hot path).

smoke-tests/build.gradle.kts
```diff
@@ -37,6 +37,7 @@ tasks {
       "jboss" to listOf("**/JBossEapSmokeTest.*"),
       "jetty" to listOf("**/JettySmokeTest.*"),
       "liberty" to listOf("**/LibertySmokeTest.*"),
+      "profiler" to listOf("**/Profiler*"),
```
breedx-splk, line 40: Eek, was this just not running before? 😬

smoke-tests/src/test/resources/tlab-sanity-test-app/TlabSanityTestApp.java
```diff
@@ -0,0 +1,67 @@
...
+    System.out.println("Done.");
+  }
+
+  private static void instrumentedMethod() throws InterruptedException {
+    long startTime = System.currentTimeMillis();
+    Object[] objects;
+
+    System.out.println("Start creating objects ...");
+    while (System.currentTimeMillis() - startTime < TEST_DURATION_MS) {
+      objects = new Object[2_000_000];
```
breedx-splk, line 58: cool allocator. 😁

breedx-splk (approved):
A couple comments, but looks good to me. Wondering if we should turn off TLABs in the existing profiler smoke test now that we have an explicit tlab test...

smoke-tests/build.gradle.kts
mateuszrzeszutek: It was -- the profiler tests were in a different module (`:testing:profiler-tests`). I put this one into `:smoke-tests` to avoid copy pasting lots of things; and I'm thinking of refactoring the other profiler tests and putting them here too.

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/SpanContextualizer.java
mateuszrzeszutek: Oh, I can return `SpanLinkage.NONE` from here -- I just noticed that.

PR #644, opened 2022-01-25
profiler: Configuration option to only report traces with spans

agoallikmaa:
Adds configuration option `splunk.profiler.tracing.stacks.only`, which if enabled (disabled by default), only sends profiling stack traces that have a span context attached. This is implemented by passing a composable `StackTracePredicate` to `ThreadDumpToStacks` which wraps `TracingStacksOnlyFilter` around the `StackTrackFilter`.

Used this approach because the `ThreadDumpToStacks` implementation already makes sure not to make unnecessary copies, therefore it made sense to do this check also before the stack trace is extracted from the `wallOfStacks`. It adds some complexity to `SpanContextualizer` and `StackDescriptorLineParser` as they now have to be able to check a stack trace that is part of a larger string. This way there should be no additional allocations related to handling stack traces that are not reported, same as with the current filter.


profiler/README.md
```diff
@@ -53,6 +53,7 @@ property or as environment variables (by making them uppercase and replacing dot
 |`splunk.profiler.tlab.enabled`            | `splunk.profiler.memory.enabled` | set to `true` to enable TLAB events even if `splunk.profiler.memory.enabled` is `false` |
 |`splunk.profiler.memory.sampler.interval` | 1                      | set to `2` or larger to enable sampling every Nth allocation event where N is the value of this property |
 |`splunk.profiler.include.agent.internals` | false                  | set to `true` to include agent internal call stacks |
+|`splunk.profiler.tracing.stacks.only`     | false                  | set to `true` to include only stack traces that are linked to a span context |
```
mateuszrzeszutek, line 56: Same question here: should this be added to official docs?

mateuszrzeszutek (approved):
The downside of this approach is that we're looking for the linked span context twice - once in the new filter decorator, second time in `SpanContextualizer`. It's probably not that much overhead, but maybe it's worth changing the implementation a little bit. Here's an idea (`ThreadDumpProcessor` pseudocode):
```java
for (StartEnd startEnd : threadDumpToStacks.parseStacks(stacks)) {
  if (!stackFilter.test(stacks, startEnd)) continue;
  int threadId = parser.parseThreadId(stacks, startEnd);
  var spanLinkage = contextualizer.link(threadId);
  if (spanLinkage == NONE && tracingStacksOnly) continue;
  processor.accept(new StackToSpanLinkage(event, threadId, startEnd.getSubstring(stacks)));
}
```
That'd probably make `SpanContextualizer` a bit simpler (it no longer has to parse anything, or construct `StackToSpanLinkage` objects), but move some of the logic to the `ThreadDumpProcessor`. WDYT?

profiler/README.md
theletterf: Following with interest.

agoallikmaa:
> The downside of this approach is that we're looking for the linked span context twice - once in the new filter decorator, second time in `SpanContextualizer`. It's probably not that much overhead, but maybe it's worth changing the implementation a little bit. Here's an idea (`ThreadDumpProcessor` pseudocode):
> 
> ```java
> for (StartEnd startEnd : threadDumpToStacks.parseStacks(stacks)) {
>   if (!stackFilter.test(stacks, startEnd)) continue;
>   int threadId = parser.parseThreadId(stacks, startEnd);
>   var spanLinkage = contextualizer.link(threadId);
>   if (spanLinkage == NONE && tracingStacksOnly) continue;
>   processor.accept(new StackToSpanLinkage(event, threadId, startEnd.getSubstring(stacks)));
> }
> ```
> 
> That'd probably make `SpanContextualizer` a bit simpler (it no longer has to parse anything, or construct `StackToSpanLinkage` objects), but move some of the logic to the `ThreadDumpProcessor`. WDYT?

I generally like the idea, but would probably not make sense to return an actual list if it can simply be iterated similar to how the stream works at the moment. One option is to make the stream produce objects similar to this `StartEnd`, but we can avoid allocation unnecessary objects with an ad-hoc iterator like this:
```
StringRegion stack = new StringRegion(stacks);
while (threadDumpToStacks.findNext(stack)) {
   // similar content as above
}
```
Where the region would be mutable and would be advanced on each iteration. If this one sounds good, I can proceed with the refactor.

mateuszrzeszutek:
> I generally like the idea, but would probably not make sense to return an actual list if it can simply be iterated similar to how the stream works at the moment.

Yeah, that was just pseudocode - it does not have to be a list at all. Your region idea would be perfectly fine as well 👍

.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
mateuszrzeszutek, line 1: Can you make sure this does not affect .bat files? Just in case 🤞

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpToStacks.java
```diff
@@ -16,84 +16,50 @@
...
+    if (end == -1) {
+      return false;
+    }
+    // Reached the end of the wall, so just set next to the end
+    if (end < start) {
+      end = region.threadDump.length();
+    }
+
+    region.startIndex = start;
+    region.endIndex = end;
```
mateuszrzeszutek, line 62: WDYT about encapsulating this modification in a `move(start, end)` method?

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpToStacks.java
```diff
@@ -16,84 +16,50 @@
...
 package com.splunk.opentelemetry.profiler;
 
-import java.util.Spliterator;
-import java.util.Spliterators;
-import java.util.function.Consumer;
-import java.util.stream.Stream;
-import java.util.stream.StreamSupport;
-
 /**
  * This class turns the "wall of stacks" from the jdk.ThreadDump event into Stream<String>, where
```
mateuszrzeszutek, line 20: This comment is no longer true. How about renaming the class to `NextRegionFinder` or sth similar?

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpRegion.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler;
+
+public class ThreadDumpRegion {
+  public final String threadDump;
+  public int startIndex;
+  public int endIndex;
```
mateuszrzeszutek, line 22: Can we make these fields package private? The class and its methods could be package private too I think

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpRegion.java
```diff
@@ -0,0 +1,52 @@
...
+  public int indexOf(int character, int fromIndex) {
+    int result = threadDump.indexOf(character, fromIndex);
+    if (result >= endIndex) {
+      return -1;
+    }
+    return result;
+  }
+
+  @Override
+  public String toString() {
```
mateuszrzeszutek, line 49: While I agree that this is a valid `toString()` implementation, can we use a more explicit method name in `ThreadDumpProcessor`? Like `getCurrentRegion()` or something

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpToStacks.java
mateuszrzeszutek: Or, maybe the `findNext()` method could be an instance method of `ThreadDumpRegion`? There would be no need to expose its fields then

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/StackDescriptorLineParser.java
```diff
@@ -28,30 +30,31 @@
    * The first line is a meta/descriptor that has information about the following call stack. This
    * method parses out the thread id, which is the second field (space separated).
    */
-  long parseThreadId(String descriptorLine) {
+  long parseThreadId(ThreadDumpRegion stack) {
```
mateuszrzeszutek, line 33: Nit: another idea: `ThreadDumpRegion` could implement `CharSequence`, this class would need almost no changes then. And you could use simple strings to test it

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpRegion.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package com.splunk.opentelemetry.profiler;
+
+public class ThreadDumpRegion {
+  public final String threadDump;
+  public int startIndex;
+  public int endIndex;
+
+  public ThreadDumpRegion(String threadDump) {
```
mateuszrzeszutek, line 24: Is this just for tests? It's a pretty simple constructor, can we inline it and not keep it in the main sources?

.gitattributes
agoallikmaa: Can also remove from here and make a separate pull request to have proper discussion. Currently the issue is that without this, spotless will try to fix all LF settings to CRLF on Windows even if in git repository settings, `autocrlf` is not enabled.

agoallikmaa: Removed for now.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpToStacks.java
agoallikmaa: Irrelevant now as `ThreadDumpToStacks` migrated to a method within `ThreadDumpRegion`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpRegion.java
agoallikmaa: They are still used from the `context` package though.

profiler/src/main/java/com/splunk/opentelemetry/profiler/context/StackDescriptorLineParser.java
agoallikmaa: That would require managing a relative vs absolute offset within `ThreadDumpRegion` (externally only relative offset is used, internally needs to mapped to absolute, and vice versa). While it might simplify the logic here, it would add unnecessary much complexity to `ThreadDumpRegion` itself.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ThreadDumpRegion.java
agoallikmaa: Removed.

profiler/src/test/java/com/splunk/opentelemetry/profiler/ThreadDumpRegionTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Test
+  void testDoesNotFindAfterRange() {
+    assertEquals(-1, new ThreadDumpRegion("abcdef", 0, 4).indexOf('e', 2));
+  }
+
+  static String readDumpFromResource(String resourcePath) {
+    try {
+      InputStream in =
+          Thread.currentThread().getContextClassLoader().getResourceAsStream(resourcePath);
```
mateuszrzeszutek, line 84: Nit
```suggestion
    try (InputStream in =
          Thread.currentThread().getContextClassLoader().getResourceAsStream(resourcePath)) {
```

mateuszrzeszutek (approved):
LGTM 👍

profiler/src/test/java/com/splunk/opentelemetry/profiler/ThreadDumpRegionTest.java
agoallikmaa: Done.

mateuszrzeszutek (approved):

profiler/src/test/java/com/splunk/opentelemetry/profiler/context/StackDescriptorLineParserTest.java
```diff
@@ -28,79 +30,120 @@
...
+    StackDescriptorLineParser parser = new StackDescriptorLineParser();
+    ThreadDumpRegion region = new ThreadDumpRegion(descriptorLine, 0, descriptorLine.length());
+    long result = parser.parseThreadId(region);
+
+    Stream.of(DISRUPTIVE_AFFIXES)
+        .forEach(
+            it -> {
+              assertEqualsWithAffixes(result, parser, descriptorLine, "", it);
+              assertEqualsWithAffixes(result, parser, descriptorLine, it, "");
+            });
```
breedx-splk, line 132: This could be improved later maybe by leveraging junit dynamic tests?

breedx-splk (approved):
Had a few small comments, but looking good otherwise. Thanks.

theletterf (approved):

profiler/src/test/java/com/splunk/opentelemetry/profiler/context/StackDescriptorLineParserTest.java
agoallikmaa: Possibly, but not going to look into it for the current PR.

profiler/src/main/java/com/splunk/opentelemetry/profiler/StackTracePredicate.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler;
+
+public interface StackTracePredicate {
```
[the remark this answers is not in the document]
agoallikmaa: Added.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -173,11 +175,22 @@ private ThreadDumpProcessor buildThreadDumpProcessor(
...
+      Config config, SpanContextualizer contextualizer) {
     boolean includeAgentInternals = config.getBoolean(CONFIG_KEY_INCLUDE_AGENT_INTERNALS, false);
     boolean includeJVMInternals = config.getBoolean(CONFIG_KEY_INCLUDE_JVM_INTERNALS, false);
-    return new StackTraceFilter(includeAgentInternals, includeJVMInternals);
+    StackTraceFilter contentFilter =
+        new StackTraceFilter(includeAgentInternals, includeJVMInternals);
+
+    if (Configuration.getTracingStacksOnly(config)) {
+      return new TracingStacksOnlyFilter(contentFilter, contextualizer);
+    } else {
```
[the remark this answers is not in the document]
agoallikmaa: Outdated file, latest version doesn't have this part of the code.

PR #645, opened 2022-01-25
profiler: Unified option for enabling internal stacks, fix logic of enabling separately

agoallikmaa:
Since having to separately assign two options for enabling internal stack traces (agent internals vs JVM internals) can be confusing, there is now an option that toggles both: `splunk.profiler.include.internal.stacks`. The previous options still exist internally, but the new option is the default value for them if they are not explicitly specified. As setting them separately could only have very corner case uses, the documentation now only mentions the unified toggle.

Previously the option to enable agent internal stack traces also made JVM internal traces be reported. This was rather unintuitive, and therefore changed it so that JVM internal traces can still stay excluded even if agent internal traces are included.

Fixed tests using a different stack trace format (whitespace difference) than we really have, causing those to be never considered as JVM internal traces even if they would have otherwise qualified.

### DIFF
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -52,7 +52,7 @@ property or as environment variables (by making them uppercase and replacing dot
 |`splunk.profiler.memory.enabled`          | false                  | set to `true` to enable all other memory profiling options unless explicitly disabled |
 |`splunk.profiler.tlab.enabled`            | `splunk.profiler.memory.enabled` | set to `true` to enable TLAB events even if `splunk.profiler.memory.enabled` is `false` |
 |`splunk.profiler.memory.sampler.interval` | 1                      | set to `2` or larger to enable sampling every Nth allocation event where N is the value of this property |
-|`splunk.profiler.include.agent.internals` | false                  | set to `true` to include agent internal call stacks |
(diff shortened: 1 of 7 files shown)


profiler/README.md
```diff
@@ -52,7 +52,7 @@ property or as environment variables (by making them uppercase and replacing dot
 |`splunk.profiler.memory.enabled`          | false                  | set to `true` to enable all other memory profiling options unless explicitly disabled |
 |`splunk.profiler.tlab.enabled`            | `splunk.profiler.memory.enabled` | set to `true` to enable TLAB events even if `splunk.profiler.memory.enabled` is `false` |
 |`splunk.profiler.memory.sampler.interval` | 1                      | set to `2` or larger to enable sampling every Nth allocation event where N is the value of this property |
-|`splunk.profiler.include.agent.internals` | false                  | set to `true` to include agent internal call stacks |
+|`splunk.profiler.include.internal.stacks` | false                  | set to `true` to include stack traces of agent internal threads and stack traces with only JDK internal frames |
```
mateuszrzeszutek, line 55: Should this be added to the official docs?

mateuszrzeszutek (approved):

profiler/README.md
theletterf: Seems like it? Happy to create an update if you'd confirm.

laurit (approved):

PR #646, opened 2022-01-25
Tag docker images with the major/stable version

owais:
This is useful for customers and services like kubernetes operator to
target latest minor/patch version from a major series.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -23,6 +23,8 @@ fi
 release_tag="$1"
 # without the starting 'v'
 release_version=$(echo "$release_tag" | cut -c2-)
+# v1 from v1.2.3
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -23,6 +23,8 @@ fi
 release_tag="$1"
 # without the starting 'v'
 release_version=$(echo "$release_tag" | cut -c2-)
+# v1 from v1.2.3
```
mateuszrzeszutek, line 26: I don't think that v2 will appear any time soon 😅

owais: Hopefully never but I'm getting push back from others when trying to use `latest` 🤷

iNikem (approved):

PR #648, opened 2022-01-27
Remove host.name resource attribute

mateuszrzeszutek:
The `host.name` attribute was incorrectly detected in cloud environments (AWS, Azure, ...) so as a quick fix it was decided to just remove it.


laurit (approved):

iNikem (changes requested):
`host.name` (together with `host.id`) are important attributes for our internal usage telemetry. If `host.name` is correctly detected in non-cloud environments, then we should not just always remove it.

mateuszrzeszutek:
The problem is, how can we reliably detect a cloud environment?

mateuszrzeszutek:
Not needed after all

PR #652, opened 2022-01-31
v1.7.2 pre release changes

mateuszrzeszutek:
Patch release based on the upstream 1.10.1 patch

### DIFF
--- a/buildSrc/src/main/kotlin/splunk.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/splunk.java-conventions.gradle.kts
@@ -13,11 +13,11 @@ repositories {
   }
 }
 
-val otelVersion = "1.10.0"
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/splunk.java-conventions.gradle.kts
```diff
@@ -13,11 +13,11 @@ repositories {
   }
 }
 
-val otelVersion = "1.10.0"
-val otelAlphaVersion = "1.10.0-alpha"
+val otelVersion = "1.10.1"
```
laurit, line 16: there is also the hard to remember `implementation("io.opentelemetry.instrumentation:gradle-plugins:1.10.0-alpha")` in `buildSrc/build.gradle.kts`

laurit (approved):

PR #653, opened 2022-01-31
Fix bug related to iteration over PriorityQueue

breedx-splk:
PriorityQueue does not have a guarantee ordered iteration. This caused a bug that prevented the JFR events from being dispatched in time order. This has considerable consequences. 

The PriorityQueue was well intentioned as part comments in of #467 (commit d7fb40ac84b8a965bc746299ca155e916acf81dc), but ultimately introduced this problem.

Note: The test is only a loose guarantee. It is possible to reorder the events and still have the test pass, due to the inconsistency of iteration.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
@@ -16,14 +16,15 @@
 
 package com.splunk.opentelemetry.profiler;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.splunk.opentelemetry.profiler.context.SpanContextualizer;
 import com.splunk.opentelemetry.profiler.events.ContextAttached;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.PriorityQueue;
 import java.util.concurrent.TimeUnit;
 import jdk.jfr.EventType;
 import jdk.jfr.consumer.RecordedEvent;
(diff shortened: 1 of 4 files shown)


mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/EventProcessingChain.java
```diff
@@ -37,19 +38,27 @@
   private final SpanContextualizer spanContextualizer;
   private final ThreadDumpProcessor threadDumpProcessor;
   private final TLABProcessor tlabProcessor;
-  private final PriorityQueue<RecordedEvent> buffer =
-      new PriorityQueue<>(Comparator.comparing(RecordedEvent::getStartTime));
+  private final List<RecordedEvent> buffer = new LinkedList<>();
```
laurit, line 41: is there a reason why you are not using `ArrayList`?

breedx-splk: I don't think the size of the buffer per flush is well defined (or understood)...which is why I chose LinkedList instead. It's been my general understanding that LinkedList is a better choice when workloads aren't predictable/known. Because `buffer` is final we call `clear()` when flushing, which means that if we used ArrayList we would hold onto the underlying array and it would only grow, right?

Honestly, my intuition for this isn't super strong and if people feel that ArrayList would be a better choice I am happy to change it and we can microbenchmark it later.

laurit: See this tweet from the author of the LinkedList https://twitter.com/joshbloch/status/583813919019573248?lang=en
If you wish you can always remove final from `buffer` and recreate it on each iteration or use https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#trimToSize--.

breedx-splk: Ok, I'll move to ArrayList, if nothing because it's more commonplace.

PR #659, opened 2022-02-07
Remove overlap factor

laurit:
As far as I understand recordings don't really overlap.


mateuszrzeszutek (approved):
Is it described anywhere in the JFR docs?

laurit:
> Is it described anywhere in the JFR docs?

I suspect that this overlap was more of our implementation detail than JFR. Only @breedx-splk knows what it really is. I just have a suspicion that whatever purpose it might have served isn't really there any more.

breedx-splk (approved):
Right, so I believe originally that the new were started and stopped for every period. With that approach, you almost certainly need an overlap in order to not miss events, which is why there was overlap.

The approach was changed, and we now have a long-running recording from which we periodically snapshot the data. As long as the max recording duration is considerably longer than the snapshot period, it should be safe to remove this overlap.

The recording max age is currently set to 10 x the snapshot period (which is still spelled `recordingDuration` in a few places, adding confusion)...so I think this should be safe.

PR #674, opened 2022-02-16
Wait for weblogic to start

laurit:
Resolves https://github.com/signalfx/splunk-otel-java/issues/673
Resolves https://github.com/signalfx/splunk-otel-java/issues/668
Resolves https://github.com/signalfx/splunk-otel-java/issues/667
Weblogic metrics are registered fairly late in startup process, waiting until running message is seen should ensure that they are registered before our tests start.

### DIFF
--- a/instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/instrumentation/weblogic/metrics/ServerLifecycleListenerInstrumentation.java
+++ b/instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/instrumentation/weblogic/metrics/ServerLifecycleListenerInstrumentation.java
@@ -39,7 +39,7 @@ public class ServerLifecycleListenerInstrumentation implements TypeInstrumentati
 
   @SuppressWarnings("unused")
   public static class InitializeAdvice {
-    @Advice.OnMethodExit(suppress = Throwable.class)
+    @Advice.OnMethodEnter(suppress = Throwable.class)
     public static void onEnter() {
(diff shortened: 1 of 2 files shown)


instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/instrumentation/weblogic/metrics/ServerLifecycleListenerInstrumentation.java
```diff
@@ -39,7 +39,7 @@ public void transform(TypeTransformer transformer) {
 
   @SuppressWarnings("unused")
   public static class InitializeAdvice {
-    @Advice.OnMethodExit(suppress = Throwable.class)
+    @Advice.OnMethodEnter(suppress = Throwable.class)
```
breedx-splk, line 42: Was this change to also get the process going earlier in the lifecycle?

breedx-splk (commented):
Looks good, but it wasn't clear that the implementation change was intentional.

instrumentation/weblogic/src/main/java/com/splunk/opentelemetry/instrumentation/weblogic/metrics/ServerLifecycleListenerInstrumentation.java
laurit: Yes. The startup message should be printed by something called from this method.

iNikem (approved):

PR #684, opened 2022-02-23
Prototype sending allocation stacks as pprof

laurit:
@gsmirnov-splk @vovencij please review


profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
```diff
@@ -0,0 +1,135 @@
...
+    String stack = stackSerializer.serialize(stackTrace);
+    RecordedThread thread = event.getThread();
+    String name = thread == null ? "unknown" : thread.getJavaName();
+    long id = thread == null ? 0 : thread.getJavaThreadId();
+    return "\""
+        + name
+        + "\""
+        + " #"
+        + id
+        + " prio=0 os_prio=0 cpu=0ms elapsed=0s tid=0 nid=0 unknown"
```
gsmirnov-splk, line 95: While we're here, can we change `nid=0` to `nid=0x0`? The thread id parser on the backend kinda expects it to be a hex number and spams the logs with warnings.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+      spanIdLabel.setKey(stringTable.get("span_id"));
+      spanIdLabel.setStr(stringTable.get(spanContext.getSpanId()));
+      sample.addLabel(spanIdLabel.build());
+
+      Label.Builder eventTypeLabel = Label.newBuilder();
+      eventTypeLabel.setKey(stringTable.get("event_type"));
+      eventTypeLabel.setStr(stringTable.get(event.getEventType().getName()));
+      sample.addLabel(eventTypeLabel.build());
+    }
+    // XXX are there any more attributes that should be added to the sample?
```
gsmirnov-splk, line 110: On the backend, the stack traces are expected to have the following attributes on top of this:

* `service.name` (as set via `otel.service.name`)
* `deployment.environment` (as set via `otel.resource.attributes -> environment`)
* `thread.name`, `thread.id`, `thread.state` (when known)
* `source.event.name` (e.g. "jdk.ObjectAllocationInNewTLAB" or "jdk.ObjectAllocationOutsideTLAB")
* `telemetry.sdk.language`
* `container.id`
* `host.name`

Now, some of these attributes like service name, env, host name, etc, are not specific to a given stack trace, so can be set on the batch level. But we do need to receive them.

@vovencij please review these too.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+
+    // XXX just to give an overview of exported data size
+    logger.info("Exporting {}, size {}", encoding, bytes.length);
+
+    Attributes attributes =
+        Attributes.builder()
+            .put(SOURCE_TYPE, PROFILING_SOURCE)
+            .put(DATA_TYPE, "allocation") // tells that this message is about allocation samples
+            .put(ENCODING, encoding) // data encoding
+            .build();
```
gsmirnov-splk, line 156: Yeah, I guess the attributes like service name, env, language etc go here.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+
+    String body = new String(bytes, StandardCharsets.ISO_8859_1);
```
gsmirnov-splk, line 158: Hmm, is this charset a part of some spec? Also, this is a lot of allocation, I wonder if we can somehow avoid this, like does the OTLP format mandate it as a string?

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+    final Profile.Builder profile;
+    final Map<String, Long> table = new HashMap<>();
+    long index = 0;
+
+    StringTable(Profile.Builder profile) {
+      this.profile = profile;
+      get(""); // 0 is reserved for the empty string
+    }
+
+    long get(String str) {
```
gsmirnov-splk, line 225: Likewise, will this `get()` method ever only be invoked from the same thread on a given instance of `StringTable`?

profiler/src/main/proto/profile.proto
```diff
@@ -0,0 +1,212 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
gsmirnov-splk, line 1: Should a link to the original pprof proto file(s) be included?

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/AllocationEventExporter.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.splunk.opentelemetry.profiler.allocation.exporter;
+
+import com.splunk.opentelemetry.profiler.allocation.sampler.AllocationEventSampler;
+import io.opentelemetry.api.trace.SpanContext;
+import jdk.jfr.consumer.RecordedEvent;
+
+public interface AllocationEventExporter {
+
+  void export(RecordedEvent event, AllocationEventSampler sampler, SpanContext spanContext);
```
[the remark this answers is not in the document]
laurit: yes, we already should have

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
laurit: It would be better to remove it completely rather than use bogus values. https://github.com/signalfx/gdi-specification/commit/b09e176ca3771c3ef19fc9d23e8722fc77a3b6e9 allows it.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+    RecordedStackTrace stackTrace = event.getStackTrace();
+    if (stackTrace == null) {
+      return;
+    }
+
+    Profile.Builder profile = profileData.profile;
+    LocationTable locationTable = profileData.locationTable;
+    StringTable stringTable = profileData.stringTable;
+
+    // XXX jdk mission control "TLAB allocations" view seems to use tlabSize as allocation size
```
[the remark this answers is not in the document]
laurit: Agent use `allocationSize`. I added this comment just as a note that there is also `tlabSize` in case you wan't to estimate allocation rate where `tlabSize` would give a much better result. Though sampling these events would still make it useless.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+    long allocationSize = event.getLong("allocationSize");
+
+    Sample.Builder sample = Sample.newBuilder();
+    sample.addValue(allocationSize);
+    // XXX StackSerializer limits stack depth. Although I believe jfr also limits it should we
```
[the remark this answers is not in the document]
laurit: `StackSerializer` uses `128` as limit. Let's do it as a separate task, please add it to jira.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -116,11 +121,28 @@ private void activateJfrAndRunForever(Config config, Resource resource) {
...
         buildThreadDumpProcessor(
             spanContextualizer, processor, buildStackTraceFilter(config), config);
+
+    AllocationDataFormat allocationDataFormat = Configuration.getAllocationDataFormat(config);
+    AllocationEventExporter allocationEventExporter;
+    if (allocationDataFormat == AllocationDataFormat.TEXT) {
+      allocationEventExporter =
+          PlainTextAllocationEventExporter.builder()
+              .logProcessor(batchingLogsProcessor)
+              .commonAttributes(commonAttributes)
```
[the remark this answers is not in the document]
laurit: `commonAttributes` has attributes that describe a singe event such as event name and event period which make it unusable when a single log entry represents a batch of events. These attributes have to be added as labels or values to pprof data if they are needed at all.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
laurit: You need to differentiate attributes that are needed once for the whole batch and the ones that are needed for every stack. Here we are dealing with attributes that are needed for every stack.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
laurit: There is a bunch of attributes that are added by default. Here we need to agree which attribute is used to tell that data contains allocation info vs cpu stack and how to specify whether the data is plain text or pprof.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+    } catch (IOException exception) {
+      throw new IllegalStateException("Failed to serialize pprof", exception);
+    }
+  }
+
+  public static final AttributeKey<String> DATA_TYPE = stringKey("com.splunk.datatype");
+  public static final AttributeKey<String> ENCODING = stringKey("com.splunk.encoding");
+
+  @Override
+  public void flush() {
```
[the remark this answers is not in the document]
laurit: yes.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
laurit: Currently api only allows adding string as body.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
laurit: yes

profiler/src/main/proto/profile.proto
laurit: If this ends up used then it will need to go through gdi spec where it will be documented.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
vsor-splk: Yes, just remove them.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
vsor-splk: Afaiu, service name, env and language will be regular common otel resource attributes, so no need to add them into pprof.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
vsor-splk: ... although thread name, state and ID are per-stack.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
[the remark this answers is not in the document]
gsmirnov-splk: Yeah. We get a very good number for allocation rate from the `jvm.experimental.memory.allocated` metric, and these tlab events represent like 2% of all allocations in my experiments. So I think we can skip this.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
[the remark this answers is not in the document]
gsmirnov-splk: Hmm, on the backend max stack trace depth in the last 7d is 169 frames. Maybe some deployments are running an old agent without this limit? :thonk: Anyways, OK, will track this separately.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
gsmirnov-splk: Yeah, the per-stack attributes are:
* event source/type
* thread name/id/state
* trace/span ids

So looks like only the thread info is missing. The plain text data format does include the name, so let's include that too, even if state and id are unavailable.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
gsmirnov-splk: So it's `com.splunk.datatype` and `com.splunk.encoding`. Is it common in custom otlp data otlp to prefix the custom attributes by `com.splunk.`? As for the names, current ones are alright with me, perhaps `datatype -> profilerType` and `encoding -> format` would be a bit less ambiguous. Encoding makes me immediately think of UTF-8 :)

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
gsmirnov-splk: Aha, got it, a possible hackaround. `ISO_8859_1` still has a bunch of undefined values, and the data might contain eight 0 bits in a row, I would expect something, somewhere in the pipeline, to break because of it. Will try

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+          || allocationDataFormat == AllocationDataFormat.PPROF_GZIP_BASE64) {
+        bytes = Base64.getEncoder().encode(bytes);
+      }
+      return bytes;
+    } catch (IOException exception) {
+      throw new IllegalStateException("Failed to serialize pprof", exception);
+    }
+  }
+
+  public static final AttributeKey<String> DATA_TYPE = stringKey("com.splunk.datatype");
```
vsor-splk, line 139: I suggest we be a bit more specific here and let's add `profiling` in between, like `com.splunk.profiler.datatype`. Same for encoding.

vsor-splk, line 139: Also it would be nice to include this attribute for plain text.

laurit: changed to `com.splunk.profiling.data.type` and `com.splunk.profiling.data.format`

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
laurit: Change to `com.splunk.profiling.data.type` and `com.splunk.profiling.data.format`. The `com.splunk.` prefix was already used for existing keys so just followed the same style. I guess it isn't really needed.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,317 @@
...
+    addLabel(sample, SOURCE_EVENT_NAME.getKey(), event.getEventType().getName());
+    Duration eventPeriod = eventPeriods.getDuration(eventName);
+    if (!EventPeriods.UNKNOWN.equals(eventPeriod)) {
+      addLabel(sample, SOURCE_EVENT_PERIOD.getKey(), String.valueOf(eventPeriod.toMillis()));
+    }
+    Instant time = event.getStartTime();
+    addLabel(sample, "source.event.time", String.valueOf(time.toEpochMilli()));
+
+    RecordedThread thread = event.getThread();
+    addLabel(sample, "thread.id", String.valueOf(thread.getJavaThreadId()));
```
vsor-splk, line 111: On the backend we index native thread Id or `nid` from the thread dump metadata line, I presume it is from `thread.getOSThreadId()`.

gsmirnov-splk: You can just add `nid=0x...`

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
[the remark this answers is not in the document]
gsmirnov-splk: This thread can be resolved (there is a separate Jira ticket). But I do not see a "resolve" button, even though I started the thread :shrug:

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,291 @@
...
+      throw new IllegalStateException("Failed to serialize pprof", exception);
+    }
+  }
+
+  public static final AttributeKey<String> DATA_TYPE = stringKey("com.splunk.datatype");
+  public static final AttributeKey<String> ENCODING = stringKey("com.splunk.encoding");
+
+  @Override
+  public void flush() {
+    // Flush is called after each JFR chunk, hopefully this will keep batch sizes small enough.
```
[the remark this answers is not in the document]
gsmirnov-splk: Resolve

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
gsmirnov-splk: Resolve

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
gsmirnov-splk: So apparently this does not work -- we have to do base64. Otherwise bytes like `0x00, 0x07, 0x0F, 0x10, 0x17, 0x1F, 0x7F, 0x80, 0x87, 0x8F, 0x90, 0x97, 0x9F, 0xD7, 0xF7` will be corrupted on the way, both collector and ingest will change them during JSON conversion.

Resolve.

profiler/src/main/java/com/splunk/opentelemetry/profiler/exporter/PprofProfilingEventExporter.java
```diff
@@ -0,0 +1,253 @@
...
+  public void export(StackToSpanLinkage stackToSpanLinkage) {
+    StackTrace st = handleStackTrace(stackToSpanLinkage.getRawStack());
+    if (st.stackTraceLines.isEmpty()) {
+      return;
+    }
+
+    Sample.Builder sample = Sample.newBuilder();
+    {
+      Label.Builder label = Label.newBuilder();
+      // XXX should we attempt to extract relevant attributes here?
```
gsmirnov-splk, line 73: It would be nice to split it up into `thread.name`, `thread.id` and `thread.status`, like we have for allocations (well, except for the status), for the sake of consistency.

profiler/src/main/java/com/splunk/opentelemetry/profiler/exporter/PprofProfilingEventExporter.java
```diff
@@ -0,0 +1,253 @@
...
+    SpanContext spanContext = stackToSpanLinkage.getSpanContext();
+    if (spanContext != null && spanContext.isValid()) {
+      pprof.addLabel(sample, "trace_id", spanContext.getTraceId());
+      pprof.addLabel(sample, "span_id", spanContext.getSpanId());
+    }
+
+    pprof.getProfileBuilder().addSample(sample);
+  }
+
+  private static StackTrace handleStackTrace(String stackTrace) {
```
gsmirnov-splk, line 107: Hey @vovencij please 👀

profiler/src/main/java/com/splunk/opentelemetry/profiler/exporter/PprofProfilingEventExporter.java
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  @Override
+  public void flush() {
+    // Flush is called after each JFR chunk, hopefully this will keep batch sizes small enough.
+    byte[] bytes = serializePprof();
+    String format = profilingDataFormat.toString().toLowerCase(Locale.ROOT).replace('_', '-');
+
+    // XXX just to give an overview of exported data size
+    logger.info("Exporting {}, size {}", format, bytes.length);
```
gsmirnov-splk, line 200: I think we can now turn it to `debug`?

profiler/src/main/java/com/splunk/opentelemetry/profiler/exporter/PprofProfilingEventExporter.java
```diff
@@ -0,0 +1,253 @@
...
+
+    Attributes attributes =
+        Attributes.builder()
+            .put(SOURCE_TYPE, PROFILING_SOURCE)
+            .put(DATA_TYPE, "profiling") // tells that this message is about allocation samples
```
gsmirnov-splk, line 205: > `allocation samples`

copy-paste issue? These are not allocation samples.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -0,0 +1,167 @@
...
+    pprof.addLabel(sample, SOURCE_EVENT_NAME, eventName);
+    Instant time = event.getStartTime();
+    pprof.addLabel(sample, SOURCE_EVENT_TIME, time.toEpochMilli());
+
+    RecordedThread thread = event.getThread();
+    if (thread.getJavaThreadId() != -1) {
+      pprof.addLabel(sample, THREAD_ID, thread.getJavaThreadId());
+      pprof.addLabel(sample, THREAD_NAME, thread.getJavaName());
+    }
+    pprof.addLabel(sample, THREAD_NATIVE_ID, thread.getOSThreadId());
```
gsmirnov-splk, line 100: What about explicitly adding `pprof.addLabel(sample, THREAD_STATUS, "RUNNABLE");` here, so that we don't have to add special handling on the backend?

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingDataType.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler;
+
+import java.util.Locale;
+
+public enum ProfilingDataType {
+  ALLOCATION,
+  PROFILING;
```
vsor-splk, line 23: `PROFILING` is not a nice name. Allocation is also profiling. Let's use `CPU` or `SAMPLER` instead.

vsor-splk: Backend is using `CPU` for this.

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -105,4 +108,33 @@ public static boolean getIncludeJvmInternalStacks(Config config) {
...
+
+  private static DataFormat getDataFormat(String value) {
+    return DataFormat.valueOf(value.toUpperCase(Locale.ROOT).replace('-', '_'));
+  }
+
+  public enum DataFormat {
+    TEXT,
+    PPROF_GZIP_BASE64;
+
+    private String value;
```
vsor-splk, line 130: `final`?

vsor-splk (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -105,4 +108,33 @@ public static boolean getIncludeJvmInternalStacks(Config config) {
...
     return config.getBoolean(CONFIG_KEY_TRACING_STACKS_ONLY, DEFAULT_TRACING_STACKS_ONLY);
   }
+
+  public static DataFormat getCpuDataFormat(Config config) {
+    String value = config.getString(CONFIG_KEY_CPU_DATA_FORMAT, "text");
+    return getDataFormat(value);
+  }
+
+  public static DataFormat getAllocationDataFormat(Config config) {
+    String value = config.getString(CONFIG_KEY_MEMORY_DATA_FORMAT, "pprof-gzip-base64");
```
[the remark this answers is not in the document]
gsmirnov-splk: To switch CPU profiling to pprof by default, we need to enable this support on the backend in all realms. This will take some time. No such limitation for memory profiling data. We will make a separate request to switch the default as soon as we're confident that the backend is ready.

profiler/src/main/java/com/splunk/opentelemetry/profiler/exporter/PprofCpuEventExporter.java
```diff
@@ -0,0 +1,149 @@
...
+  private final EventPeriods eventPeriods;
+  private final PprofLogDataExporter pprofLogDataExporter;
+  private Pprof pprof = createPprof();
+
+  private PprofCpuEventExporter(Builder builder) {
+    this.dataFormat = builder.dataFormat;
+    this.eventPeriods = builder.eventPeriods;
+    this.pprofLogDataExporter =
+        new PprofLogDataExporter(
+            builder.logProcessor, builder.resource, ProfilingDataType.CPU, builder.dataFormat);
```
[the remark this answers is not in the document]
laurit: This is because initially we had more formats. I'd keep it in case we need to experiment with different format, it would be nice if we could get rid of the base64.

profiler/src/test/java/com/splunk/opentelemetry/profiler/exporter/StackTraceParserTest.java
```diff
@@ -0,0 +1,74 @@
...
+import com.splunk.opentelemetry.profiler.ThreadDumpRegion;
+import com.splunk.opentelemetry.profiler.exporter.StackTraceParser.StackTrace;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class StackTraceParserTest {
```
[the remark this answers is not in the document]
laurit: I added assertions for stack trace lines. Test data contains multiple stack traces, test verifies actual values for only one stack trace, for the rest it just tests for not null.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
```diff
@@ -0,0 +1,137 @@
...
+    RecordedThread thread = event.getThread();
+    String name = thread == null ? "unknown" : thread.getJavaName();
+    long id = thread == null ? 0 : thread.getJavaThreadId();
+    String result = "\"" + name + "\"" + " #" + id;
+    if (thread != null) {
+      result += " nid=0x" + Long.toHexString(thread.getOSThreadId());
+    }
+    result += "\n";
+    result += "   java.lang.Thread.State: UNKNOWN\n" + stack;
+    return result;
```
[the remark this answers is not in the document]
laurit: I would leave it as it is because `StringBuilder` reduces the readability. Also the plan is to use `pprof` exporter by default so this class really won't be used by anything else besides our tests if all goes as planned. Perhaps we can even delete it at some point.

profiler/src/test/java/com/splunk/opentelemetry/profiler/exporter/StackTraceParserTest.java
```diff
@@ -0,0 +1,91 @@
...
+    ThreadDumpRegion stack = new ThreadDumpRegion(wallOfStacks, 0, 0);
+
+    boolean found = false;
+    while (stack.findNextStack()) {
+      StackTrace stackTrace = StackTraceParser.parse(stack.getCurrentRegion());
+      if (stackTrace == null) {
+        continue;
+      }
+
+      Assertions.assertNotNull(stackTrace.getThreadName());
```
mateuszrzeszutek, line 43: Nit: static import

mateuszrzeszutek (approved):

PR #686, opened 2022-02-28
Add micrometer heap pressure metrics

laurit:
@gsmirnov-splk please review
Should we always enable this or only when memory profiling is enabled?


gsmirnov-splk:
LGTM. I think, it makes sense to enable it when other metrics like `JvmMemoryMetrics` or `JvmGcMetrics` are enabled.

breedx-splk (approved):

mateuszrzeszutek (approved):

mateuszrzeszutek:
Can you add an assertion for one of these new meters to `JvmMetricsTest`?

mateuszrzeszutek (approved):

PR #687, opened 2022-03-02
Remove bogus info from allocation stacks

laurit:
Shouldn't be needed since https://github.com/signalfx/gdi-specification/commit/b09e176ca3771c3ef19fc9d23e8722fc77a3b6e9


breedx-splk (approved):
Cool, I confirmed with the backend folks that it should be ok as well. 👍🏻

PR #694, opened 2022-03-07
Fix servlet instrumentation using internal spans for Server-Timings header

agoallikmaa:
Servlet instrumentation currently overwrites `Server-Timings` header upon entering every filter/servlet invocation. This causes it to be set to an internal span if one of those filters switched span context to an internal span that is a child of the server span. Added a call depth filter to the instrumentation to only do it for the outermost filter/servlet that has a valid span context, and a test to verify it (made sure test fails without the fix).

Had to update OpenTelemetry SDK versions used, since the latest instrumentation version was just switched to use new SDK version as well (exploded when trying to find recently added histogram classes otherwise).


mateuszrzeszutek (approved):
@agoallikmaa can you rebase with the `main` branch? The SDK version has already been bumped there

agoallikmaa:
> @agoallikmaa can you rebase with the `main` branch? The SDK version has already been bumped there

Done.

agoallikmaa:
`RequestDispatcher#forward` doesn't seem to actually reset headers, only the body. As for `ServletResponse#reset`, ideally we should indeed handle this - it seems to be used for internal 500 errors in some appservers (the built-in kind, which wouldn't have a RUM agent on it anyway) - but perhaps in the epic dedicated to `Server-Timings`, as this fix should cover most common cases.

As for how to do that, that may be a bit trickier. I was thinking of having a virtual field (`StickyHeaders`) attached to response object which basically just contains a map to which we add headers together with calling `setHeader`, so after reset we could just reapply all headers from that map. Would also need to make sure we clear this map then.

PR #739, opened 2022-04-18
Add an option to limit stack depth

laurit:
@gsmirnov-splk @vovencij please review

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
@@ -61,7 +61,7 @@ public class Configuration implements ConfigPropertySource {
   public static final String CONFIG_KEY_INCLUDE_INTERNAL_STACKS =
       "splunk.profiler.include.internal.stacks";
   public static final String CONFIG_KEY_TRACING_STACKS_ONLY = "splunk.profiler.tracing.stacks.only";
-  private static final String CONFIG_KEY_STACK_DEPTH = "splunk.profiler.stack.depth";
+  private static final String CONFIG_KEY_STACK_DEPTH = "splunk.profiler.stack.max.depth";
 
   @Override
   public Map<String, String> getProperties() {
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
@@ -61,7 +61,7 @@ public class Configuration implements ConfigPropertySource {
   public static final String CONFIG_KEY_INCLUDE_INTERNAL_STACKS =
       "splunk.profiler.include.internal.stacks";
   public static final String CONFIG_KEY_TRACING_STACKS_ONLY = "splunk.profiler.tracing.stacks.only";
-  private static final String CONFIG_KEY_STACK_DEPTH = "splunk.profiler.stack.max.depth";
+  private static final String CONFIG_KEY_STACK_DEPTH = "splunk.profiler.max.stack.depth";
 
   @Override
   public Map<String, String> getProperties() {
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
@@ -36,6 +36,7 @@ public class Configuration implements ConfigPropertySource {
   public static final Duration DEFAULT_CALL_STACK_INTERVAL = Duration.ofSeconds(10);
   public static final boolean DEFAULT_INCLUDE_INTERNAL_STACKS = false;
   public static final boolean DEFAULT_TRACING_STACKS_ONLY = false;
+  private static final int DEFAULT_STACK_DEPTH = 1024;
 
   public static final String CONFIG_KEY_ENABLE_PROFILER = PROFILER_ENABLED_PROPERTY;
   public static final String CONFIG_KEY_PROFILER_DIRECTORY = "splunk.profiler.directory";
@@ -46,8 +47,8 @@ public class Configuration implements ConfigPropertySource {
   public static final String CONFIG_KEY_MEMORY_ENABLED = PROFILER_MEMORY_ENABLED_PROPERTY;
   public static final String CONFIG_KEY_MEMORY_SAMPLER_INTERVAL =
       "splunk.profiler.memory.sampler.interval";
-  public static final String CONFIG_KEY_CPU_DATA_FORMAT = "splunk.profiler.cpu.data.format";
-  public static final String CONFIG_KEY_MEMORY_DATA_FORMAT = "splunk.profiler.memory.data.format";
+  private static final String CONFIG_KEY_CPU_DATA_FORMAT = "splunk.profiler.cpu.data.format";
+  private static final String CONFIG_KEY_MEMORY_DATA_FORMAT = "splunk.profiler.memory.data.format";
   public static final String CONFIG_KEY_TLAB_ENABLED = "splunk.profiler.tlab.enabled";
   public static final String CONFIG_KEY_CALL_STACK_INTERVAL = "splunk.profiler.call.stack.interval";
   public static final String CONFIG_KEY_DEPRECATED_THREADDUMP_PERIOD =
@@ -60,6 +61,7 @@ public class Configuration implements ConfigPropertySource {
   public static final String CONFIG_KEY_INCLUDE_INTERNAL_STACKS =
       "splunk.profiler.include.internal.stacks";
   public static final String CONFIG_KEY_TRACING_STACKS_ONLY = "splunk.profiler.tracing.stacks.only";
+  private static final String CONFIG_KEY_STACK_DEPTH = "splunk.profiler.stack.depth";
 
   @Override
   public Map<String, String> getProperties() {
@@ -107,6 +109,10 @@ public class Configuration implements ConfigPropertySource {
 
   public static boolean getTracingStacksOnly(Config config) {
     return config.getBoolean(CONFIG_KEY_TRACING_STACKS_ONLY, DEFAULT_TRACING_STACKS_ONLY);
+  }
+
+  public static int getStackDepth(Config config) {
+    return config.getInt(CONFIG_KEY_STACK_DEPTH, DEFAULT_STACK_DEPTH);
   }
 
   public static DataFormat getCpuDataFormat(Config config) {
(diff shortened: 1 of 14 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/JFR.java
```diff
@@ -61,4 +70,40 @@ public RecordedEvent readEvent(RecordingFile file, Path path) {
       throw new JfrException("Error reading events from " + path, e);
     }
   }
+
+  public void setStackDepth(int stackDepth) {
+    try {
+      JfrAccess.init();
+      int currentDepth = Options.getStackDepth();
+      // we only increase stack depth
+      if (currentDepth < stackDepth) {
```
laurit, line 79: If user has set a large stack depth from command line we won't change it to a smaller one. I don't know whether this is necessary.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
```diff
@@ -71,6 +74,10 @@ public void export(RecordedEvent event, AllocationEventSampler sampler, SpanCont
     if (sampler != null) {
       sampler.addAttributes((k, v) -> builder.put(k, v), (k, v) -> builder.put(k, v));
     }
+    if (stackTrace.isTruncated() || stackTrace.getFrames().size() > stackDepth) {
+      builder.put(THREAD_STACK_TRUNCATED, true);
```
gsmirnov-splk, line 78: Maybe add a comment explaining that the actual truncation happens in `StackTraceParser`? For a bit I was confused that it's missing.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -69,21 +71,27 @@ public void export(RecordedEvent event, AllocationEventSampler sampler, SpanCont
...
+              if (method == null) {
+                sample.addLocationId(pprof.getLocationId("unknown", "unknown.unknown", -1));
+              } else {
+                sample.addLocationId(
+                    pprof.getLocationId(
+                        "unknown", // file name is not known
+                        method.getType().getName() + "." + method.getName(),
+                        frame.getLineNumber()));
+              }
+            });
```
gsmirnov-splk, line 93: This logic is present for pprof allocation exporter, but not for pprof cpu exporter (or any text exporters). Maybe add a bunch of comments to the exporters that rely on other classes to do the truncation? Or somehow unify it.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
laurit: usually truncation is already done by jfr, truncation in `StackTraceParser` is only used when jfr stack depth is greater than ours

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
laurit: for pprof cpu truncation is done in `StackTraceParser`, for text allocation in `StackSerializer`. The only once that does not truncate is text cpu where truncating would require parsing & rewriting the stacks

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -60,6 +61,7 @@ public class Configuration implements ConfigPropertySource {
   public static final String CONFIG_KEY_INCLUDE_INTERNAL_STACKS =
       "splunk.profiler.include.internal.stacks";
   public static final String CONFIG_KEY_TRACING_STACKS_ONLY = "splunk.profiler.tracing.stacks.only";
+  private static final String CONFIG_KEY_STACK_DEPTH = "splunk.profiler.stack.depth";
```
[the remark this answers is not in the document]
laurit: renamed

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
gsmirnov-splk: Right, I understand that. What I mean is that it was very easy to find this logic for "hey, if JFR gives us a stack trace taller than we want, truncate it" in case of pprof allocations, but for other type/format it is less obvious.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
gsmirnov-splk: I see. In case of text+cpu, do we expect that JFR will respect the limit that was passed along to it? (except when the user overrides it)?

laurit: no, I would not expect the stack depth limit to apply to thread dump events.

gsmirnov-splk: Alright, then are the following correct (as of this current PR)?

1. **text/cpu**: the truncation option is ignored
2. **text/allocation**: we try to ask JFR to truncate and also truncate on our own if it's still too tall
3. **pprof/cpu**: we truncate on our own
4. **pprof/allocation**: we try to ask JFR to truncate and also truncate on our own if it's still too tall

laurit: yes

profiler/src/main/java/com/splunk/opentelemetry/profiler/JFR.java
```diff
@@ -61,4 +70,40 @@ public RecordedEvent readEvent(RecordingFile file, Path path) {
...
+
+  private static class JfrAccess {
+    static {
+      Instrumentation instrumentation = InstrumentationHolder.getInstrumentation();
+      if (instrumentation != null && JavaModule.isSupported()) {
+        // ensure that we have access to jdk.jfr.internal.Options.setStackDepth
+        JavaModule currentModule = JavaModule.ofType(JFR.class);
+        JavaModule flightRecorder = JavaModule.ofType(FlightRecorder.class);
+        if (flightRecorder != null && flightRecorder.isNamed() && currentModule != null) {
+          ClassInjector.UsingInstrumentation.redefineModule(
```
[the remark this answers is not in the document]
gsmirnov-splk: Please bear with my ignorance, as last time I did instrumentation was in 2016.

So on Java 8, the `JavaModule.isSupported()` check above would return false, so nothing will be done. Do we need no access hacks on java 8?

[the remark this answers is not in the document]
laurit: It should work on java8 without extra tricks.

breedx-splk (changes requested):
A couple small things.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
```diff
@@ -69,21 +71,27 @@ public void export(RecordedEvent event, AllocationEventSampler sampler, SpanCont
...
-                method.getType().getName() + "." + method.getName(),
-                frame.getLineNumber()));
-      }
+
+    if (stackTrace.isTruncated() || stackTrace.getFrames().size() > stackDepth) {
+      pprof.addLabel(sample, THREAD_STACK_TRUNCATED, true);
     }
 
+    stackTrace.getFrames().stream()
+        .limit(stackDepth)
```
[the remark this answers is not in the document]
laurit: comment added, it is top down so the root frame is the first to get removed

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -60,6 +61,7 @@ public class Configuration implements ConfigPropertySource {
   public static final String CONFIG_KEY_INCLUDE_INTERNAL_STACKS =
       "splunk.profiler.include.internal.stacks";
   public static final String CONFIG_KEY_TRACING_STACKS_ONLY = "splunk.profiler.tracing.stacks.only";
+  private static final String CONFIG_KEY_STACK_DEPTH = "splunk.profiler.stack.max.depth";
```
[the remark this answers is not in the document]
laurit: changed to `splunk.profiler.max.stack.depth`

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PprofAllocationEventExporter.java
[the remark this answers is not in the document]
breedx-splk: K cool, thanks. I like this because it keeps the more important context (I think).

breedx-splk (approved):
👍🏻

profiler/src/main/java/com/splunk/opentelemetry/profiler/JFR.java
gsmirnov-splk: This thread can be resolved

profiler/src/main/java/com/splunk/opentelemetry/profiler/JFR.java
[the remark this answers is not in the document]
gsmirnov-splk: This thread can be resolved

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
gsmirnov-splk: This thread can be resolved

profiler/src/main/java/com/splunk/opentelemetry/profiler/JFR.java
```diff
@@ -61,4 +70,40 @@ public RecordedEvent readEvent(RecordingFile file, Path path) {
       throw new JfrException("Error reading events from " + path, e);
     }
   }
+
+  public void setStackDepth(int stackDepth) {
+    try {
+      JfrAccess.init();
```
vsor-splk, line 76: What will happen if `setStackDepth` is called multiple times (for whatever reason, but it's a public method so you never know)?

mateuszrzeszutek (approved):

PR #755, opened 2022-04-27
Run overhead tests automatically based on the commit message

mateuszrzeszutek:
The only sensible way of automating this seems to be through commit messages (it's surprisingly hard to create a tag on main once a given PR is merged), so I added a `[agent-version-update]` "tag" that'll trigger the overhead test run.

Related to https://github.com/signalfx/splunk-otel-java-overhead-test/pull/45


breedx-splk (approved):
This seems like a perfectly reasonable bodge. 😁

Issue #754, opened 2022-04-27
Enhance support for running Profiling with multiple processes on the same machine


gsmirnov-splk:
There are a few use cases where running Profiling in multiple JVMs on the same host (or in the same container) is required. For example, Apache Druid has a node type called "Middle Manager" that spawns worker JVMs called "Peons". There is no possibility to pass different arguments to different peons, so setting a dedicated folder for `splunk.profiler.directory` for each peon is not an option. Therefore, this can be observed in the agent logs:

```
ERROR com.splunk.opentelemetry.profiler.JfrRecorder - Error flushing JFR snapshot data to disk
java.nio.file.FileAlreadyExistsException: ./otel-profiler-2022-04-27T12_19_55.jfr
```

So, two feature requests here:

1. Make the jfr file names more unique to avoid clashes. Perhaps include the PID or some random identifier in there.
2. Log the absolute path to the JFR files, eg `/opt/druid/data/otel-profiler-xxx.jfr` instead of `./otel-profiler-xxx.jfr`

PR #763, opened 2022-05-02
Fix jfr snapshots using the same file

laurit:
Resolves https://github.com/signalfx/splunk-otel-java/issues/754


profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingFileNamingConvention.java
```diff
@@ -28,30 +29,26 @@ class RecordingFileNamingConvention {
...
-
-  private Path buildRecordingName(LocalDateTime dateTime) {
+  Path newOutputPath(LocalDateTime dateTime) throws IOException {
     String timestamp =
         DateTimeFormatter.ISO_DATE_TIME.format(dateTime.truncatedTo(ChronoUnit.SECONDS));
-    return Paths.get(PREFIX + "-" + timestamp.replace(':', '_') + ".jfr");
+    Path path =
+        Files.createTempFile(outputDir, PREFIX + "-" + timestamp.replace(':', '_') + "-", ".jfr");
+    System.err.println("created path" + path + " " + Files.isReadable(path));
+    new Exception().printStackTrace();
```
breedx-splk, line 50: or this.

breedx-splk (commented):
Sorry for the premature review earlier (failed to realize it was still in draft).

I think I lost some review comments (concerns) around the use of `Files.createTempFile()`, but basically, the javadoc on that method includes:

* `The details as to how the name of the file is constructed is implementation dependent and therefore not specified.`
* With that, it does say that it tries to follow `File.createTempFile()` which says `the new file will be generated by concatenating the prefix, five or more internally-generated characters, and the suffix.`. Even if we've observed these characters to be numeric (`\d`), I'm worried that there is no assurance of that and the regex could break.

The original intent of the issue was to fend against multiple VMs in a single container, and this method provides no guarantee that two VMs won't use the same strategy and collide.

Have we thought about using pid?

laurit:
> Sorry for the premature review earlier (failed to realize it was still in draft).
> 
I needed to do some debugging and changed the pr back to draft. There was a test that behaved differently on mac vs linux host so I couldn't reproduce the issue locally.

> I think I lost some review comments (concerns) around the use of `Files.createTempFile()`, but basically, the javadoc on that method includes:
> 
> * `The details as to how the name of the file is constructed is implementation dependent and therefore not specified.`
> * With that, it does say that it tries to follow `File.createTempFile()` which says `the new file will be generated by concatenating the prefix, five or more internally-generated characters, and the suffix.`. Even if we've observed these characters to be numeric (`\d`), I'm worried that there is no assurance of that and the regex could break.
> 
Would it be fine with you if I relaxed the regex, could use `.*`.? The regex is only used by `RecordingEscapeHatch` as far as I can tell.

> The original intent of the issue was to fend against multiple VMs in a single container, and this method provides no guarantee that two VMs won't use the same strategy and collide.
> 
I believe this is where you are mistaken, temporary files are guaranteed to be unique. Temporary file creation will retry until it comes up with a path that doesn't already exist.

> Have we thought about using pid?
If I remember correctly getting pid isn't easy before jdk 9, best not to use it.

breedx-splk:
> Would it be fine with you if I relaxed the regex, could use `.*`.? The regex is only used by `RecordingEscapeHatch` as far as I can tell.

I think that's a reasonable compromise, sure.
 
> > The original intent of the issue was to fend against multiple VMs in a single container, and this method provides no guarantee that two VMs won't use the same strategy and collide.
> 
> I believe this is where you are mistaken, temporary files are guaranteed to be unique. Temporary file creation will retry until it comes up with a path that doesn't already exist.

Is that documented somewhere?

> > Have we thought about using pid?
> 
> If I remember correctly getting pid isn't easy before jdk 9, best not to use it.

Yeah ok. Bummer. 

Let's change the regex and then go with this. It's definitely an improvement.

laurit:
> Is that documented somewhere?
> 
See https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempFile(java.nio.file.Path,%20java.lang.String,%20java.lang.String,%20java.nio.file.attribute.FileAttribute...)
`Returns: the path to the newly created file that did not exist before this method was invoked`

PR #764, opened 2022-05-02
Use a specialized class instead of string as key in pprof tables

laurit:
Key class instances require less memory than string concatenation

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/pprof/Pprof.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/pprof/Pprof.java
@@ -44,10 +44,6 @@ public class Pprof {
     return stringTable.get(str);
   }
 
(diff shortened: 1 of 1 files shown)


breedx-splk:
Would be cool to have a little description about _why_ the change is good. :)

profiler/src/main/java/com/splunk/opentelemetry/profiler/pprof/Pprof.java
```diff
@@ -181,4 +205,31 @@ long get(String file, String function, long line) {
           });
     }
   }
+
+  private static class LocationKey {
+    private final String file;
+    private final String function;
```
breedx-splk, line 211: You can delegate to a `FunctionKey` here.

breedx-splk (approved):
This seems ok to me. It would be nice to have some concrete results to see the actual savings here tho.

Issue #795, opened 2022-05-20
Setting OTEL_EXPORTER_OTLP_ENDPOINT does not override the SPLUNK_REALM property.


atoulme:
The `OTEL_EXPORTER_OTLP_ENDPOINT` does not override the `splunk.realm` property. The `splunk.realm` setting applies and overrides `OTEL_EXPORTER_OTLP_ENDPOINT`.

The [documentation](https://github.com/signalfx/splunk-otel-java/blob/d5da8ef6bf99ff7652da12eb84270e2756cdc468/docs/advanced-config.md) states on the property `OTEL_EXPORTER_OTLP_ENDPOINT`:
> The OTLP endpoint to connect to. Setting this will override the `splunk.realm` property.

Steps to reproduce:
* Setting the environment variable `OTEL_EXPORTER_OTLP_ENDPOINT` to example.com
* Set the `SPLUNK_REALM` environment variable to `us1`

Expected:
* Traces should be sent to example.com.
Actual:
* Traces are sent to ingest.us1.signalfx.com.

PR #796, opened 2022-05-23
Set otel.exporter.otlp.endpoint instead of otel.exporter.otlp.traces.…

laurit:
…endpoint when realm is set
Resolves https://github.com/signalfx/splunk-otel-java/issues/795


breedx-splk (approved):
Thanks for picking this up! 🥇

PR #800, opened 2022-05-24
Change metric names for memory allocation/reclaim.

breedx-splk:
This changes the profiling memory allocation metrics names. It wasn't immediately clear from the names that these metrics are monotonically increasing (cumulative) values, so I added `cumulative` and removed the `experimental` namespace.

I am not married to these names, so please feel free to suggest some improvements. The goal was to keep them in line with the [otel spec jvm metric names](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics), even tho these are not yet present.

### DIFF
--- a/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
+++ b/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Set;
 
 public class AllocatedMemoryMetrics implements MeterBinder {
+  public static final String METRIC_NAME = "process.runtime.jvm.memory.allocated.cumulative";
   private final boolean hasComSunThreadMXBean = hasComSunThreadMXBean();
 
   @Override
@@ -41,9 +42,10 @@ public class AllocatedMemoryMetrics implements MeterBinder {
     }
 
     AllocationTracker allocationTracker = new AllocationTracker();
-    Gauge.builder("jvm.experimental.memory.allocated", allocationTracker::getAllocated)
+    Gauge.builder(METRIC_NAME, allocationTracker::getCumulativeAllocationTotal)
         .description("Approximate sum of heap allocations")
         .baseUnit(BaseUnits.BYTES)
+        .tag("type", "heap")
         .register(registry);
   }
 
@@ -78,7 +80,7 @@ public class AllocatedMemoryMetrics implements MeterBinder {
     // allocated bytes by threads that have terminated
     private long allocatedByDeadThreads = 0;
 
-    long getAllocated() {
+    long getCumulativeAllocationTotal() {
       Set<Long> threads = new HashSet<>();
       long allocatedByLiveThreads = 0;
       long[] threadIds = threadBean.getAllThreadIds();
--- a/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
+++ b/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
@@ -28,7 +28,7 @@ import java.util.Map;
 import java.util.Set;
 
 public class AllocatedMemoryMetrics implements MeterBinder {
-  public static final String METRIC_NAME = "process.runtime.jvm.memory.allocated.cumulative";
+  public static final String METRIC_NAME = "process.runtime.jvm.memory.allocated";
   private final boolean hasComSunThreadMXBean = hasComSunThreadMXBean();
 
   @Override
--- a/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/GcMemoryMetrics.java
+++ b/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/GcMemoryMetrics.java
@@ -40,7 +40,7 @@ import javax.management.NotificationListener;
 import javax.management.openmbean.CompositeData;
 
 public class GcMemoryMetrics implements MeterBinder, AutoCloseable {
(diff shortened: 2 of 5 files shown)


instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/GcMemoryMetrics.java
```diff
@@ -55,7 +56,7 @@ public void bindTo(MeterRegistry registry) {
     GcMetricsNotificationListener gcNotificationListener =
         new GcMetricsNotificationListener(registry);
 
-    Gauge.builder("jvm.experimental.gc.memory.reclaimed", deltaSum, AtomicLong::get)
+    Gauge.builder(METRIC_NAME, deltaSum, AtomicLong::get)
```
laurit, line 59: as far as I can tell this metric reports deltas. @gvsmirnov would it be better if this metric reported cumulative value?

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
```diff
@@ -41,9 +42,10 @@ public void bindTo(MeterRegistry registry) {
     }
 
     AllocationTracker allocationTracker = new AllocationTracker();
-    Gauge.builder("jvm.experimental.memory.allocated", allocationTracker::getAllocated)
+    Gauge.builder(METRIC_NAME, allocationTracker::getCumulativeAllocationTotal)
```
laurit, line 45: I don't think we should add `cumulative` to the name because none of the other metrics that represent cumulative values have on. For example `jvm.gc.memory.allocated`, `db.pool.connections.timeouts`. If we decide to add something then maybe `total` instead of `cumulative`? Should we change this from gauge to counter?
@gvsmirnov would gauge -> counter be ok or would you prefer if this metic to reported delta instead?

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/GcMemoryMetrics.java
gsmirnov-splk: Whoopsie, missed this, sry. @gvsmirnov is my private account, @gsmirnov-splk is my work account.

Looking at the data, it already reports a cumulative value -- starts at 0 and only ever goes up until restart. 🤔 Looking at the code, it reports the cumulative sum of deltas of memory used before&after GC. It is good as it is now.

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
gsmirnov-splk: It is currently reporting the cumulative number of bytes allocated since JVM start. I am happy to keep the values like this. I am also fine with it becoming a counter, though the way we measure looks more like a gauge.

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/GcMemoryMetrics.java
laurit: @gsmirnov-splk Sorry didn't remember that you have 2 accounts, for some reason github doesn't autocomplete your handle.
Seems like I misread, it is cumulative not delta.

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
laurit: Similar micrometer metrics `jvm.gc.memory.allocated` and `jvm.gc.memory.promoted` are counters. https://github.com/open-telemetry/opentelemetry-java/blob/v1.14.0/api/all/src/main/java/io/opentelemetry/api/metrics/Meter.java#L25 hints that counter might be more appropriate for metrics where one wishes to track rate of change. Though this particular piece of javadoc has been rewritten in latest version.

gsmirnov-splk: Yeah, conceptually these are both counters. But in case of allocated bytes, the way we measure them does not quite allow us to do `counter.add(delta)` and there is no `counter.set()` in the API as far as I can see.

laurit: We could try
```java
    FunctionCounter.builder("jvm.experimental.memory.allocated", allocationTracker, (unused) -> allocationTracker.getAllocated())
        .description("Approximate sum of heap allocations")
        .baseUnit(BaseUnits.BYTES)
        .register(registry);
```
I wouldn't be too concerned about how this looks in the agent but rather how backend handles these. Like counters and gauges having different rollup. Unfortunately I don't know enough of the topic to say with certainty how these should be handled.

laurit (approved):
@breedx-splk your call whether to keep the names as they are or change something

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
gsmirnov-splk: I think the backend does NOT offer the "rate" **rollup** on gauges, but so far we've been using a "rate of change" **transformation** and it worked. But if we can use a counter that would be great

breedx-splk:
@laurit Sorry it took me some time to come around on this, but I think I do agree that the `cumulative` word in the name is inconsistent with the other metrics and can be handled with documentation.

PR #808, opened 2022-06-09
Implement allocation metrics as otel metrics

laurit:


instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
```diff
@@ -42,12 +39,7 @@ public void bindTo(MeterRegistry registry) {
     }
 
     AllocationTracker allocationTracker = new AllocationTracker();
-    FunctionCounter.builder(
-            METRIC_NAME, allocationTracker, AllocationTracker::getCumulativeAllocationTotal)
-        .description("Approximate sum of heap allocations")
-        .baseUnit(BaseUnits.BYTES)
-        .tag("type", "heap")
-        .register(registry);
+    meterFactory.accept(allocationTracker, allocationTracker::getCumulativeAllocationTotal);
```
breedx-splk, line 42: This feels unnecessarily callbacky to me now. Why not get rid of `install` completely and replace it with a public method that just delegates to `allocationTracker.getCumulativeAllocationTotal()`? With that approach, the two implementations can just call that directly, rather than going thru an extra `Supplier`?

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/GcMemoryMetrics.java
```diff
@@ -33,33 +29,29 @@
...
 
-    GcMetricsNotificationListener gcNotificationListener =
-        new GcMetricsNotificationListener(registry);
+    GcMetricsNotificationListener gcNotificationListener = new GcMetricsNotificationListener();
 
-    FunctionCounter.builder(METRIC_NAME, deltaSum, AtomicLong::get)
-        .description("Sum of heap size differences before and after gc")
-        .baseUnit(BaseUnits.BYTES)
-        .register(registry);
+    meterFactory.accept(deltaSum);
```
breedx-splk, line 54: Same comment as `AllocatedMemoryMetrics` -- seems overly callbacky. Let's just put a method on here to get the value.

instrumentation/jvm-metrics/src/test/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/JvmMetricsTest.java
```diff
@@ -64,6 +64,10 @@ void shouldRegisterOtelJvmMeters() {
     assertOtelMetricPresent("runtime.jvm.memory.used", "bytes");
     // thread metrics
     assertOtelMetricPresent("runtime.jvm.threads.peak", "threads");
+    // allocated memory metrics
+    assertOtelMetricPresent(AllocatedMemoryMetrics.METRIC_NAME, "bytes");
+    // Our custom GC metrics
+    assertOtelMetricPresent(GcMemoryMetrics.METRIC_NAME, "bytes");
```
breedx-splk, line 70: Is it worth trying to verify the type of these metrics?

breedx-splk (changes requested):

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
laurit: I went with callbacks specially for you :) An alternative would be to use inheritance and use this class as a super class and implement registering the meter in a child class. Would you like that more?

laurit: The problem is that the `install` method also does some other things besides invoking callback. So it really would look more like
```java
public void install() {
  AllocatedMemoryMetrics metrics = new AllocatedMemoryMetrics();
  if (!metrics.canEnable()) {
    return;
  }
  createMeter();
  metrics.registerListener(); // only for gc metrics
}
```
so there will be a bit of duplication. Is this how you'd like it?

breedx-splk: Right, it also contains those checks to see if we even _can_ do the memory metrics stuff. This might be highlighting the fact that the class has more than one responsibility.

I think this duplication isn't egregious and does make things slightly cleaner.

instrumentation/jvm-metrics/src/test/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/JvmMetricsTest.java
laurit: removed

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
laurit: changed as requested

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
```diff
@@ -17,42 +17,30 @@
...
-    FunctionCounter.builder(
-            METRIC_NAME,
-            allocationTracker,
-            (unused) -> allocationTracker.getCumulativeAllocationTotal())
-        .description("Approximate sum of heap allocations")
-        .baseUnit(BaseUnits.BYTES)
-        .tag("type", "heap")
-        .register(registry);
+  private AllocationTracker createAllocationTracker() {
+    if (hasComSunThreadMXBean && isThreadAllocatedMemoryEnabled()) {
```
breedx-splk, line 40: This is confusing now because it relies on the `hasComSunThreadMXBean` field to have been assigned before we get here, which now relies on declaration ordering, which is fragile. Better to remove the `hasComSunThreadMXBean` field altogether and just invoke the method.

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
```diff
@@ -17,42 +17,30 @@
...
-    }
+  public boolean isAvailable() {
+    return allocationTracker != null;
+  }
 
-    if (!isThreadAllocatedMemoryEnabled()) {
-      return;
-    }
+  public long getCumulativeAllocationTotal() {
+    return allocationTracker.getCumulativeAllocationTotal();
```
breedx-splk, line 36: This will NPE in the cases where `createAllocationTracker()` returns null.

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
laurit: `hasComSunThreadMXBean` is a final field, I changed it also to be static. Is this enough?

breedx-splk: Sure. I think it's less readable than calling it inline because the reader has to understand when/where that field was assigned, but it's fine.

breedx-splk (approved):

PR #809, opened 2022-06-09
Fix allocation counter

laurit:


instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
```diff
@@ -42,8 +42,13 @@ public void bindTo(MeterRegistry registry) {
...
 
     AllocationTracker allocationTracker = new AllocationTracker();
+    // FunctionCounter keeps a weak reference to allocationTracker. To ensure it is not collected
+    // we pass a capturing lambda as the function instead of method reference
+    // AllocationTracker::getCumulativeAllocationTotal
     FunctionCounter.builder(
-            METRIC_NAME, allocationTracker, AllocationTracker::getCumulativeAllocationTotal)
+            METRIC_NAME,
+            allocationTracker,
+            (unused) -> allocationTracker.getCumulativeAllocationTotal())
```
breedx-splk, line 51: This api is goofy, no doubt. I think the intention is to have the `toDoubleFunction` (last parameter) actually reference the parameter? Anyway, good work around.

In any case, I think that the fact that there is an important semantic difference with this change is a language failure! 🤯

laurit: https://github.com/signalfx/splunk-otel-java/pull/808 will replace this anyway

PR #831, opened 2022-06-30
Add deprecation notice and remove mention from front readme

breedx-splk:
I assume we will want to keep support for `jaeger-thrift` until the SmartAgent is deprecated in December. Once that happens, we can remove `jaeger-thrift` from our distro. 

This change just adds a deprecation notice about that and removes the mention from the front page (to discourage creating new user dependencies on it).

### DIFF
--- a/docs/advanced-config.md
+++ b/docs/advanced-config.md
@@ -39,6 +39,8 @@ export SPLUNK_METRICS_ENDPOINT=https://ingest.us0.signalfx.com
 | `otel.exporter.jaeger.endpoint` | `OTEL_EXPORTER_JAEGER_ENDPOINT` | `http://localhost:9080/v1/trace` | Stable  | The Jaeger endpoint to connect to. Setting this will override the `splunk.realm` property.                                               |
 | `otel.traces.exporter`          | `OTEL_TRACES_EXPORTER`          | `otlp`                           | Stable  | Select the traces exporter to use. We recommend using either the OTLP exporter (`otlp`) or the Jaeger exporter (`jaeger-thrift-splunk`). |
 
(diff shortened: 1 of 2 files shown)


aurbiztondo-splunk (approved):

mateuszrzeszutek (approved):

docs/advanced-config.md
```diff
@@ -39,6 +39,8 @@ export SPLUNK_METRICS_ENDPOINT=https://ingest.us0.signalfx.com
 | `otel.exporter.jaeger.endpoint` | `OTEL_EXPORTER_JAEGER_ENDPOINT` | `http://localhost:9080/v1/trace` | Stable  | The Jaeger endpoint to connect to. Setting this will override the `splunk.realm` property.                                               |
 | `otel.traces.exporter`          | `OTEL_TRACES_EXPORTER`          | `otlp`                           | Stable  | Select the traces exporter to use. We recommend using either the OTLP exporter (`otlp`) or the Jaeger exporter (`jaeger-thrift-splunk`). |
 
+:warning: **Support for the `jaeger-thrift-splunk` exporter will be removed after December 17th, 2022. See the [deprecation notice](https://github.com/signalfx/signalfx-agent/blob/main/docs/smartagent-deprecation-notice.md) on the SmartAgent for details. ** :warning:
```
laurit, line 42: I don't think this is going to be that easy. Isn't this the recommended way for direct ingest also? Before deprecating it shouldn't there be an alternative for it? I don't think that any of the other exporters do anything with `splunk.access.token`

laurit: I was wrong, `splunk.access.token` is also used for `otlp` exporter, so deprecating should be ok

PR #838, opened 2022-07-08
Fix build after upstream changes

laurit:
Resolves https://github.com/signalfx/splunk-otel-java/issues/837


instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/JvmMetricsInstaller.java
```diff
@@ -83,4 +82,22 @@ public void afterAgent(
...
     }
   }
+
+  private static boolean isInstrumentationEnabled(
+      ConfigProperties config, Iterable<String> instrumentationNames, boolean defaultEnabled) {
+    // If default is enabled, we want to enable individually,
+    // if default is disabled, we want to disable individually.
+    boolean anyEnabled = defaultEnabled;
+    for (String name : instrumentationNames) {
+      String propertyName = "otel.instrumentation." + name + ".enabled";
```
mateuszrzeszutek, line 92: We should also normalize the name
```suggestion
      String propertyName = "otel.instrumentation." + name.replace('-', '.') + ".enabled";
```

mateuszrzeszutek (approved):

PR #839, opened 2022-07-08
Fix deprecation warning

laurit:


mateuszrzeszutek:
`ConfigProperties` still doesn't normalize the property name, and some of the properties used here use hyphes. WDYT about waiting with merging this until the next SDK release (which should happen this week or the next week)?

laurit:
> `ConfigProperties` still doesn't normalize the property name, and some of the properties used here use hyphes. WDYT about waiting with merging this until the next SDK release (which should happen this week or the next week)?

Sure, there is no hurry with this.

mateuszrzeszutek (approved):

PR #864, opened 2022-08-03
Fix allocation metric

laurit:


breedx-splk (commented):
I was able to reproduce the zero-value metric, but this change didn't actually fix it for me. With this change in place, the behavior I observed went from zero value metric to no metric at all!  🤯

breedx-splk (approved):

laurit:
> I was able to reproduce the zero-value metric, and initially I thought this change failed to fix it for me...but I had left the field instead of moving it to the local variable. Moving it to a local var does indeed seem to fix things. 🤘🏻

It should not make any difference whether it is a field or a local variable, the important part is avoiding it getting gcd.

PR #866, opened 2022-08-04
Use "By" instead of "bytes" as unit

laurit:
Upstream uses `"By"`


instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelJvmGcMetrics.java
```diff
@@ -107,29 +107,29 @@ public void install() {
     maxDataSize = new AtomicLong((long) maxLongLivedPoolBytes);
     meter
         .gaugeBuilder("runtime.jvm.gc.max.data.size")
-        .setUnit("bytes")
+        .setUnit("By")
```
mateuszrzeszutek, line 110: This is one of the classes copied from micrometer, right? Shouldn't we avoid changing metrics in these?

laurit: Good point, I'll close this pr

mateuszrzeszutek: I think it's fine to change the unit for the other two, "our" classes

PR #882, opened 2022-08-17
Update javaagent version in splunk-otel-collector after release

mateuszrzeszutek:


scripts/propagate-version.sh
```diff
@@ -83,8 +83,34 @@ create_overhead_test_pr() {
     --head "$update_version_branch"
 }
 
+create_collector_pr() {
+  local repo="signalfx/splunk-otel-collector"
+  local repo_url="https://srv-gh-o11y-gdi:${GITHUB_TOKEN}@github.com/${repo}.git"
+  local update_version_branch="javaagent-version-update-$release_tag"
```
breedx-splk, line 89: Gotta love shell scripts! Where does `release_tag` come from? It's just an assumed global?

breedx-splk (approved):

scripts/propagate-version.sh
mateuszrzeszutek: It's an actual global in this file, defined in line 25

laurit (approved):

PR #886, opened 2022-08-23
Auto-detect service name from web.xml display-name

laurit:


theletterf (approved):

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
```diff
@@ -187,6 +189,16 @@ Map<String, String> customize(ConfigProperties config) {
       customized.put("otel.exporter.otlp.headers", otlpHeaders);
     }
 
+    if (config.getBoolean(SPLUNK_AUTO_DETECT_SERVICE_NAME, true)) {
+      if (ServiceNameDetector.serviceNameNotConfigured(config)) {
+        String serviceName = ServiceNameDetector.detectServiceName();
+        if (serviceName != null) {
+          log.info("Auto-detected service name '{}'.", serviceName);
```
laurit, line 196: should we also log some kind of description (path to the web.xml?) based on what we auto-detected?

custom/src/main/java/com/splunk/opentelemetry/servicename/LibertyServiceNameDetector.java
```diff
@@ -0,0 +1,56 @@
...
+    String wlpOutputDir = System.getenv("WLP_OUTPUT_DIR");
+    log.debug("Using WLP_USER_DIR '{}', WLP_OUTPUT_DIR '{}'.", wlpUserDir, wlpOutputDir);
+    if (wlpUserDir != null
+        && wlpOutputDir != null
+        && !Paths.get(wlpOutputDir).equals(Paths.get(wlpUserDir, "servers"))) {
+      Path serverName = Paths.get(wlpOutputDir).getFileName();
+      serverDir = Paths.get(wlpUserDir).resolve(serverName);
+    }
+
+    return serverDir.resolve("dropins");
```
laurit, line 54: test app is added to xml https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/892fb8a38e4e3e8c7817c5836fe3902e9188fd3d/smoke-tests/images/servlet/src/main/docker/liberty/server.xml#L8 Should we attempt to parse the server.xml also?

custom/src/main/java/com/splunk/opentelemetry/servicename/TomeeServiceNameDetector.java
```diff
@@ -0,0 +1,71 @@
...
+
+class TomeeServiceNameDetector extends AppServerServiceNameDetector {
+
+  TomeeServiceNameDetector(ResourceLocator locator) {
+    super(locator, "org.apache.catalina.startup.Bootstrap", true);
+  }
+
+  @Override
+  Path getDeploymentDir() throws URISyntaxException {
+    Path rootDir = getRootDir();
```
laurit, line 33: our test uses webapps directory so this isn't tested

custom/src/main/java/com/splunk/opentelemetry/servicename/WebSphereServiceNameDetector.java
```diff
@@ -0,0 +1,129 @@
...
+        "Parsed arguments: cell '{}', node '{}', server '{}', configuration directory '{}'.",
+        cell,
+        node,
+        server,
+        configDirectory);
+
+    // construct installedApps directory path based on the config path
+    // in docker image it is
+    // /opt/IBM/WebSphere/AppServer/profiles/AppSrv01/installedApps/DefaultCell01
+    Path cellApplications = configDirectory.resolve("../installedApps/" + cell);
```
laurit, line 90: I believe the installedApps location is configurable so there could be installations where this is not the correct path.

custom/src/main/java/com/splunk/opentelemetry/servicename/WildflyServiceNameDetector.java
```diff
@@ -0,0 +1,57 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class WildflyServiceNameDetector extends AppServerServiceNameDetector {
+  private static final Logger log = LoggerFactory.getLogger(WildflyServiceNameDetector.class);
+
+  WildflyServiceNameDetector(ResourceLocator locator) {
+    super(locator, "org.jboss.modules.Main", true);
+  }
+
```
laurit, line 32: wildfly deploys applications from expoded directories if a marker file is present (something like appname.dodeploy). should we try to filter out exploded apps that won't be deployed?

custom/src/main/java/com/splunk/opentelemetry/servicename/JettyServiceNameDetector.java
```diff
@@ -0,0 +1,42 @@
...
+    return !name.endsWith(".d");
+  }
+
+  @Override
+  Path getDeploymentDir() {
+    // Jetty expects the webapps directory to be in the directory where jetty was started from.
+    // Alternatively the location of webapps directory can be specified by providing jetty base
+    // directory as an argument to jetty e.g. java -jar start.jar jetty.base=/dir where webapps
+    // would be located in /dir/webapps. We don't handle jetty.base yet.
+    return Paths.get("webapps").toAbsolutePath();
```
laurit, line 40: I mostly skipped the jetty.base part because it the path contains a space then extracting it from `sun.java.command` is a bit painful. I think this could be done by testing whether the path exists and if it doesn't append space + text until next space and try again. Should I implement this also?

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
mateuszrzeszutek: If it's not too much trouble, then why not -- it'd make our lives easier. I'd log it at debug though, some customers are kinda sensitive to the amount of logs printed by the agent.

custom/src/main/java/com/splunk/opentelemetry/servicename/LibertyServiceNameDetector.java
mateuszrzeszutek: I think it's fine as it is -- we can add that once somebody requests it. You could leave a comment mentioning the server.xml war location, just in case.

custom/src/main/java/com/splunk/opentelemetry/servicename/WebSphereServiceNameDetector.java
mateuszrzeszutek: Can you add a comment that mentions it? I think it's fine to hard-code it for now, but it would be nice to have a pointer in case there're any issues/requests about that.

custom/src/main/java/com/splunk/opentelemetry/servicename/JettyServiceNameDetector.java
mateuszrzeszutek: 👍 
That sounds like a useful thing.

custom/src/main/java/com/splunk/opentelemetry/servicename/WildflyServiceNameDetector.java
mateuszrzeszutek: I think it's probably fine not to do that right now; just leave a TODO about it, we can come back to it if somebody asks for it.

mateuszrzeszutek (commented):
This is awesome 👍 
Even if it's only a fraction of all possible ways to deploy servlet apps, I think it's more than enough to get us started with this -- all other cases would be covered on demand (or not at all, if they're too insane)

custom/src/main/java/com/splunk/opentelemetry/servicename/ServiceNameDetector.java
```diff
@@ -0,0 +1,84 @@
...
+
+  public static String detectServiceName() {
+    for (ServiceNameDetector detector : detectors()) {
+      try {
+        String name = detector.detect();
+        if (name != null) {
+          return name;
+        }
+      } catch (Exception exception) {
+        log.warn(
```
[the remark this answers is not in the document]
laurit: my reasoning was that if there are bugs that make this wail having it at warn level increases the chance that someone will report it as a bug

custom/src/main/java/com/splunk/opentelemetry/servicename/LibertyServiceNameDetector.java
```diff
@@ -0,0 +1,56 @@
...
+    // in default installation we already have the right directory and don't need to do anything
+    Path serverDir = Paths.get("").toAbsolutePath();
+    String wlpUserDir = System.getenv("WLP_USER_DIR");
+    String wlpOutputDir = System.getenv("WLP_OUTPUT_DIR");
+    log.debug("Using WLP_USER_DIR '{}', WLP_OUTPUT_DIR '{}'.", wlpUserDir, wlpOutputDir);
+    if (wlpUserDir != null
+        && wlpOutputDir != null
+        && !Paths.get(wlpOutputDir).equals(Paths.get(wlpUserDir, "servers"))) {
+      Path serverName = Paths.get(wlpOutputDir).getFileName();
+      serverDir = Paths.get(wlpUserDir).resolve(serverName);
```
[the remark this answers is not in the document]
laurit: good catch, unfortunately we don't have a test for this

custom/src/main/java/com/splunk/opentelemetry/servicename/AppServerServiceNameDetector.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    Path deploymentDir = getDeploymentDir();
+    if (deploymentDir == null) {
+      return null;
+    }
+
+    if (Files.isDirectory(deploymentDir)) {
+      log.debug("Looking for deployments in '{}'.", deploymentDir);
+      for (Path path : Files.list(deploymentDir).collect(Collectors.toList())) {
```
mateuszrzeszutek, line 78: Just noticed it now: `Files.list()` returns a stream that must be closed; can you run it in a try-with-resources block?

custom/src/main/java/com/splunk/opentelemetry/servicename/AppServerServiceNameDetector.java
```diff
@@ -0,0 +1,259 @@
...
+    } else if (name.endsWith(".war")) {
+      return handlePackagedWar(path);
+    } else if (supportsEar && name.endsWith(".ear")) {
+      return handlePackagedEar(path);
+    }
+
+    return null;
+  }
+
+  private String handleExplodedApp(Path path, boolean handleEar) {
```
mateuszrzeszutek, line 110: Nit: `supportsEar` is a field in this class, no need to pass it explicitly

custom/src/main/java/com/splunk/opentelemetry/servicename/WebSphereServiceNameDetector.java
```diff
@@ -0,0 +1,131 @@
...
+
+    // construct installedApps directory path based on the config path
+    // in docker image it is
+    // /opt/IBM/WebSphere/AppServer/profiles/AppSrv01/installedApps/DefaultCell01
+    // NOTE: installedApps directory location is configurable
+    Path cellApplications = configDirectory.getParent().resolve("installedApps").resolve(cell);
+    if (Files.isDirectory(cellApplications)) {
+      log.debug("Looking for deployments in '{}'.", cellApplications);
+
+      for (Path path : Files.list(cellApplications).collect(Collectors.toList())) {
```
mateuszrzeszutek, line 96: Both `Files.list()` streams in this method need to be closed, unfortunately

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
```diff
@@ -187,6 +189,16 @@ Map<String, String> customize(ConfigProperties config) {
       customized.put("otel.exporter.otlp.headers", otlpHeaders);
     }
 
+    if (config.getBoolean(SPLUNK_AUTO_DETECT_SERVICE_NAME, true)) {
+      if (ServiceNameDetector.serviceNameNotConfigured(config)) {
+        String serviceName = ServiceNameDetector.detectServiceName();
```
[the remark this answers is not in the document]
laurit: changed to resource provider

custom/src/main/java/com/splunk/opentelemetry/servicename/AppServerServiceNameDetector.java
```diff
@@ -0,0 +1,262 @@
...
+      }
+    }
+  }
+
+  private static class SaxParserFactoryHolder {
+    private static final SAXParserFactory saxParserFactory = getSaxParserFactory();
+
+    private static SAXParserFactory getSaxParserFactory() {
+      try {
+        return SAXParserFactory.newInstance();
```
breedx-splk, line 255: Odd implementation -- I'd have expected a holder to hold and return a single instance rather than instantiating new ones each time.

breedx-splk: Oh, I see it now, it's just direct referenced. 🤷🏻

smoke-tests/src/test/java/com/splunk/opentelemetry/SmokeTest.java
```diff
@@ -66,6 +66,10 @@ protected List<ResourceMapping> getExtraResources() {
     return List.of();
   }
 
+  protected boolean getAutodetectServiceName() {
```
breedx-splk, line 69: NIt: might have been spelled `shouldGetAutodetectServiceName()`?

smoke-tests/src/test/java/com/splunk/opentelemetry/WebSphereSmokeTest.java
```diff
@@ -28,13 +28,13 @@
 public class WebSphereSmokeTest extends AppServerTest {
 
   public static final ExpectedServerAttributes WEBSPHERE8_SERVER_ATTRIBUTES =
-      new WebSphereAttributes("8.5.5.20");
+      new WebSphereAttributes("8.5.5.19");
```
breedx-splk, line 31: Why older?

custom/src/main/java/com/splunk/opentelemetry/servicename/JettyServiceNameDetector.java
```diff
@@ -0,0 +1,91 @@
...
+      if (nextSpace == -1) {
+        Path candidate = Paths.get(programArguments.substring(start));
+        return Files.exists(candidate) ? candidate : null;
+      }
+      Path candidate = Paths.get(programArguments.substring(start, nextSpace));
+      next = nextSpace + 1;
+      if (Files.exists(candidate)) {
+        return candidate;
+      }
+    }
```
breedx-splk, line 69: I totally think that this is fine for now, but I suspect there's some remotely possible edge case out there that could cause this to parse the commandline string differently than a shell might have parsed it, resulting in a path that includes more than the user specified. I'm speculating, but imagine a dir with two subdirs

```
jetty
   +--- jetty base 
   +--- jetty base dev
```

If the user runs a commandline like
```
jetty <whatever> "jetty.base=jetty base dev" jetty.foo=bar
```
and if the JVM reconstitutes this into the `sun.java.command` as 
```
jetty <whatever> jetty.base=jetty base dev jetty.foo=bar
``` `
then I _think_ this will end up using `jetty base` as the path even tho the user intended `jetty base dev`. 

Seems very unlikely that this will happen, so I dunno if it's worth trying to fix now. #famouslastwords

File: custom/src/test/java/com/splunk/opentelemetry/servicename/AppServerServiceNameDetectorTest.java [hunk 17]
```diff
@@ -0,0 +1,177 @@
...
+import java.nio.file.Paths;
+import java.util.function.Consumer;
+import java.util.jar.Attributes;
+import java.util.jar.JarEntry;
+import java.util.jar.JarOutputStream;
+import java.util.jar.Manifest;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+public class AppServerServiceNameDetectorTest {
```
breedx-splk, line 35: I don't think this is really a unit test because it does so much with the environment and tests across detectors. Feels like an integration test to me. Is there a better home for this?

breedx-splk (approved):
Agree with others that this is fantastic! I would have loved to see it broken up into more bite sized PRs for reviewing, but great anyway. 👍🏻

smoke-tests/src/test/java/com/splunk/opentelemetry/WebSphereSmokeTest.java
laurit: this is because we rely on images build in upstream repo and have to use the same versions. I think at some point there was an issue with `8.5.5.20` image so it got downgraded to `8.5.5.19` to get the build passing. if you want to change this I'd suggest you first change it in upstream and when that change is merged change it here also

custom/src/main/java/com/splunk/opentelemetry/servicename/JettyServiceNameDetector.java
laurit: It should be possible to do it accurately with https://docs.oracle.com/javase/9/docs/api/java/lang/ProcessHandle.Info.html#arguments-- on jdk9

custom/src/test/java/com/splunk/opentelemetry/servicename/AppServerServiceNameDetectorTest.java
laurit: It does not test across detectors. It only test logic in the base detector class.

smoke-tests/src/test/java/com/splunk/opentelemetry/SmokeTest.java
laurit: renamed to `shouldAutodetectServiceName`

custom/src/main/java/com/splunk/opentelemetry/servicename/ServiceNameResourceProvider.java
```diff
@@ -0,0 +1,50 @@
...
+public class ServiceNameResourceProvider implements ResourceProvider {
+  private static final Logger log = LoggerFactory.getLogger(ServiceNameResourceProvider.class);
+
+  @Override
+  public Resource createResource(ConfigProperties config) {
+    if (ServiceNameDetector.serviceNameNotConfigured(config)) {
+      String serviceName = ServiceNameDetector.detectServiceName();
+      if (serviceName != null) {
+        log.info("Auto-detected service name '{}'.", serviceName);
+        // customized.put("otel.service.name", serviceName);
```
mateuszrzeszutek, line 39:
```suggestion
```

mateuszrzeszutek (approved):
Looks great! 👍

smoke-tests/src/test/java/com/splunk/opentelemetry/WebSphereSmokeTest.java
breedx-splk: Cool, thanks for clarifying.

custom/src/main/java/com/splunk/opentelemetry/servicename/JettyServiceNameDetector.java
breedx-splk: Yeah totally. A change for another time tho.

PR #890, opened 2022-08-29
Update logback and slf api

breedx-splk:
Going to see if this work better than #888. It looks like, according to [the docs](https://www.slf4j.org/codes.html#StaticLoggerBinder), 

> Backends such as logback 1.3 and later which target slf4j-api 2.x, do not ship with org.slf4j.impl.StaticLoggerBinder. If you place a logging backend which targets slf4j-api 2.0.x, you need slf4j-api-2.x.jar on the classpath.

So I think adding slf-api 2.0.0 into the test runtime classpath should help.

### DIFF
--- a/smoke-tests/build.gradle.kts
+++ b/smoke-tests/build.gradle.kts
@@ -16,7 +16,8 @@ dependencies {
   testImplementation("io.opentelemetry.proto:opentelemetry-proto")
   testImplementation("io.opentelemetry:opentelemetry-api")
   testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
-  testImplementation("ch.qos.logback:logback-classic:1.2.11")
+  testImplementation("ch.qos.logback:logback-classic:1.4.0")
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

mateuszrzeszutek:
That reminds me, I've an open PR updating the slf4j-api in the upstream...

mateuszrzeszutek (changes requested):
Hold on with this change:
```
java.lang.UnsupportedClassVersionError: ch/qos/logback/classic/spi/LogbackServiceProvider has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```
Looks like we might have to stay on slf4j 1.x for the time being

breedx-splk:
Should we just close this now then?

mateuszrzeszutek:
Yeah, let's close it for now -- the upstream `testing-common` package does not use slf4j 2 either.

PR #895, opened 2022-09-01
Add CloudFoundry buildpack index

breedx-splk:
This is in support of the CF buildpack work. Will need to be in place for things to function.


scripts/update-version-after-release.sh
```diff
@@ -53,3 +53,7 @@ readme_sed_args=(
 sed ${sed_flag} "${readme_sed_args[@]}" README.md
 
 rm dev_docs_warning.md.tmp
+
+# update main cloudfoundry buildpack index with new released version
+CF_INDEX=deployments/cloudfoundry/buildpack/index.yml
```
mateuszrzeszutek, line 58: AFAIR the `index.yml` file has to be present in the top directory of the repo to be detected by the java-buildpack

breedx-splk: Hmmm....it looked like there is/was prior art around being able to [specify a location](https://raw.githubusercontent.com/datadog/dd-trace-java/cloudfoundry/). I think as long as the config matches that's fine.

mateuszrzeszutek: I get a `400: Invalid request` on that page

mateuszrzeszutek: But anyway, if it's possible to set up a base path, my suggestion would be to use `deployments/cloudfoundry/` so that it does not get confused with the already existing custom buildpack that we're providing.

breedx-splk: Yeah ok. 👍🏻

breedx-splk: > I get a `400: Invalid request` on that page

Yeah sorry I linked to the wrong file. I intended to show how you can specify the exact location in the config. Fixed the link above.

mateuszrzeszutek (approved):

PR #905, opened 2022-09-06
Add display name to the sample JBoss app

mateuszrzeszutek:


laurit (approved):
I believe that the actual image needs to be built and published manually.

mateuszrzeszutek:
> I believe that the actual image needs to be built and published manually.

Yeah, I'm going to do that this week.

PR #908, opened 2022-09-07
Enable testing service name detection for JBoss EAP

mateuszrzeszutek:


smoke-tests/src/test/java/com/splunk/opentelemetry/JBossEapSmokeTest.java
```diff
@@ -42,13 +42,6 @@ private static Stream<Arguments> jboss() {
         .stream();
   }
 
-  @Override
-  protected boolean shouldAutodetectServiceName() {
-    // web.xml in jboss eap test image doesn't have a display-name tag, enable after
-    // matrix/src/main/webapp/WEB-INF/web.xml is updated and image is rebuilt
-    return false;
-  }
-
```
laurit, line 51: should be
```java
  @Override
  protected boolean shouldAutodetectServiceName() {
    return true;
  }
```
base class method returns false, the override was only used to add the comment
maybe we should also run the proprietary image tests on pr? or add a label that could be used to run them

mateuszrzeszutek: > or add a label that could be used to run them

👍

PR #910, opened 2022-09-12
Update SDK to 1.18 and use the ConditionalResourceProvider

mateuszrzeszutek:


custom/src/main/java/com/splunk/opentelemetry/servicename/ServiceNameResourceProvider.java
```diff
@@ -22,28 +22,33 @@
...
 
+  @Override
+  public boolean shouldApply(ConfigProperties config, Resource existing) {
+    return ServiceNameChecker.serviceNameNotConfigured(config, existing);
+  }
+
   @Override
   public int order() {
-    return 1000;
+    // make it run later than the spring boot resource provider (100)
```
breedx-splk, line 51: But 1000 > 100, just sayin'. 😁

breedx-splk (approved):
LGTM!

custom/src/main/java/com/splunk/opentelemetry/servicename/ServiceNameResourceProvider.java
mateuszrzeszutek: Yeah but I thought of using nice 100-200-300 progression 😄 
The JAR name one will get 300

PR #914, opened 2022-09-13
Disable the Kafka metrics reporter; update micrometer

mateuszrzeszutek:


custom/src/test/java/com/splunk/opentelemetry/SplunkConfigurationTest.java
```diff
@@ -172,12 +173,14 @@ private static void verifyThatMicrometerInstrumentationsAreDisabled(ConfigProper
...
             "oracle-ucp-splunk",
             "tomcat-jdbc-splunk",
-            "vibut-dbcp-splunk")) {
+            "vibur-dbcp-splunk")) {
       assertInstrumentationDisabled(config, instrumentationName);
     }
   }
 
   private static void assertInstrumentationDisabled(ConfigProperties config, String name) {
-    assertFalse(config.getBoolean("otel.instrumentation." + name + ".enabled", false));
```
mateuszrzeszutek, line 181: I noticed that these assertions were pretty crappy, and there were several errors here -- so I replaced them all with assertj; it should be a bit easier to catch these kind of errors now

aurbiztondo-splunk (approved):

laurit (approved):

PR #920, opened 2022-09-15
v1.16.0 post release changes

srv-gh-o11y-gdi:
v1.16.0 post release changes

PR #921, opened 2022-09-15
remove -all suffix from agent jars

breedx-splk:
Realized that we don't need the `-all` suffix on the jar URLs.

I can rebase this after #920.


laurit (approved):

laurit:
I think we should stop publishing the `-all` artifact.

mateuszrzeszutek (approved):

PR #925, opened 2022-09-16
Implement service.name detection based on the jar name

mateuszrzeszutek:


custom/src/main/java/com/splunk/opentelemetry/servicename/JarServiceNameProvider.java
```diff
@@ -0,0 +1,150 @@
...
+        return candidate;
+      }
+    }
+  }
+
+  private static String getServiceName(Path jarPath) {
+    String jarName = jarPath.getFileName().toString();
+    int dotIndex = jarName.lastIndexOf(".");
+    return dotIndex == -1 ? jarName : jarName.substring(0, dotIndex);
+  }
```
breedx-splk, line 143: I like that this doesn't assume `.jar` and is purely case insensitive. 👍🏻

custom/src/test/java/com/splunk/opentelemetry/servicename/JarServiceNameProviderTest.java
```diff
@@ -0,0 +1,98 @@
...
+    var serviceNameProvider = new JarServiceNameProvider(() -> args, prop -> null, file -> false);
+
+    Resource resource = serviceNameProvider.createResource(config);
+
+    assertThat(resource.getAttributes()).isEmpty();
+  }
+
+  @Test
+  void createResource_processHandleJar() {
+    var args = new String[] {"-Dtest=42", "-Xmx666m", "-jar", "/path/to/app/my-service.jar"};
```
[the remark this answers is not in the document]
mateuszrzeszutek: Good idea, will do 👍

custom/src/main/java/com/splunk/opentelemetry/servicename/JarServiceNameProvider.java
```diff
@@ -0,0 +1,150 @@
...
+    for (int i = 0; i < javaArgs.length; ++i) {
+      if ("-jar".equals(javaArgs[i]) && (i < javaArgs.length - 1)) {
+        return Paths.get(javaArgs[i + 1]);
+      }
+    }
+    return null;
+  }
+
+  @SuppressWarnings("unchecked")
+  private static String[] getArgumentsFromProcessHandle() {
```
laurit, line 95: To avoid reflection we could introduce a new module. I guess wether it is worth it depends on whether we'll have more code that need access to jdk9 apis.

custom/src/main/java/com/splunk/opentelemetry/servicename/ServletServiceNameProvider.java
```diff
@@ -28,10 +28,9 @@
 import java.util.logging.Logger;
 
 @AutoService(ResourceProvider.class)
-public class ServiceNameResourceProvider implements ConditionalResourceProvider {
+public class ServletServiceNameProvider implements ConditionalResourceProvider {
```
laurit, line 31: Maybe WebXmlServiceNameProvider or JeeServiceNameProvider is better?

laurit (approved):

custom/src/main/java/com/splunk/opentelemetry/servicename/JarServiceNameProvider.java
mateuszrzeszutek: We could use MR jar too, I guess. I went with that because it seemed the simplest way to call `ProcessHandle` -- I agree that we should replace it with something better when we introduce any more java9+ API usages.

PR #943, opened 2022-10-10
Change default `splunk.profiler.directory` 

breedx-splk:
Some users have permissions issues when writing to the current working directory. This changes the default location to System.getProperty("java.io.tmpdir") which is the system temp dir. We think more users will have success with this instead.

### DIFF
--- a/profiler/README.md
+++ b/profiler/README.md
@@ -41,21 +41,21 @@ property or as environment variables (by making them uppercase and replacing dot
 
 > We strongly recommend using defaults for the following settings.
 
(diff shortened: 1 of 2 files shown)


aurbiztondo-splunk (approved):

mateuszrzeszutek (approved):

theletterf:
We'll need a small docs update for this one. Which release is this bound to?

mateuszrzeszutek:
The incoming 1.17 one I assume

PR #965, opened 2022-10-19
Update license header

pellared:
1. "Project" term is used in an informal way and we can  skip this word here
1. "license (verb)" - grant license to something
1. I think we want to license not only the releases but also other artifacts such as the codebase


breedx-splk (approved):
😎

PR #975, opened 2022-10-27
Update advanced-config.md

theletterf:


mateuszrzeszutek:
The profiler configuration is in https://github.com/signalfx/splunk-otel-java/blob/main/profiler/README.md -- do you think we should integrate it into the `docs/` directory?
cc @breedx-splk

theletterf:
The field expected a 1:1 match of GH docs and official docs.

breedx-splk:
> The profiler configuration is in https://github.com/signalfx/splunk-otel-java/blob/main/profiler/README.md -- do you think we should integrate it into the `docs/` directory? cc @breedx-splk

I definitely don't want to maintain it in two places. Seems weird to have it in `/docs` when it could be living closer to the implementation (like it is now).

PR #982, opened 2022-11-01
Fix local build prompting for gradle enterprise 👎🏻 

breedx-splk:
With the upgrade to `3.11.3` in #977, local builds began prompting to accept the license agreement. This looked especially strange in IntelliJ, where it seemed like builds were just hung (which I guess they were, hung waiting for human input).

Anyway, this short circuits the prompt for local builds.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -13,12 +13,10 @@ plugins {
   id("com.gradle.enterprise") version "3.11.3"
 }
 
-if (System.getenv("CI") != null) {
-  gradleEnterprise {
-    buildScan {
-      termsOfServiceUrl = "https://gradle.com/terms-of-service"
-      termsOfServiceAgree = "yes"
-    }
+gradleEnterprise {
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):
Just wondering -- wouldn't the previous version disable gradle enterprise altogether for non-CI machines? Perhaps just including the plugin enables that stuff?

breedx-splk:
> Just wondering -- wouldn't the previous version disable gradle enterprise altogether for non-CI machines? Perhaps just including the plugin enables that stuff?

Yeah, I think it's the inclusion of the plugin. I also tinkered with doing a conditional include on the plugin, but then it can't resolve some of the symbols in the conditional config block and errors out that way. 🐘 💩

mateuszrzeszutek:
> but then it can't resolve some of the symbols in the conditional config block and errors out that way. 🐘 💩

maybe we need some good ol' reflection in our build files? 🙈

PR #984, opened 2022-11-04
Check for output dir writability before starting profiler

breedx-splk:
And log a specific warning about the failure modes:

* exists and couldn't be created
* exists but is not a directory
* exists but is not writable

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -47,6 +47,7 @@ import io.opentelemetry.sdk.logs.export.BatchLogRecordProcessor;
 import io.opentelemetry.sdk.logs.export.LogRecordExporter;
 import io.opentelemetry.sdk.logs.export.SimpleLogRecordProcessor;
 import io.opentelemetry.sdk.resources.Resource;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -67,34 +69,69 @@ public class JfrActivator implements AgentListener {
...
+        "PROFILER WILL NOT BE STARTED: The configured output directory "
+            + dir
+            + " "
+            + suffix
+            + ".");
+    return true;
   }
 
   private void activateJfrAndRunForever(ConfigProperties config, Resource resource) {
     // can't be null, default value is set in Configuration.getProperties
```
mateuszrzeszutek, line 126: Can you move the comment too?

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -67,34 +69,69 @@ public class JfrActivator implements AgentListener {
...
+    return false;
+  }
+
+  private boolean outdirWarn(Path dir, String suffix) {
+    logger.warning(
+        "PROFILER WILL NOT BE STARTED: The configured output directory "
+            + dir
+            + " "
+            + suffix
+            + ".");
```
mateuszrzeszutek, line 121: Should we use the string template here?
```suggestion
    logger.log(
        WARNING,
        "PROFILER WILL NOT BE STARTED: The configured output directory {0} {1}.",
        new Object[] {dir, suffix});
```

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
breedx-splk: whoopsie!

PR #997, opened 2022-11-14
Remove JarServiceNameProvider - it's present in the upstream agent

mateuszrzeszutek:


breedx-splk:
Just for our future selves: The upstream class name is `JarServiceNameDetector` but it's also `@AutoService(ResourceProvider.class)` so we don't have to reference it by name. Just calling it out here in case this helps us later.

breedx-splk (approved):

PR #1005, opened 2022-11-17
Remove some metrics that are already covered in upstream

mateuszrzeszutek:
Removed some of the duplicated metrics (OTel-based) that are already emitted in the upstream javaagent (mostly GC related)


instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelJvmHeapPressureMetrics.java
```diff
@@ -72,28 +57,15 @@ private OtelJvmHeapPressureMetrics(Duration lookback, Duration testEvery) {
...
-              "The percentage of long-lived heap pool used after the last GC event, in the range [0..1].")
-          .buildWithCallback(
-              measurement -> measurement.record(lastLongLivedPoolUsageAfterGc.get(), attributes));
-    }
+    // runtime.jvm.memory.usage.after.gc is replaced by OTel
+    //   process.runtime.jvm.memory.usage_after_last_gc{pool=<long lived pools>,type=heap} /
+    //   process.runtime.jvm.memory.limit{pool=<long lived pools>,type=heap}
 
     meter
         .gaugeBuilder("runtime.jvm.gc.overhead")
```
mateuszrzeszutek, line 71: The `runtime.jvm.gc.overhead` metrics does not have any matching counterpart on the OTel side, but I think that we should probably remove that: this sort of aggregation (% of time used for GC in the last 5 minutes) should be done in dashboards, not in the instrumentation code.

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelJvmThreadMetrics.java
```diff
@@ -50,25 +50,10 @@ public void install() {
...
-    meter
-        .gaugeBuilder("runtime.jvm.threads.live")
-        .setUnit("threads")
-        .setDescription(
-            "The current number of live threads including both daemon and non-daemon threads.")
-        .buildWithCallback(measurement -> measurement.record(threadBean.getThreadCount()));
+    // runtime.jvm.threads.peak was removed; easy to replace with
+    //   max(process.runtime.jvm.threads.count)
+    // runtime.jvm.threads.daemon is replaced by OTel process.runtime.jvm.threads.count{daemon=true}
+    // runtime.jvm.threads.live is replaced by OTel process.runtime.jvm.threads.count{daemon=false}
```
mateuszrzeszutek, line 57: There is one more meter `runtime.jvm.threads.states` in this class: there is currently no equivalent in OTel metrics, but there's an issue requesting this: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/7006
I think we can leave it as it is for now, hopefully at some point it'll get implemented in the upstream

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelJvmGcMetrics.java
```diff
@@ -98,24 +85,13 @@ public void install() {
...
+    //   process.runtime.jvm.memory.limit{pool=<long lived pools>}
 
-    liveDataSize = new AtomicLong();
-    meter
-        .gaugeBuilder("runtime.jvm.gc.live.data.size")
-        .setUnit("bytes")
-        .setDescription("Size of long-lived heap memory pool after reclamation.")
-        .buildWithCallback(measurement -> measurement.record(liveDataSize.get()));
+    // runtime.jvm.gc.live.data.size is replaced by OTel
+    //   process.runtime.jvm.memory.usage_after_last_gc{pool=<long lived pools>}
```
mateuszrzeszutek, line 95: There are two other metrics left in this class: `runtime.jvm.gc.memory.allocated` and `runtime.jvm.gc.memory.promoted`. Both of these emit data that needs to be calculated on the instrumentation side, I think it's probably not possible to reproduce these by querying the backend (dashboards).

Still, I think we should remove them: both of them depend on hard-coded GC & memory pool names, long-term maintaining them and ensuring they're correct for all GC possible implementations is going to be painful.
Also I don't know how useful these really are, maybe the "standard" memory & GC metrics are enough anyway?

breedx-splk: Ok good with that. It does seem weird that we have `memory.allocated` inside a `runtime.jvm.gc` namespace anyway. 

I'm not sure if you saw the JFR based approach in contrib...but it's also nontrivial and bakes in some expectations about names.

breedx-splk (approved):
Onward!

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelJvmHeapPressureMetrics.java
mateuszrzeszutek: removed then ✂️

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/JvmMetricsInstaller.java
```diff
@@ -59,8 +57,25 @@ public void afterAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetr
...
-      new OtelJvmGcMetrics().install();
-      new OtelJvmHeapPressureMetrics().install();
+      // gc metrics were removed:
+      //   runtime.jvm.gc.concurrent.phase.time is replaced by OTel
+      //     process.runtime.jvm.gc.duration{gc=<concurrent gcs>}
+      //   runtime.jvm.gc.pause is replaced by OTel
+      //     process.runtime.jvm.gc.duration{gc!=<concurrent gcs>}
+      //   runtime.jvm.gc.max.data.size is replaced by OTel
+      //     process.runtime.jvm.memory.limit{pool=<long lived pools>}
+      //   runtime.jvm.gc.live.data.size is replaced by OTel
```
gsmirnov-splk, line 67: The `runtime.jvm.gc.pause` and `runtime.jvm.gc.live.data.size` metrics are used by Splunk Memory Profiling, we should not remove them from the agent until there is full server-side support for the new metrics.

PR #1009, opened 2022-11-22
Don't use ::set-output in GHA, it's deprecated

mateuszrzeszutek:
See: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


breedx-splk (approved):
Thanks, yeah, it'll be nice to be rid of those warnings.

PR #1029, opened 2022-12-14
Update to instrumentation and core 1.21.0.

breedx-splk:
In preparation of the release...


mateuszrzeszutek (approved):
Thanks!
I pushed a commit that fixes a couple of failing tests.

PR #1057, opened 2023-01-05
Truncate commandline when metrics enabled.

breedx-splk:
In order to work around ingest issues with commandline being larger than 255 chars, we now truncate when metrics are enabled. The workaround for users wanting a longer commandline is provided via config item, but enabling it will create ingest problems. Once ingest is fixed we can remove this workaround.


docs/advanced-config.md
```diff
@@ -13,15 +13,16 @@ Below you will find all the configuration options supported by this distribution
...
+|-----------------------------------------|-----------------------------------------|-------------------------|--------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `splunk.access.token`                   | `SPLUNK_ACCESS_TOKEN`                   | unset                   | Stable       | (Optional) Auth token allowing exporters to communicate directly with the Splunk cloud, passed as `X-SF-TOKEN` header. Currently, the [Jaeger trace exporter](#trace-exporters) as well as the [SignalFx metrics exporter](metrics.md) support this property. |
+| `splunk.realm`                          | `SPLUNK_REALM`                          | `none`                  | Stable       | The Splunk cloud realm where the telemetry should be sent to; e.g. `us0`, `us1`. Defaults to `none`, which means sending data to a Splunk OpenTelemetry Collector deployed on `localhost`.                                                                    |
```
theletterf, line 19:
```suggestion
| `splunk.realm`                          | `SPLUNK_REALM`                          | `none`                  | Stable       | The Splunk Observability Cloud realm where the telemetry should be sent to. For example, `us0` or `us1`. Defaults to `none`, which means that data goes to a Splunk OpenTelemetry Collector deployed on `localhost`.                                                                    |
```

docs/advanced-config.md
```diff
@@ -13,15 +13,16 @@ Below you will find all the configuration options supported by this distribution
...
+| `splunk.metrics.export.interval`        | `SPLUNK_METRICS_EXPORT_INTERVAL`        | `30s`                   | Experimental | The interval between pushing metrics.  <details><summary>Format</summary>Durations can be of the form `{number}{unit}`, where unit is one of `ms`, `s`, `m`, `h`, `d`. If no unit is specified, milliseconds is the assumed duration unit.</details>          |
+| `splunk.metrics.implementation`         | `SPLUNK_METRICS_IMPLEMENTATION`         | `micrometer`            | Experimental | The metrics implementation used by the agent. Valid values are `micrometer` and `opentelemetry`.                                                                                                                                                              |
+| `splunk.metrics.force_full_commandline` | `SPLUNK_METRICS_FORCE_FULL_COMMANDLINE` | `false`                 | Experimental | Forces the full commandline to be included as a resource attribute for all metrics. Default truncates to 255 chars.                                                                                                                                           |
```
theletterf, line 24:
```suggestion
| `splunk.metrics.force_full_commandline` | `SPLUNK_METRICS_FORCE_FULL_COMMAND_LINE` | `false`                 | Experimental | Adds the full command line as a resource attribute for all metrics. Maximum length is 255 characters. Commands longer than 255 characters are truncated.                                                                                                                                           |
```

@breedx-splk It's `COMMAND_LINE` according to the code, not `COMMANDLINE`, right? By the way, what do we refer to with that? The full JVM invocation?

theletterf (changes requested):

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
```diff
@@ -0,0 +1,68 @@
...
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+import java.util.logging.Logger;
+
+@AutoService(ResourceProvider.class)
+public class TruncateCommandLineWhenMetricsEnabled implements ConditionalResourceProvider {
+
+  private static final Logger logger =
+      Logger.getLogger(TruncateCommandLineWhenMetricsEnabled.class.getName());
+
+  private String commandLine;
```
mateuszrzeszutek, line 37: Keeping state this way seems kinda hacky. Instead of using the `ConditionalResourceProvider`, how about making it a `AutoConfigurationCustomizerProvider` and calling `AutoConfigurationCustomizer#addResourceCustomizer()`?

Also, using the resource customizer has one more advantage: it receives the final (well, almost final) version of the resource, that also includes attributes from the system props/env vars.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
```diff
@@ -41,6 +41,7 @@ public class SplunkConfiguration implements AutoConfigurationCustomizerProvider
   public static final String METRICS_ENDPOINT_PROPERTY = "splunk.metrics.endpoint";
   public static final String METRICS_EXPORT_INTERVAL_PROPERTY = "splunk.metrics.export.interval";
   public static final String METRICS_IMPLEMENTATION = "splunk.metrics.implementation";
+  public static final String METRICS_FULL_COMMAND_LINE = "splunk.metrics.force_full_commandline";
```
mateuszrzeszutek, line 44: Usually we use dashes instead of underscores in properties' names:
```suggestion
  public static final String METRICS_FULL_COMMAND_LINE = "splunk.metrics.force-full-commandline";
```

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public boolean shouldApply(ConfigProperties config, Resource existing) {
+    boolean metricsEnabled = config.getBoolean(METRICS_ENABLED_PROPERTY, false);
+    boolean forceFullCommandline = config.getBoolean(METRICS_FULL_COMMAND_LINE, false);
+    boolean wantTruncate = metricsEnabled && !forceFullCommandline;
+    if (wantTruncate) {
+      logger.warning(
+          "Warning: Metrics are enabled, so process.command_line resource attribute is being truncated.");
```
mateuszrzeszutek, line 46: WDYT about printing the warning only if the attribute is actually truncated?

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
```diff
@@ -0,0 +1,68 @@
...
+      commandLine = existing.getAttribute(ResourceAttributes.PROCESS_COMMAND_LINE);
```
mateuszrzeszutek, line 47: Just wondering: have you considered truncating all attributes, not just the command line? I mean, the command line is sure to cause problems, but maybe it'd be worth to check others too.

docs/advanced-config.md
breedx-splk: I admit, it's confusing. The code constant has `COMMAND_LINE` but the system property has `COMMANDLINE` because commandline is one word for now and forever in my brain and you won't convince me otherwise. 🙃  This markdown is intended to document to the user what the actual config item is, not the code constant.

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
breedx-splk: Agreed, I hate this. If the resource customizer will work, that sounds better. Lemme try it. THANKS!

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
breedx-splk: Will you allow me to break with tradition here in the name of making things better? I think the hyphens are actually problematic because they don't translate cleanly to environment vars. We've done a pretty good job of documenting both, but it definitely still opens up an opportunity for users to start with or copy the system property and manually convert it to env var by uppercasing and changing periods to underscores, which WILL NOT WORK if the original system property contains hyphens.

I dunno, I don't feel that strongly about it and can err on the side of consistency, even if it's worse. 😁

docs/advanced-config.md
theletterf: Haha, that's perfectly OK. Just wanted to be sure. :)

theletterf (approved):
Just nits, really. Approving so as not to block master Plumb!

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
breedx-splk: I didn't want to risk messing with identity or anything. I wanted to do the most surgical change to the one (resource) attribute that we know causes consistent problems. I can reason that the result is a truncated commandline, and I can reason that its utility is modest at best. I dunno, I just didn't want to make such a sweeping change....would prefer to treat other cases as they arise.

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
breedx-splk: Looks good, thanks!

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
```diff
@@ -0,0 +1,73 @@
...
+
+      String commandLine = existing.getAttribute(ResourceAttributes.PROCESS_COMMAND_LINE);
+      if (commandLine == null || commandLine.length() < 256) {
+        return existing;
+      }
+
+      logger.warning("Metrics are enabled. Truncating process.command_line resource attribute.");
+      String newCommandLine = commandLine.substring(0, 250) + "[...]";
+      return Resource.create(
+          Attributes.of(ResourceAttributes.PROCESS_COMMAND_LINE, newCommandLine));
```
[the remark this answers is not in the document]
breedx-splk: LOL wow whoops...that would have been really bad. Thanks.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
breedx-splk: What do you mean `handle underscores too`? If they're already underscores, what would need to change?

custom/src/test/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabledTest.java
```diff
@@ -0,0 +1,129 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class TruncateCommandLineWhenMetricsEnabledTest {
+
+  @Mock ConfigProperties config;
+  @Mock Resource existingResource;
```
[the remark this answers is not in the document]
breedx-splk: it's actually only used in one test, so lemme just make it a local.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
mateuszrzeszutek: Like, you could pass either `splunk.metrics.force-full-commandline` or `splunk.metrics.force_full_commandline` as a system property and both would work the same way

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.function.BiFunction;
+import java.util.logging.Logger;
+
+@AutoService(AutoConfigurationCustomizerProvider.class)
+public class TruncateCommandLineWhenMetricsEnabled implements AutoConfigurationCustomizerProvider {
+
+  private static final Logger logger =
+      Logger.getLogger(TruncateCommandLineWhenMetricsEnabled.class.getName());
+
+  private String commandLine;
```
mateuszrzeszutek, line 39: Field no longer used?
```suggestion
```

custom/src/test/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabledTest.java
```diff
@@ -0,0 +1,132 @@
...
+    var testClass = new TruncateCommandLineWhenMetricsEnabled.CommandLineTruncator();
+    var result = testClass.apply(existing, config);
+
+    assertSame(existing, result);
+    assertThat(result.getAttribute(PROCESS_COMMAND_LINE)).isEqualTo(cmd);
+  }
+
+  @Test
+  void shouldNotApplyIfConfigItemOverrides() {
+    var existing = Resource.create(Attributes.of(PROCESS_COMMAND_LINE, "blargus"));
```
mateuszrzeszutek, line 84: WDYT about creating a resource that contains some other attribute, in addition to the command line -- this would verify that we're not losing any of the other resource attributes, and cover the `existing.merge()` case.

breedx-splk: Yup, a good idea. Will do.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
breedx-splk: [Is that not true today](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ConfigPropertiesUtil.java#L32)? I thought it was basically just "dots and hyphens to underscores, then uppercase".

mateuszrzeszutek: Nope: https://github.com/open-telemetry/opentelemetry-java/blob/727850e94faa1edcb45c469620e7fa1f9baecf60/api/all/src/main/java/io/opentelemetry/api/internal/ConfigUtil.java#L61

`ConfigPropertiesUtil` is only used in very few cases where the SDK config is not available.

mateuszrzeszutek (approved):

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
breedx-splk: Ah, thanks for clarifying! That separation between `Config` and `ConfigProperties` always seemed questionable to me. :) I was thinking more of the interop between properties and env vars, and less about normalizing among system property names.

PR #1067, opened 2023-01-11
Rework gradle deps

breedx-splk:
Remove the spring dependency management plugin in favor of the gradle built-in `java-platform` approach. 

[Ref1](https://docs.gradle.org/current/userguide/platforms.html) and [ref2](https://docs.gradle.org/current/userguide/java_platform_plugin.html).

The spring plugin was not terribly compatible with dependabot and we were missing critical updates as they were released. This borrows from the approach in upstream, which seems much happier with dependabot poking around.

### DIFF
--- a/agent/build.gradle.kts
+++ b/agent/build.gradle.kts
@@ -29,9 +29,10 @@ val upstreamAgent: Configuration by configurations.creating {
   isCanBeConsumed = false
 }
 
-val otelInstrumentationVersion: String by extra
+val otelInstrumentationVersion: String by rootProject.extra
 
 dependencies {
+  add("upstreamAgent", platform(project(":dependencyManagement")))
   bootstrapLibs(project(":bootstrap"))
 
   javaagentLibs(project(":custom"))
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,7 +30,14 @@ group = "com.splunk"
 subprojects {
   version = rootProject.version
 
-  apply(plugin = "splunk.java-conventions")
-  apply(plugin = "splunk.spotless-conventions")
+  if (this.name != "dependencyManagement") {
+    apply(plugin = "splunk.java-conventions")
+    apply(plugin = "splunk.spotless-conventions")
+    dependencies {
+      for(conf in configurations) {
+        add(conf.name, platform(project(":dependencyManagement")))
+      }
+    }
+  }
 }
 
--- a/instrumentation/khttp/src/main/java/com/splunk/opentelemetry/instrumentation/khttp/KHttpSingletons.java
+++ b/instrumentation/khttp/src/main/java/com/splunk/opentelemetry/instrumentation/khttp/KHttpSingletons.java
@@ -45,8 +45,9 @@ public final class KHttpSingletons {
                 INSTRUMENTATION_NAME,
                 HttpSpanNameExtractor.create(httpAttributesGetter))
             .setSpanStatusExtractor(HttpSpanStatusExtractor.create(httpAttributesGetter))
-            .addAttributesExtractor(HttpClientAttributesExtractor.create(httpAttributesGetter))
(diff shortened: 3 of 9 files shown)


instrumentation/khttp/src/main/java/com/splunk/opentelemetry/instrumentation/khttp/KHttpSingletons.java
```diff
@@ -45,7 +45,8 @@ public final class KHttpSingletons {
                 INSTRUMENTATION_NAME,
                 HttpSpanNameExtractor.create(httpAttributesGetter))
             .setSpanStatusExtractor(HttpSpanStatusExtractor.create(httpAttributesGetter))
-            .addAttributesExtractor(HttpClientAttributesExtractor.create(httpAttributesGetter))
+            .addAttributesExtractor(
+                HttpClientAttributesExtractor.create(httpAttributesGetter, netAttributesGetter))
```
breedx-splk, line 49: This is unrelated, but I saw the warning on an otherwise clean build, so I threw it in.

agent/build.gradle.kts
```diff
@@ -29,9 +29,10 @@ val upstreamAgent: Configuration by configurations.creating {
   isCanBeConsumed = false
 }
 
-val otelInstrumentationVersion: String by extra
+val otelInstrumentationVersion: String by rootProject.extra
 
 dependencies {
+  add("upstreamAgent", platform(project(":dependencyManagement")))
```
mateuszrzeszutek, line 35: Does the loop in the main `build.gradle.kts` file not handle this case?

instrumentation/micrometer-1.3-shaded-for-instrumenting/build.gradle.kts
```diff
@@ -2,8 +2,19 @@ plugins {
...
 dependencies {
-  implementation("io.micrometer:micrometer-core:1.3.16")
+  implementation("io.micrometer:micrometer-core")
+}
+
+configurations.all {
+  resolutionStrategy.eachDependency {
+    if (requested.group == "io.micrometer" && requested.name == "micrometer-core") {
+      useVersion(micrometerOldVersion)
+      because("We need the old version")
```
mateuszrzeszutek, line 15: Nice 👍

build.gradle.kts
```diff
@@ -30,7 +30,14 @@ group = "com.splunk"
...
 
-  apply(plugin = "splunk.java-conventions")
-  apply(plugin = "splunk.spotless-conventions")
+  if (this.name != "dependencyManagement") {
+    apply(plugin = "splunk.java-conventions")
+    apply(plugin = "splunk.spotless-conventions")
+    dependencies {
+      for(conf in configurations) {
+        add(conf.name, platform(project(":dependencyManagement")))
+      }
```
mateuszrzeszutek, line 39: Nit: WDYT about putting it into the `java-conventions` file?

mateuszrzeszutek (approved):
Awesome! Finally dependabot will work properly 🤖

agent/build.gradle.kts
breedx-splk: Yeah this surprised me as well, but in practice, no, it did not handle it. I'm I assume it is caused by the fact that those configurations are created/defined local to the agent subproject build script? There are a couple other places that required this as well.

build.gradle.kts
breedx-splk: Oh that's a good idea I think. I do expect any small change to just break all of the gradle build completely, but let's try it...

breedx-splk: Looks like that worked out. Keeps some of the muck out of the top-level build script so I think I like it. 🤘🏻

agent/build.gradle.kts
mateuszrzeszutek: Gradle black magic I guess 🧙 💀

PR #1086, opened 2023-01-18
Post 1.20.0 release - update versions

breedx-splk:


dependencyManagement/build.gradle.kts
```diff
@@ -2,10 +2,10 @@ plugins {
   `java-platform`
 }
 
-val otelVersion = "1.22.0"
-val otelAlphaVersion = "1.22.0-alpha"
-val otelInstrumentationVersion = "1.22.1"
-val otelInstrumentationAlphaVersion = "1.22.1-alpha"
+val otelVersion = "1.23.0-SNAPSHOT"
+val otelAlphaVersion = "1.23.0-alpha-SNAPSHOT"
```
mateuszrzeszutek, line 6: We usually don't bump the SDK version to snapshots -- cause the upstream agent also doesn't do that
```suggestion
val otelVersion = "1.22.0"
val otelAlphaVersion = "1.22.0-alpha"
```

mateuszrzeszutek (approved):

Issue #647, opened 2022-01-27
Publish docker images to Splunk docker registry


iNikem:
Instead of (or in addition to) publishing docker images with java agent to `quay.io` we should publish them to `repo.splunkdev.com`

breedx-splk:
Do we think that would be preferred over GHCR?

breedx-splk:
We publish to GHCR, so this should be fine for public consumption.

PR #1108, opened 2023-02-10
Add workflow to publish to GHCR

breedx-splk:
This is related to #647.
We want to be publishing a more minimal docker layer that only contains our released agent jars. Also, we are switching from quay.io for open source images, so this sets up a GHA workflow that triggers on release creation and builds/publishes to GHCR.

There is a [sample repo here](https://github.com/breedx-splk/ghcr-publish-exp) that I used in development. You can look at the actions there and see the packages that result.

### DIFF
--- a/.github/workflows/publish-docker-image-ghcr.yml
+++ b/.github/workflows/publish-docker-image-ghcr.yml
@@ -29,7 +29,7 @@ jobs:
         uses: docker/build-push-action@v4.0.0
         with:
           push: true
-          file: Dockerfile
+          file: Dockerfile-ghcr
           build-args: |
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek (approved):
Awesome! 🚀

breedx-splk:
Pushed a change to the filename (I forgot that I called the new one `Dockerfile-ghcr` in order to leave the older one in place for a while).

Given that we have a release coming up soon, I think we should merge this to try it out with the next release. The old stuff should remain unchanged for now. We could possibly run into permissions problems (in my work-personal repo I had to enable a setting for it to not 403 on publish).

PR #1119, opened 2023-02-15
Remove usage of internal ConfigPropertiesUtil class

breedx-splk:
Not sure how we feel about using classes in the upstream `internal` packages, but I thought it maybe seemed like a bad idea. I found another way about this here....curious what we think.

There are usages of `EmbeddedInstrumentationProperties` in `ThreadPoolMetrics` but I didn't touch that because I'm not aware of a replacement...


mateuszrzeszutek (approved):

mateuszrzeszutek:
> There are usages of `EmbeddedInstrumentationProperties` in `ThreadPoolMetrics` but I didn't touch that because I'm not aware of a replacement...

There isn't any; but since we're going to remove that instrumentation in 2.x I guess it's fine if we leave it as it is for now.

PR #1135, opened 2023-02-23
Update versions to SNAPSHOTs on main branch

mateuszrzeszutek:


.gitlab-ci.yml
```diff
@@ -46,6 +46,7 @@ snapshot:
 release:
   stage: release
   artifacts:
+    when: always
```
mateuszrzeszutek, line 49: This line will save us in case the gitlab build fails because of random crap again

agoallikmaa (approved):

PR #1150, opened 2023-03-08
Build fix amalgamation and dep upgrades

breedx-splk:
* Updates to gradle 8+
* Update shadow jar to latest (supersedes #1140 #1143)
* Update protobuf (supersedes #1129)
* Update nexus plugin (supersedes #1149)

Resolves #1151

PR #1154, opened 2023-03-08
Upgrade to gradle 8.0.2

laurit:


breedx-splk:
Gradle 8 is part of #1150 fyi

PR #1155, opened 2023-03-08
Update Dockerfile-ghrc with changes to support auto-instrumentation

jvoravong:
Description:
Trying to get this image setup to work with auto-instrumentation. These are the last changes we need, I tested them manually. I didn't realize we needed to be able to run the command cp, so these latest changes are to fix this issue.

I was hoping we could use the busybox base image going forward. Let me know if there are any issues with this.


breedx-splk:
Hey @jvoravong. I guess I'm confused about this. We specifically went away from busybox to reduce the number of vulns that might be sitting in a this image at a given time, especially from system packages. The intent was to provide a thin layer that ONLY contains the java agent jar, and we added one symlink. At least I think that's what we talked about.

I'm not familiar enough with what you're building, but I think that rather than trying to run `cp` from a container created with this image, that the tooling should build a new image that has sourced the agent from this layer. 

Sorry, it's confusing. We can also chat about it.

PR #1192, opened 2023-04-04
Profiling data can't be sent directly to ingest

laurit:


profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -82,6 +85,17 @@ Map<String, String> defaultProperties() {
 
   public static String getConfigUrl(ConfigProperties config) {
     String ingestUrl = config.getString(CONFIG_KEY_OTEL_OTLP_URL, null);
+    if (ingestUrl != null
+        && ingestUrl.startsWith("https://ingest.")
+        && ingestUrl.endsWith(".signalfx.com")
```
breedx-splk, line 90: This will fail for the default [HEC endpoint](https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/upstream_agent_config.yaml#L11), right? 

`https://ingest.us0.signalfx.com/v2/trace` doesn't end with `.signalfx.com`.

laurit: this uses otlp endpoint not hec, it is set in
https://github.com/signalfx/splunk-otel-java/blob/7876c18b63f368d4e2e72aabaf515d628d0f2570/custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java#L167-L171

breedx-splk: Oh that's right, we don't HEC out of the java agent. Sorry.

breedx-splk (approved):

mateuszrzeszutek (approved):

PR #1194, opened 2023-04-05
Deprecate text format for profiling

breedx-splk:
We will remove it completely in 2 releases.


profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
```diff
@@ -39,6 +39,9 @@
 import jdk.jfr.consumer.RecordedThread;
 
 public class PlainTextAllocationEventExporter implements AllocationEventExporter {
+
+  private static final java.util.logging.Logger LOGGER =
```
laurit, line 43: nit: in this project we use lowercase `logger`
any reason for not importing `java.util.logging.Logger`?

theletterf (approved):
Seems like we don't even document those in our official docs.

laurit (approved):

mateuszrzeszutek (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/exporter/PlainTextAllocationEventExporter.java
breedx-splk: I made the var name lowercase for consistency. Thanks!

PR #1196, opened 2023-04-06
Add server timing header using response customizer

laurit:
Remove custom instrumentation that adds `Server-Timing` header and use `HttpServerResponseCustomizer` from upstream.


custom/src/main/java/com/splunk/opentelemetry/micrometer/SplunkMetricsConfig.java
```diff
@@ -56,7 +57,8 @@ class SplunkMetricsConfig implements SignalFxConfig {
   @Override
   public boolean enabled() {
     return config.getBoolean(METRICS_ENABLED_PROPERTY, false)
-        && "micrometer".equals(config.getString(METRICS_IMPLEMENTATION));
+        && "micrometer".equals(config.getString(METRICS_IMPLEMENTATION))
+        && !config.getBoolean(METRICS_MICROMETER_DISABLED, false);
```
breedx-splk, line 61: Help me understand the use case here. Why would we allow someone to enable metrics and have micrometer be the implementation but actually have micrometer disabled? I'm not following. 

Is this merely for the test case? I'm guessing so because the new setting didn't make it into the docs. Wondering if we should qualify it with internal or add a comment that it's for internal use only or something.

breedx-splk (approved):
LGTM, one small typo and a question about the new config item.

custom/src/main/java/com/splunk/opentelemetry/micrometer/SplunkMetricsConfig.java
mateuszrzeszutek: Nit: can we add `internal` or `testing` into the property name? E.g. `splunk.metrics.testing.micrometer.disabled` or sth similar

mateuszrzeszutek (approved):

PR #1208, opened 2023-04-12
Fix agent startup with security manager

laurit:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/7983 adds an experimental option to improve compatibility with security manager to upstream agent. This pr adds a test for running our agent with enabled security manger and moves calls to methods that require privileges, such as `System.setProperty` and `System.getenv`, to a later point where agent code has the required privileges.

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/logging/SplunkSlf4jLoggingCustomizer.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/logging/SplunkSlf4jLoggingCustomizer.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


custom/src/main/java/com/splunk/opentelemetry/logging/SplunkSlf4jLoggingCustomizer.java
```diff
@@ -0,0 +1,71 @@
...
+import io.opentelemetry.javaagent.logging.simple.Slf4jSimpleLoggingCustomizer;
+import io.opentelemetry.javaagent.tooling.LoggingCustomizer;
+
+@AutoService(LoggingCustomizer.class)
+public class SplunkSlf4jLoggingCustomizer implements LoggingCustomizer {
+  private final Slf4jSimpleLoggingCustomizer delegate = new Slf4jSimpleLoggingCustomizer();
+
+  @Override
+  public String name() {
+    return delegate.name();
```
mateuszrzeszutek, line 29: There are now 2 `LoggingCustomizer`s with the same name. Is it guaranteed that ours will always be the one that runs? Should we perhaps exclude the original one from the SPI file?

laurit: Currently all our services are listed before upstream services in the service files.
Excluding it seems like fine idea, but actually doing it will probably require an unhealthy amount of gradle magic. I think we would need build a transformer as described in https://imperceptiblethoughts.com/shadow/configuration/merging/#merging-service-descriptor-files WDYT is there a simpler way to do this? Should I attempt the transformer approach?

breedx-splk:
@laurit Can you please provide a brief description in this PR that explains the core problem and how this change improves it?

breedx-splk (approved):

PR #1229, opened 2023-04-19
Use jfr parser from jdk mission control

laurit:


profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/sampler/AllocationEventSampler.java
```diff
@@ -17,17 +17,14 @@
 package com.splunk.opentelemetry.profiler.allocation.sampler;
 
 import java.util.function.BiConsumer;
-import jdk.jfr.consumer.RecordedEvent;
 
 /** A sampler that decides whether given allocation event should be sampled or discarded. */
```
breedx-splk, line 21: With the api change here this comment isn't really accurate any longer, and the name of the class is probably inaccurate too. I'm not sure what to call it thought, maybe just `SamplerStrategy` or something?

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
```diff
@@ -40,15 +40,18 @@ public String serialize(RecordedStackTrace stack) {
...
     }
     maybeNewline(sb);
-    String className = method.getType().getName();
-    String methodName = method.getName();
-    int lineNumber = frame.getInt("lineNumber");
+    String className = method.getType().getFullName();
+    String methodName = method.getMethodName();
+    Integer lineNumber = frame.getFrameLineNumber();
+    if (lineNumber == null) {
+      lineNumber = -1;
```
breedx-splk, line 53: [GDI spec ](https://github.com/signalfx/gdi-specification/blob/dd07dbd736dad4fd1b1a677473284799df2ee50c/specification/semantic_conventions.md#call-stack-format-for-text-data-format)says 0 when unknown.

profiler/src/test/java/com/splunk/opentelemetry/profiler/EventProcessingChainTest.java
```diff
@@ -88,108 +89,25 @@ void testLifecycle() {
     inOrder.verifyNoMoreInteractions();
   }
 
-  @Test
-  void testFlushOnChunkChange() {
```
breedx-splk, line 92: I guess this is less interesting now that the chunk tracker exists elsewhere? Should we still verify that flush is called somewhere though?

breedx-splk (commented):
Looks good overall, had a few thoughts/questions sprinkled throughout. Nothing too interesting....

I have a few hangups on this that I think we should discuss.

1. I was a little surprised at the net _increase_ in LOC. I kinda would have expected a reduction, but ok, it's approximately the same. Is the addition of the jmc dependency and all its `I` prefixed classnames worth it? It's certainly less readable....
2. I know performance was likely to be a motivation behind this change, specifically reducing allocations which ultimately result in GC. Are there test numbers to back up this change as an improvement? Can those be shared?
3. Does going this route prevent us from adopting the upstream/collab `jfr-connection`? 
4. Are we ok with this divergence from the community collab code? Do we think this ultimately makes things more compatible or less compatible in the mid/long term?

agent/license-normalizer-bundle.json
```diff
@@ -4,24 +4,30 @@
...
       "bundleName": "apache2",
       "licenseNamePattern": "The Apache License, Version 2.0"
     },
     {
       "bundleName": "apache2",
       "licenseNamePattern": "The Apache Software License, Version 2.0"
+    },
+    {
+      "bundleName": "UPL-BSD-3-Clause",
+      // org.openjdk.jmc:flightrecorder gets detected as UPL, it is really UPL+BSD dual license"
```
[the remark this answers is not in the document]
laurit: Whatever parses this seems to be able to cope with such comments. Also tried adding comment as a field, but that failed because the parser tried to bind this to a class field that doesn't exist.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrPathHandler.java
```diff
@@ -54,12 +62,29 @@ public void accept(Path path) {
...
+        IChunkLoader chunkLoader = createChunkLoader(chunkSupplier, context, buffer, true);
+        if (chunkLoader == null) {
+          // we have parsed all chunks
+          break;
+        }
+        // update buffer to reuse it when parsing the next chunk
+        buffer = chunkLoader.call();
+        IItemCollection items = EventCollectionUtil.build(context.buildEventArrays());
+        items.stream().flatMap(iItems -> iItems.stream()).forEach(eventProcessingChain::accept);
+        eventProcessingChain.flushBuffer();
```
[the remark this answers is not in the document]
laurit: In previous parser chunks were hidden so we used our own chunk tracker that called `flushBuffer` on chunk change. The behavior didn't change, just the `flushBuffer` call moved.

profiler/src/test/java/com/splunk/opentelemetry/profiler/EventProcessingChainTest.java
laurit: I think this is covered with other tests, if `flushBuffer` is not called nothing will get exported with pprof.

profiler/src/main/java/com/splunk/opentelemetry/profiler/allocation/sampler/AllocationEventSampler.java
laurit: I think I'll add the event argument back

mateuszrzeszutek (commented):
Looks promising.

I'm a little bit worried about all the internal JMC classes that we use; but the JFR based implementation also depends on internal behavior (chunks) so it's not exactly a new thing.

profiler/src/main/java/com/splunk/opentelemetry/profiler/util/StackSerializer.java
laurit: Thanks, will change. Note that previous parser returned `-1` when line number was not available

laurit: Replaced all occurrences `-1`, please review.

mateuszrzeszutek (approved):

breedx-splk (approved):

PR #1237, opened 2023-04-28
Remove thread.os.id

laurit:
`thread.os.id` is an optional attribute, we can skip filling it


breedx-splk (approved):
Sure, we can always add it back if it becomes important.

PR #1241, opened 2023-05-01
After release, push a version update PR to the helm chart repo.

breedx-splk:
In order for the helm charts to get updated, we need to push a PR after each release.

I really wanted to use `yq` to do yaml editing instead of `sed`, but sadly both flavors I found and tried have outstanding issues around keeping existing comments and formatting. I didn't want a one-line version change to result in thousands of lines touched or important comments removed.


breedx-splk:
Converting to draft, we might be able to leverage dependabot instead...

breedx-splk:
Hopefully the dependabot approach will work out.

PR #1284, opened 2023-05-19
Helm chart PR after release

breedx-splk:
New version of #1241 

In order for the helm charts to get updated, we need to push a PR after each release.

I really wanted to use yq to do yaml editing instead of sed, but sadly both flavors I found and tried have outstanding issues around keeping existing comments and formatting. I didn't want a one-line version change to result in thousands of lines touched or important comments removed.

Ok, after talking about this with the collector helm chart team, the other approach doesn't work because it requires running this `make render` command to update the templates...and dependabot won't work with that. So we'll just hack it up.


mateuszrzeszutek (approved):

breedx-splk:
Hacktastic.

PR #1290, opened 2023-05-26
Add OTLP logs exporter smoke test

mateuszrzeszutek:


smoke-tests/src/test/java/com/splunk/opentelemetry/OtlpLogsSmokeTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class OtlpLogsSmokeTest extends SmokeTest {
+
+  @Override
+  protected Map<String, String> getExtraEnv() {
+    return Map.of("OTEL_LOGS_EXPORTER", "otlp");
+  }
+
+  @ParameterizedTest(name = "{index} => OTLP logs exporter test on JDK{0}.")
+  @ValueSource(ints = {8, 11, 15})
```
laurit, line 41: It would be better to test agains `17` instead of `15`, we should update the other smoke tests too

laurit (approved):

PR #1304, opened 2023-06-06
Profiler output dir is only used by the keep files setting

laurit:


profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -83,35 +81,42 @@ private boolean notClearForTakeoff(ConfigProperties config) {
       return true;
     }
 
-    Path outputDir = Paths.get(config.getString(CONFIG_KEY_PROFILER_DIRECTORY));
+    return false;
```
breedx-splk, line 84: Combine with `if` to make something like

`return JFR.instance.isAvailable()`

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -83,35 +81,42 @@ private boolean notClearForTakeoff(ConfigProperties config) {
...
+
+  private boolean checkOutputDir(Path outputDir) {
     if (!Files.exists(outputDir)) {
       // Try creating the directory for the user...
       try {
         Files.createDirectories(outputDir);
       } catch (IOException e) {
-        return outdirWarn(outputDir, "does not exist and could not be created");
+        outdirWarn(outputDir, "does not exist and could not be created");
+        return false;
```
breedx-splk, line 94: These used to return true, so the semantic is changed. Does the caller account for this? What happens if the user specifies `keepfiles=true` but the output path is not available or writable or whatever? Do we still profile, but just not output files? Is this worth documenting maybe?

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -83,35 +81,42 @@ private boolean notClearForTakeoff(ConfigProperties config) {
...
     if (!Files.isDirectory(outputDir)) {
-      return outdirWarn(outputDir, "exists but is not a directory");
+      outdirWarn(outputDir, "exists but is not a directory");
+      return false;
     }
 
     if (!Files.isWritable(outputDir)) {
-      return outdirWarn(outputDir, "exists but is not writable.");
+      outdirWarn(outputDir, "exists but is not writable");
+      return false;
```
breedx-splk, line 104: The return value is duplicated here across these cases, where it used to be encapsulated in the method. I guess you just didn't like that or thought it was hard to read?

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -83,35 +81,42 @@ private boolean notClearForTakeoff(ConfigProperties config) {
...
 
+  private void outdirWarn(Path dir, String suffix) {
+    logger.log(WARNING, "The configured output directory {0} {1}.", new Object[] {dir, suffix});
+  }
+
   private void activateJfrAndRunForever(ConfigProperties config, Resource resource) {
+    boolean keepFiles = keepFiles(config);
     Path outputDir = Paths.get(config.getString(CONFIG_KEY_PROFILER_DIRECTORY));
+    if (keepFiles && !checkOutputDir(outputDir)) {
+      keepFiles = false;
```
breedx-splk, line 118: Ah ok, I think this answers the question I asked above about the behavior. I think the user should be told clearly here that files will not be written.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
laurit: Yes, the caller expects failure to return `false`. If user specifies `keepfiles=true` and output path is not writable then profiling will be enabled but files will not be written. This deviates from the current behaviour of turning profiling off when output path is not writable regardless of the `keepfiles` setting. `keepfiles` setting is not really useful for end users, I would not expect anybody using it, If users want jfr recordings they should use other tools to produce them.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
laurit: I didn't like it.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
laurit: The warning about output directory not being writable should be enough to inform user why there are not recordings. Also consider that the `keepfiles` flag is only useful for debugging the agent, from the users perspective it does not make sense to use the `keepfiles` flag.

breedx-splk (approved):

PR #1338, opened 2023-06-22
Add an option to publish docker image manually

mateuszrzeszutek:


.github/workflows/publish-docker-image-ghcr.yml
```diff
@@ -3,6 +3,7 @@ name: publish splunk-otel-java docker image to ghcr
 on:
   release:
     types: [published]
+  workflow_dispatch:
```
breedx-splk, line 6: 👍🏻

PR #1339, opened 2023-06-22
When publishing the docker image, checkout all history

mateuszrzeszutek:


.github/workflows/publish-docker-image-ghcr.yml
```diff
@@ -13,6 +13,8 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v3.5.3
+        with:
+          fetch-depth: 0
```
breedx-splk, line 17: Wow, bummer that this is needed.

PR #1341, opened 2023-06-22
No longer push a PR to collector, it will poll

breedx-splk:
Based on[ this discussion](https://github.com/signalfx/splunk-otel-collector/pull/3304#issuecomment-1603226793), the collector team will be adding their own polling mechanism, so we are no longer on the hook for this.


jeffreyc-splunk (approved):
This will be handled within the collector repo: https://github.com/signalfx/splunk-otel-collector/pull/3305

mateuszrzeszutek (approved):

PR #1355, opened 2023-07-07
Update deps add reflection based accessor

breedx-splk:
Supersedes #1353 and #1354 

I thought that copying the internal class over to be our own internal class was maybe slightly better than actually using the external internal class. 🤯 Feel free to disagree with this and I'll change it.


laurit:
> Supersedes #1353 and #1354
> 
> I thought that copying the internal class over to be our own internal class was maybe slightly better than actually using the external internal class. 🤯 Feel free to disagree with this and I'll change it.

Using internal classes is ok. Internal just means that it can change without prior notice and that is fine.

breedx-splk:
> Using internal classes is ok. Internal just means that it can change without prior notice and that is fine.

I guess it depends on which approach you think is less vulgar. I copied the class because I thought it would feel less awful, but I haven't convinced myself.

@laurit @mateuszrzeszutek what's your preference? Copy the class or use the upstream internal?

PR #1426, opened 2023-09-07
Truncate process.command_args

laurit:
Similarly to `process.command_line` we need to limit the size of `process.command_args`

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
@@ -27,6 +27,9 @@ import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvide
 import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
 import java.util.function.BiFunction;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
```diff
@@ -57,15 +62,66 @@ public Resource apply(Resource existing, ConfigProperties config) {
...
+      }
+      return resource;
+    }
+
+    private static List<String> truncate(List<String> list) {
+      // when List is translated to String it is surrounded in [], each element is surrounded in
+      // double quotes, elements are separated with a comma
+      int maxLength = MAX_LENGTH - 4;
+      List<String> result = new ArrayList<>();
+      int size = 0;
```
mateuszrzeszutek, line 98: Nit: maybe rename it to `totalLength` or sth similar?

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
```diff
@@ -57,15 +62,66 @@ public Resource apply(Resource existing, ConfigProperties config) {
...
+      }
+
+      if (existing != resource) {
+        logger.warning(
+            "Metrics are enabled. Truncating process.command_line and process.command_args resource attributes.");
+      }
+      return resource;
+    }
+
+    private static List<String> truncate(List<String> list) {
```
mateuszrzeszutek, line 93: So what's the reason behind it? The length of all values summed must be <255 as well?

mateuszrzeszutek (approved):

custom/src/main/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabled.java
mateuszrzeszutek: Can you add a comment saying that the JSON encoding of the list must be <255 chars long?

laurit: done

mateuszrzeszutek (approved):

PR #1428, opened 2023-09-08
Add otel version of profiling metrics

laurit:
Add otel version of metrics used by profiling


mateuszrzeszutek (approved):

laurit:
> We still have a metrics reference doc inside `docs`. Will the PR update it?

as far as I can tell that doc describes micrometer metrics, this pr is about otel metrics

PR #1465, opened 2023-10-03
Remove dependency on signalfx-metrics

breedx-splk:
I didn't actually see any usages, so maybe this was a carryover from when `signalfx-java` was a monolith? Were we shading this or something? In any case, there seemed to be just the one transitive test dependency on StringUtils.


laurit:
this is needed you shouldn't remove it. we exclude signalfx-java that micrometer depends on and add signalfx-metrics to replace it.

breedx-splk:
I assumed that it must be required, but I didn't easily see how/why/where it was used. If you can help me understand that I'll just put it back with a comment for future selves to better understand.

PR #1470, opened 2023-10-10
Silence profiler warning on jdk21

laurit:
Silence a non-fatal warning when starting profiler with jdk21. Also fixes log level setting for `org.apache.http.impl.execchain.RetryExec`

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/logging/SplunkSlf4jLoggingCustomizer.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/logging/SplunkSlf4jLoggingCustomizer.java
@@ -47,18 +47,33 @@ public class SplunkSlf4jLoggingCustomizer implements LoggingCustomizer {
     delegate.onStartupFailure(throwable);
   }
 
-  private static final String METRICS_RETRY_LOGGER_PROPERTY =
-      "org.apache.http.impl.execchain.RetryExec";
+  private static final String LOGGER_PREFIX = "io.opentelemetry.javaagent.slf4j.simpleLogger.log.";
 
   private static void addCustomLoggingConfiguration() {
-    // metrics exporter sometimes logs "Broken pipe (Write failed)" at INFO; usually in
-    // docker-based environments
-    // most likely docker resets long-running connections and the exporter has to retry, and it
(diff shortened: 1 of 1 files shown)


custom/src/main/java/com/splunk/opentelemetry/logging/SplunkSlf4jLoggingCustomizer.java
```diff
@@ -47,18 +47,33 @@ public void onStartupFailure(Throwable throwable) {
...
+      // org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader - Could not
+      // find field with name 'virtual' in reader for 'thread'
+      setLogLevelIfNotSet(
+          "org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader",
+          "ERROR");
+    }
+  }
+
+  private static void setLogLevelIfNotSet(String className, String logLevel) {
+    String loggerProperty = LOGGER_PREFIX + className;
```
breedx-splk, line 73: The system properties are really like this?
* `io.opentelemetry.javaagent.slf4j.simpleLogger.log.org.apache.http.impl.execchain.RetryExec`
* `io.opentelemetry.javaagent.slf4j.simpleLogger.log.org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader`

Previously we were checking to see if `org.apache.http.impl.execchain.RetryExec` was set, and not the concatenated version. If we check the concatenated version, is this a behavior change?

laurit: Using only `org.apache.http.impl.execchain.RetryExec` without the logger prefix was a bug.

mateuszrzeszutek (approved):

custom/src/main/java/com/splunk/opentelemetry/logging/SplunkSlf4jLoggingCustomizer.java
breedx-splk: Ah, cool. Thanks for the clarification. 😎

PR #1496, opened 2023-10-20
Pin servlet api version for smoke test app

laurit:


mateuszrzeszutek (approved):
+1 to pinning the version

breedx-splk:
I think I agree with this, but realistically, how often does the servlet api actually change anyway? 😁

PR #1508, opened 2023-10-31
fix: Check mxBean compatibility

asuresh4:
Check for mxBean compatibility ahead of checking whether Allocated Memory is enabled. This is related to #1435 .

### DIFF
--- a/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
+++ b/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/AllocatedMemoryMetrics.java
@@ -36,7 +36,7 @@ public class AllocatedMemoryMetrics {
   }
 
   private AllocationTracker createAllocationTracker() {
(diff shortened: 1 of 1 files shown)


breedx-splk (approved):

laurit:
@asuresh4 this repository requires commits to be signed https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits you also need to add the gpg key to your github account. Also take a look at https://safjan.com/git-sign-previous-commit/

asuresh4:
Thanks @laurit , just amended the commit.

asuresh4:
@breedx-splk @laurit thanks for reviewing and merging this. Would it be possible to do a patch release including the fix? We have a customer waiting for this. If that is not a possibility when is the next planned release? Thanks again!

laurit:
@asuresh4 I think the next planned release would be on 16 nov. Patch release can be done if needed.

asuresh4:
@laurit I opened #1515  for updating CHANGELOG. PTAL. It would be awesome to provide a patch release for the customer since they've wanted to install the agent and this exception has been blocking them.

laurit:
@asuresh4 is this fixing the same issue as described in https://github.com/signalfx/splunk-otel-java/pull/1435?

PR #1527, opened 2023-11-20
Update micrometer and jackson and rebuild licenses

breedx-splk:
This should supersede #1522 and #1519


laurit:
Let's keep the jackson version same as in the upstream for the release

PR #1551, opened 2023-12-04
Group spotless updates

laurit:


breedx-splk (approved):
Thanks! I'm going to steal this for the Android repos as well.

PR #1564, opened 2024-01-02
Document allocation profiler event rate limiting flag

laurit:


breedx-splk (approved):

breedx-splk:
@theletterf in case this isn't yet reflected in docs too.

PR #1567, opened 2024-01-04
Skip publishing splunk-otel-javaagent-all.jar

laurit:


breedx-splk (approved):
Weird, I thought we got rid of or stopped publishing that a long time ago...but maybe I'm thinking of upstream.

breedx-splk:
Actually, we might have some internal uses that rely on that name. Let's chat offline.

laurit:
> Weird, I thought we got rid of or stopped publishing that a long time ago...but maybe I'm thinking of upstream.

Yes upstream removed it a long time ago, we deprecated it in https://github.com/signalfx/splunk-otel-java/releases/tag/v1.8.0

breedx-splk:
Cool, I think the other mitigations we have in place now should be enough. I'm ok with merging this. Thanks!

PR #1568, opened 2024-01-04
Generate metadata yaml

laurit:
metadata yaml can be downloaded from https://github.com/signalfx/splunk-otel-java/actions/runs/7411564164/artifacts/1147926232


.github/workflows/pr.yaml
```diff
@@ -196,3 +196,29 @@ jobs:
             echo "$(git diff --cached licenses)"
             exit 1
           fi
+
+  generate-metadata:
```
laurit, line 200: I mainly added this so that the generated yaml could be reviewed. @breedx-splk wdyt do we want to have the metadata yaml attached to pr/ci builds? Or just add it to releases?

breedx-splk: Is the intent for the yaml to live in the repo alongside the code, or to be published, or something else? On release seems fine, especially if the target consumer is a docs generator of some future design. I don't think docs will necessarily branch/rev like normal repos.

breedx-splk:
This seems fine, thanks for doing all this cumbersome leg work. Did you just pull it all together by hand? Is this something that we should document in `CONTRIBUTING.md` as part of the regular process of keeping docs up to date? 

The intent is for us to just maintain this by hand, right? If so, is it better than just maintaining the yaml file itself? I mean, I appreciate that the version and deps come from the build itself, but the rest of it just seems like a bunch of hard coded values hiding in code. 🤷🏻

.github/workflows/pr.yaml
laurit: ok, I'll remove this after @theletterf has had a chance to review the generated yaml

laurit:
> This seems fine, thanks for doing all this cumbersome leg work. Did you just pull it all together by hand? Is this something that we should document in `CONTRIBUTING.md` as part of the regular process of keeping docs up to date?

This was created by hand. It may be possible to automate checking whether the source files have changed to make keeping it up to date easier.

> The intent is for us to just maintain this by hand, right? If so, is it better than just maintaining the yaml file itself? I mean, I appreciate that the version and deps come from the build itself, but the rest of it just seems like a bunch of hard coded values hiding in code. 🤷🏻

As the structure of this yaml may still change generating it with code will make such changes easier.

I'm a bit skeptical about the usefulness of this and wouldn't spend time on improving it before it is shown that this yaml can actually be used.

.github/workflows/pr.yaml
theletterf: FWIW, .NET is releasing the YAML as a release artifact.

theletterf (changes requested):
Fantastic work, @laurit. There are only a few things that need addressing:

- There seem to be a lack of indentation for certain items. For example, items inside each first level object are not indented. Not sure if this would pose a problem to YAML parsers, but I'd much rather have indentation for readability and browsing purposes.

Screenshot 2024-01-09 at 11 58 31

- The same indentation issues appears in nested items, such as the dependencies within the resource detectors.

- To preserve the same structure used by .NET, please rename the `name` key in `settings` to `env`.

With this, I think we'd already have an useful artifact for testing.

laurit:
@theletterf see the update yaml https://github.com/signalfx/splunk-otel-java/actions/runs/7462323350/artifacts/1156871225

theletterf (approved):
Perfect. Thanks SO much! :)

PR #1569, opened 2024-01-05
Use http/protobuf in smoke tests

laurit:
Resolves https://github.com/signalfx/splunk-otel-java/issues/1570
Resolves https://github.com/signalfx/splunk-otel-java/issues/1571
Resolves https://github.com/signalfx/splunk-otel-java/issues/1572
Resolves https://github.com/signalfx/splunk-otel-java/issues/1574

### DIFF
--- a/smoke-tests/src/test/java/com/splunk/opentelemetry/OtlpLogsSmokeTest.java
+++ b/smoke-tests/src/test/java/com/splunk/opentelemetry/OtlpLogsSmokeTest.java
@@ -37,7 +37,11 @@ public class OtlpLogsSmokeTest extends SmokeTest {
 
   @Override
   protected Map<String, String> getExtraEnv() {
-    return Map.of("OTEL_LOGS_EXPORTER", "otlp");
+    return Map.of(
+        "OTEL_LOGS_EXPORTER",
+        "otlp",
+        "OTEL_INSTRUMENTATION_COMMON_EXPERIMENTAL_CONTROLLER_TELEMETRY_ENABLED",
+        "true");
   }
 
   @ParameterizedTest(name = "{index} => OTLP logs exporter test on JDK{0}.")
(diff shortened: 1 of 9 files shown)


smoke-tests/src/test/java/com/splunk/opentelemetry/OtlpLogsSmokeTest.java
```diff
@@ -37,7 +37,11 @@ public class OtlpLogsSmokeTest extends SmokeTest {
 
   @Override
   protected Map<String, String> getExtraEnv() {
-    return Map.of("OTEL_LOGS_EXPORTER", "otlp");
+    return Map.of(
+        "OTEL_LOGS_EXPORTER",
+        "otlp",
+        "OTEL_INSTRUMENTATION_COMMON_EXPERIMENTAL_CONTROLLER_TELEMETRY_ENABLED",
```
breedx-splk, line 43: Unclear to me why we need this on a logs smoke test. Is it necessary?

breedx-splk (approved):
Do you think we want to leave the experimental controller stuff turned on long-term? I'm cool getting the nightly to pass again, just want us to consider if that's what we actually want going forward.

I guess I'm just thinking that it may be overkill for a "smoke" test to cover an experimental feature.

smoke-tests/src/test/java/com/splunk/opentelemetry/OtlpLogsSmokeTest.java
laurit: It is needed because the test application is a spring boot application that generates extra spans when controller telemetry is enabled. This test searches for a span named `WebController.greeting` that is created only when controller telemetry is enabled.

laurit:
> Do you think we want to leave the experimental controller stuff turned on long-term? I'm cool getting the nightly to pass again, just want us to consider if that's what we actually want going forward.

I think this is yet to be decided. As the otel agent is going to disable these by default we might do it as well.

> I guess I'm just thinking that it may be overkill for a "smoke" test to cover an experimental feature.

We can change the tests later on to not depend on these.

PR #1576, opened 2024-01-09
Remove micrometer

laurit:
also enable metrics by default


agoallikmaa (approved):
Any reason to keep around the `TestMetrics` and `TestMicrometerInstaller` in a tmp directory?

laurit:
> Any reason to keep around the `TestMetrics` and `TestMicrometerInstaller` in a tmp directory?

Thanks, removed.

breedx-splk:
Woah nice! Thanks!

PR #1586, opened 2024-01-15
Update upstream version to 2.0.0

laurit:
Resolves https://github.com/signalfx/splunk-otel-java/issues/1588


theletterf (approved):

theletterf:
@laurit Wonder if this PR could contain a change for the README where we point folks to both the 2.0.0 version and the last 1.x version and its docs.

laurit:
@theletterf we can do this either in this pr or a separate pr. Could you point me to where you want to make the change in the README and what you wish to change.

agoallikmaa (approved):

PR #1589, opened 2024-01-16
Remove remaining references to micrometer

laurit:
Also restores `splunk.metrics.enabled` that will control whether splunk custom metrics are emitted.


laurit:
> It would be nice to have the the `"splunk.metrics.enabled"` use in `MetricsInstrumentationModule` use the new `SplunkConfiguration.METRICS_ENABLED_PROPERTY` constant too, but that would make `common` depend on `custom`, which I'm not sure if you want to do.

I removed `splunk.metrics.enabled` along with the micrometer and now restored it as it was previously.

docs/advanced-config.md
```diff
@@ -13,25 +13,13 @@ Below you will find all the configuration options supported by this distribution
...
-| `splunk.trace-response-header.enabled`  | `SPLUNK_TRACE_RESPONSE_HEADER_ENABLED`  | `true`                  | Stable       | Enables adding server trace information to HTTP response headers. See [this document](server-trace-info.md) for more information.                                                                                                                    |
-
-The SignalFx exporter can be configured to export metrics directly to Splunk ingest. To achieve that, you need to set
-the `splunk.access.token` configuration property and set the `splunk.metrics.endpoint` to Splunk ingest URL. For
-example:
-
-```bash
-export SPLUNK_ACCESS_TOKEN=my_splunk_token
-export SPLUNK_METRICS_ENDPOINT=https://ingest.us0.signalfx.com
-```
```
breedx-splk, line 34: YASS! 🥳

docs/metrics.md
```diff
@@ -6,188 +6,46 @@
 > configuration properties may change.
 
 The Splunk Distribution of OpenTelemetry Java agent gathers basic application metrics.
-[Micrometer](https://micrometer.io/)
-and [Micrometer SignalFx registry](https://micrometer.io/docs/registry/signalFx)
-gather and export metrics to either [SignalFx SmartAgent](https://github.com/signalfx/signalfx-agent/)
-or the [Splunk distribution of OpenTelemetry Collector](https://github.com/signalfx/splunk-otel-collector).
-
-Because this feature is still experimental, metrics are not enabled by default. 
+Because these metrics are still experimental they are not enabled by default.
```
breedx-splk, line 9:
```suggestion
Because these splunk-specific metrics are still experimental they are not enabled by default.
```

docs/metrics.md
```diff
@@ -6,188 +6,46 @@
...
 
 | Library/Framework                                                    | Instrumentation name   | Versions |
 | -------------------------------------------------------------------- |------------------------| -------- |
 | [JVM metrics](#jvm)                                                  | `jvm-metrics-splunk`   | [Java runtimes version 8 and higher](../README.md#requirements)
-| [Apache DBCP2 connection pool metrics](#connection-pool-metrics)     | `commons-dbcp2-splunk` | 2.0 and higher
-| [c3p0 connection pool metrics](#connection-pool-metrics)             | `c3p0-splunk`          | 0.9.5 and higher
-| [HikariCP connection pool metrics](#connection-pool-metrics)         | `hikaricp-splunk`      | 3.0 and higher
-| [Oracle Universal Connection Pool metrics](#connection-pool-metrics) | `oracle-ucp-splunk`    | 11.2.0.4 and higher
-| [Tomcat JDBC connection pool metrics](#connection-pool-metrics)      | `tomcat-jdbc-splunk`   | 8.5 and higher
-| [Vibur DBCP connection pool metrics](#connection-pool-metrics)       | `vibur-dbcp-splunk`    | 20.0 and higher
```
breedx-splk, line 95: These are all in upstream then, eh?

docs/metrics.md
```diff
@@ -6,188 +6,46 @@
...
+Besides OpenTelemetry JVM Metrics we include the following JVM metrics.
 
-| Metric name                         | Instrument   | Description |
-|-------------------------------------| ------------ | ----------- |
-| `runtime.jvm.gc.overhead`           | [Gauge][g]   | An approximation of the percentage of CPU time used by GC activities over the last lookback period or since monitoring began, whichever is shorter, in the range [0..1].
-| `runtime.jvm.memory.usage.after.gc` | [Gauge][g]   | The percentage of long-lived heap pool used after the last GC event, in the range [0..1].
+#### Memory profiler metrics
 
-#### Memory metrics
+These metrics are enabled only when memory profiler is enabled.
```
breedx-splk, line 33:
```suggestion
These metrics are only enabled when the memory profiler is enabled or when setting `splunk.metrics.enabled=true`.
```

docs/metrics.md
```diff
@@ -6,188 +6,46 @@
...
-| Tag name | Tag value |
-| -------- | --------- |
-| `area`   | Either `heap` or `nonheap`.
-| `id`     | Name of the memory pool, e.g. `Perm Gen`.
+| Metric name                            | Instrument   | Description                                                      |
+|----------------------------------------|--------------|------------------------------------------------------------------|
+| `process.runtime.jvm.memory.allocated` | [Counter][c] | Approximate sum of heap allocations.                             |
+| `process.runtime.jvm.memory.reclaimed` | [Counter][c] | Sum of heap size differences before and after gc.                |
+| `runtime.jvm.gc.pause.count`           | [Counter][c] | Number of gc pauses.                                             |
+| `runtime.jvm.gc.pause.totalTime`       | [Counter][c] | NTime spent in GC pause.                                         |
```
breedx-splk, line 40: Will readers here understand that these are monotonically increasing? Does it matter?

instrumentation/jvm-metrics/build.gradle.kts
```diff
@@ -7,38 +7,7 @@ dependencies {
...
-    dependsOn(testOtelJvmMetrics)
-  }
-
-  withType<Test>().configureEach {
-    jvmArgs("-Dsplunk.metrics.enabled=true")
-    jvmArgs("-Dsplunk.metrics.experimental.enabled=true")
-  }
+tasks.withType<Test>().configureEach {
+  jvmArgs("-Dsplunk.metrics.enabled=true")
+  jvmArgs("-Dsplunk.metrics.experimental.enabled=true")
```
breedx-splk, line 12: `splunk.metrics.experimental.enabled` isn't mentioned in `metrics.md` yet? I didn't see it. Do we need to include that?

custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java
```diff
@@ -36,6 +36,7 @@ public class SplunkConfiguration implements AutoConfigurationCustomizerProvider
   public static final String SPLUNK_REALM_PROPERTY = "splunk.realm";
   public static final String SPLUNK_REALM_NONE = "none";
 
+  public static final String METRICS_ENABLED_PROPERTY = "splunk.metrics.enabled";
```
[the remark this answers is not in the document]
laurit: This pr restored the constant as it was before I removed it along with micrometer code. If you so wish you can rename it in a follow up pr.

custom/src/test/java/com/splunk/opentelemetry/SplunkConfigurationTest.java
```diff
@@ -115,40 +114,6 @@ private static ConfigProperties configuration(
...
-            "commons-dbcp2-splunk",
-            "hikari-splunk",
-            "micrometer-splunk",
-            "oracle-ucp-splunk",
-            "tomcat-jdbc-splunk",
-            "vibur-dbcp-splunk")) {
-      assertInstrumentationDisabled(config, instrumentationName);
-    }
-  }
-
```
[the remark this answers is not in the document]
laurit: Not really. The meaning of `splunk.metrics.enabled` changed from don't enable metrics to don't enable splunk specific metrics. Previously we disabled a bunch of otel metrics that we had our own replacements for, similarly when otel metrics were enabled we had to disable our own versions. Now otel metrics are always enabled and our own versions were deleted.

docs/metrics.md
laurit: correct.

docs/metrics.md
laurit: Setting `splunk.metrics.enabled=true` is not enough to enable all of them. See https://github.com/signalfx/splunk-otel-java/blob/3099b48ee6c890495022098a4c6c31b9dea369ae/instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/JvmMetricsInstaller.java#L59-L66 there is another flag that is needed to enable most of them (as far as I remember we use that flag for testing).

instrumentation/jvm-metrics/build.gradle.kts
laurit: I think the only purpose of this flag is to enable testing these metrics. I think we don't need to document this unless there is someone that wishes to use these metrics without having memory profiling enabled. As far as I understand these metrics are there only to be used in memory profiling charts. I guess if we wanted we could delete this flag and use the memory profiling flag here instead.

PR #1593, opened 2024-01-19
Remove gc live data size metric

laurit:
Not needed according to @gsmirnov-splk


docs/metrics.md
```diff
@@ -38,7 +38,6 @@ These metrics are enabled only when memory profiler is enabled.
 | `process.runtime.jvm.memory.reclaimed` | [Counter][c] | Sum of heap size differences before and after gc.                |
 | `runtime.jvm.gc.pause.count`           | [Counter][c] | Number of gc pauses.                                             |
 | `runtime.jvm.gc.pause.totalTime`       | [Counter][c] | Time spent in GC pause.                                          |
-| `runtime.jvm.gc.live.data.size`        | [Gauge][g]   | Size of long-lived heap memory pool after reclamation, in bytes. |
```
theletterf, line 41: What about `runtime.jvm.gc.max.data.size`?

gsmirnov-splk: Profiling was not using it and not planning to, not sure about others.

laurit: `runtime.jvm.gc.max.data.size` is not present any more

agoallikmaa (approved):

PR #1603, opened 2024-01-29
Remove thread state metric

laurit:
@akubik-splunk proposed that `runtime.jvm.threads.states` metric can be removed


breedx-splk (approved):

theletterf:
Removed in docs draft.

PR #1605, opened 2024-01-30
Don't enable thread pool metrics with profiler

laurit:
As discussed with @akubik-splunk in preparation for the potential removal of the thread pool metrics I'm going to make memory profiling enable only the profiling metrics and leave the thread pool metrics disabled.


theletterf (commented):
No changes to docs required, as we don't currently state that thread pool metrics are enabled when enabling memory profiling.

@laurit How would one enable thread pool metrics manually anyway after this PR gets merged?

laurit:
@theletterf either with the `splunk.metrics.enabled` flag (currently thread pool metrics are the only thing enabled by this) or by enabling the instrumentations that provide these metrics with `otel.instrumentation.[instrumentation name].enabled` flag.

agoallikmaa (approved):

PR #1610, opened 2024-02-01
2.x transform schema

breedx-splk:
Here's a first-pass attempt at defining the transform rules for 2.x migration.

The idea is to be able to take the new/stable otel semconv names (left) and transform them to the old micrometer names (right). In some cases, there are also attributes that get renamed.

This format should be very similar to the upstream schema files produced by semconv.

### DIFF
--- a/migration/transformation_schema_2_to_1.yaml
+++ b/migration/transformation_schema_2_to_1.yaml
@@ -15,10 +15,10 @@ versions:
         #--------------------------------------------------------------
         # Memory metrics
         - rename_metrics:
-            jvm.memory.used:               runtime.jvm.memory.used
-            jvm.memory.committed:          runtime.jvm.memory.committed
-            jvm.memory.limit:              runtime.jvm.memory.max
-            jvm.memory.used_after_last_gc: runtime.jvm.memory.usage.after.gc
+            jvm.memory.used:               runtime.jvm.memory.used           # micrometer
+            jvm.memory.committed:          runtime.jvm.memory.committed      # micrometer
+            jvm.memory.limit:              runtime.jvm.memory.max            # micrometer
+            jvm.memory.used_after_last_gc: runtime.jvm.memory.usage.after.gc # micrometer
 #            jvm.memory.init:               process.runtime.jvm.memory.init  # experimental
         - rename_attributes:
             attribute_map:
@@ -29,23 +29,23 @@ versions:
               - jvm.memory.committed
               - jvm.memory.limit
               - jvm.memory.used_after_last_gc
-              - jvm.memory.init
+#              - jvm.memory.init
 
         #--------------------------------------------------------------
         # Garbage collection metrics
-        - rename_metrics:
-            jvm.gc.duration: process.runtime.jvm.gc.duration
-        - rename_attributes:
-            attribute_map:
-              name: jvm.gc.name
-              action: jvm.gc.action
-            apply_to_metrics:
-              - jvm.gc.duration
+#        - rename_metrics:
+#            jvm.gc.duration: process.runtime.jvm.gc.duration # otel
+#        - rename_attributes:
+#            attribute_map:
+#              name: jvm.gc.name
+#              action: jvm.gc.action
+#            apply_to_metrics:
+#              - jvm.gc.duration
 
         #--------------------------------------------------------------
         # Threads metrics
         - rename_metrics:
-            jvm.thread.count: runtime.jvm.threads.live
+            jvm.thread.count: runtime.jvm.threads.live # micrometer
         - rename_attributes:
             attribute_map:
               jvm.thread.daemon: daemon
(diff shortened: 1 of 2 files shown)


migration/transformation_schema_2_to_1.yaml
```diff
@@ -0,0 +1,134 @@
...
+            attribute_map:
+              jvm.thread.daemon: daemon
+            apply_to_metrics:
+              - jvm.threads.count
+
+        #--------------------------------------------------------------
+        # Classes metrics
+        - rename_metrics:
+#            jvm.class.loaded:   process.runtime.jvm.classes.loaded
+            jvm.class.unloaded: runtime.jvm.classes.unloaded  # micrometer
```
laurit, line 48: this one should also have the aggregation temporality mismatch, how would you express that?

migration/transformation_schema_2_to_1.yaml
```diff
@@ -0,0 +1,134 @@
...
+  2.0.0:
+    metrics:
+      changes:
+        #--------------------------------------------------------------
+        # Memory metrics
+        - rename_metrics:
+            jvm.memory.used:               runtime.jvm.memory.used           # micrometer
+            jvm.memory.committed:          runtime.jvm.memory.committed      # micrometer
+            jvm.memory.limit:              runtime.jvm.memory.max            # micrometer
+            jvm.memory.used_after_last_gc: runtime.jvm.memory.usage.after.gc # micrometer
```
theletterf, line 21: We're stating in the docs draft that there's no equivalent for this one. It is, in fact, red in the latest spreadsheet put together by Mateusz. Can we double check? Has anything changed?

migration/transformation_schema_2_to_1.yaml
```diff
@@ -0,0 +1,134 @@
...
+#            jvm.memory.init:               process.runtime.jvm.memory.init  # experimental
```
theletterf, line 22: Why is this commented out?

migration/transformation_schema_2_to_1.yaml
```diff
@@ -0,0 +1,134 @@
...
+              - jvm.memory.used
+              - jvm.memory.committed
+              - jvm.memory.limit
+              - jvm.memory.used_after_last_gc
+#              - jvm.memory.init
+
+        #--------------------------------------------------------------
+        # Threads metrics
+        - rename_metrics:
+            jvm.thread.count: runtime.jvm.threads.live # micrometer
```
theletterf, line 37: According to our source, `runtime.jvm.threads.daemon` also renames to `jvm.thread.count`.

migration/transformation_schema_2_to_1.yaml
```diff
@@ -0,0 +1,134 @@
...
+        - rename_metrics:
+            db.client.connections.usage:            db.pool.connections                 # micrometer
+            db.client.connections.idle.min:         db.pool.connections.idle.min        # micrometer
+            db.client.connections.idle.max:         db.pool.connections.idle.max        # micrometer
+            db.client.connections.max:              db.pool.connections.max             # micrometer
+            db.client.connections.pending_requests: db.pool.connections.pending_threads # micrometer
+            db.client.connections.timeouts:         db.pool.connections.timeouts        # micrometer
+#           db.client.connections.create_time:      db.pool.connections.create_time     # micrometer, timer vs. histogram
+#           db.client.connections.wait_time:        db.pool.connections.wait_time       # micrometer, timer vs. histogram
+#           db.client.connections.use_time:         db.pool.connections.use_time        # micrometer, timer vs. histogram
```
theletterf, line 95: Why are these commented out?

theletterf (changes requested):
PTAL at https://github.com/splunk/private-o11y-docs/pull/1589/files to check that the metrics equivalences there are accurate and whether we should have a separate table for attributes.

migration/transformation_schema_2_to_1.yaml
breedx-splk: There is no way in this format that I'm aware of to express a change in temporality. Maybe there is, but I also couldn't find the spec for the format and was just riffing on what exists today across 24 versions.

migration/transformation_schema_2_to_1.yaml
breedx-splk: Well, the semantic convention might have solidified since then, I'm not sure. The one listed above is the comparable one, however the difference that you're probably pointing out is that the micrometer one is apparently a _percentage_ of total memory, whereas the otel one is in bytes.

So we have a choice -- do we map the micromter name to the new name so that people can use the old one in calculations, or do we just drop it from mappings so that the new units can be used with the new name?

migration/transformation_schema_2_to_1.yaml
breedx-splk: This, along with the other commented out lines, are the otel (non-micrometer) breaking semconv changes. I had stubbed those out first in case we also wanted to help migrate those with transforms.  I don't think this was in the micrometer metrics.

migration/transformation_schema_2_to_1.yaml
breedx-splk: They were gray in the spreadsheet and don't have semantic equivalents afaik...largely due to the way micrometer timers are handled.

migration/transformation_schema_2_to_1.yaml
breedx-splk: If it's a problem we can always remove it.

migration/transformation_schema_2_to_1.yaml
theletterf: If folks keep using an old name but with different units, wouldn't they get a nasty surprise in their charts? It really depends on the principle we want to follow here. If we want to map everything at all costs, let's keep it as it is.

Would love to hear from @laurit, too.

migration/transformation_schema_2_to_1.yaml
theletterf: @laurit @breedx-splk Can you confirm? Could you reflect that in the mapping if that's the case?

breedx-splk: I think it's a problem if we try and map multiple metrics to the same metric.

theletterf: What should we do with `threads.daemon` then?

breedx-splk: * [Here](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java#L64) is the definition for `jvm.threads.daemon`
* [Here](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java#L70) is the definition for `jvm.threads.live`. 
* [Here](https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/#metric-jvmthreadcount) is the definition for otel `jvm.thread.count`. 

So yeah, it's not exactly the same thing, and I'm not convinced there's a clean mapping.

migration/transformation_schema_2_to_1.yaml
```diff
@@ -0,0 +1,134 @@
...
+        - rename_metrics:
+            jvm.memory.used:               runtime.jvm.memory.used           # micrometer
+            jvm.memory.committed:          runtime.jvm.memory.committed      # micrometer
+            jvm.memory.limit:              runtime.jvm.memory.max            # micrometer
+            jvm.memory.used_after_last_gc: runtime.jvm.memory.usage.after.gc # micrometer
+#            jvm.memory.init:               process.runtime.jvm.memory.init  # experimental
+        - rename_attributes:
+            attribute_map:
+              jvm.memory.type: area
+              jvm.memory.pool.name: pool
```
breedx-splk, line 26:
```suggestion
              jvm.memory.pool.name: id
```

breedx-splk:
We are going a different route for this. 👋🏻

PR #1618, opened 2024-02-06
Add metadata for sdk metrics

laurit:


metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -3265,7 +3296,8 @@ Some instrumentations (e.g. Spring Web MVC instrumentation) produce SpanKind.Int
             .metric(
                 "jvm.gc.pause.totalTime",
                 MetricInstrument.COUNTER,
-                "Time spent in GC pause. This metric will be removed in a future release."));
+                "Time spent in GC pause. This metric will be removed in a future release.")
+            .build());
```
laurit, line 3300: I think we should build the metadata for pr builds even when we are not going to publish the metadata file as build artifact to avoid bugs like this that completely break the metadata generation.

Kielek (approved):

agoallikmaa (approved):

PR #1622, opened 2024-02-08
Enable renovate on v1.30.x branch

laurit:
@breedx-splk wdyt?


breedx-splk (approved):
Yeah this is a great idea, thanks. 🤘🏻

PR #1625, opened 2024-02-08
Prepare release 2.0.0-alpha

laurit:


scripts/common.sh
```diff
@@ -58,12 +58,12 @@ get_minor_version() {
...
 get_patch_version() {
   local release_tag="$1"
-  get_release_version "$release_tag" | awk -F'.' '{print $3}'
+  get_release_version "$release_tag" | awk -F'[.-]' '{print $3}'
 }
 
 validate_version() {
   local version="$1"
-  if [[ ! $version =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
+  if [[ ! $version =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.*)?$ ]]
```
laurit, line 66: The `-alpha` qualifier is scary, I'm afraid that it'll break something in the release script.

agoallikmaa (approved):

scripts/common.sh
breedx-splk: > How so?

Because the previous version numbers forever didn't have the suffix, and there's some expectation in the conventions like `1.2.3` is different from `2.0.0-alpha` because it contains alphabetic chars.

We absolutely need those changes, and a few more.....

breedx-splk (approved):
🚀

PR #1628, opened 2024-02-09
Exclude compile only dependencies from fossa analysis

laurit:
See VULN-9916 for discussion, the list of configurations was also provided in that ticket. The intent is to exclude dependencies that are not bundled in the agent jar from the analysis.

### DIFF
--- a/.fossa.yml
+++ b/.fossa.yml
@@ -1,0 +1,10 @@
+version: 3
+
+experimental:
+  gradle:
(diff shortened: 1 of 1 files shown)


agoallikmaa (approved):

breedx-splk:
I want to go on record as saying that I think this is a drag. The scanning is done outside of this repo, and the configuration should live there as well IMO. We don't own fossa, we don't operate it, we shouldn't have to manage the configuration.

PR #1652, opened 2024-02-13
v2.0.0-alpha post release changes

srv-gh-o11y-gdi:
v2.0.0-alpha post release changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,11 @@ see [Migrate from the SignalFx Java Agent](https://quickdraw.splunk.com/redirect
 
 <!-- Comments, spacing, empty and new lines in the section below are intentional, please do not modify them! -->
 <!--DEV_DOCS_WARNING-->
+<!--DEV_DOCS_WARNING_START-->
+The following documentation refers to the in-development version of `splunk-otel-java`. Docs for the latest version ([v2.0.0-alpha](https://github.com/signalfx/splunk-otel-java/releases/latest)) can be found [here](https://github.com/signalfx/splunk-otel-java/blob/v2.0.0-alpha/README.md).
+
+---
+<!--DEV_DOCS_WARNING_END-->
 
 ## Requirements
 
--- a/deployments/cloudfoundry/index.yml
+++ b/deployments/cloudfoundry/index.yml
@@ -18,4 +18,3 @@
 1.28.0: https://github.com/signalfx/splunk-otel-java/releases/download/v1.28.0/splunk-otel-javaagent.jar
 1.29.0: https://github.com/signalfx/splunk-otel-java/releases/download/v1.29.0/splunk-otel-javaagent.jar
 1.30.0: https://github.com/signalfx/splunk-otel-java/releases/download/v1.30.0/splunk-otel-javaagent.jar
-2.0.0-alpha: https://github.com/signalfx/splunk-otel-java/releases/download/v2.0.0-alpha/splunk-otel-javaagent.jar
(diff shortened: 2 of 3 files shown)


version.gradle.kts
```diff
@@ -1,5 +1,5 @@
 // do NOT update that variable manually - it is managed by the pre/post release scripts
-val distroVersion = "2.0.0-alpha"
+val distroVersion = "2.1.0-SNAPSHOT"
```
breedx-splk, line 2:
```suggestion
val distroVersion = "2.0.1-SNAPSHOT"
```

README.md
```diff
@@ -73,6 +73,11 @@ see [Migrate from the SignalFx Java Agent](https://quickdraw.splunk.com/redirect
 
 <!-- Comments, spacing, empty and new lines in the section below are intentional, please do not modify them! -->
 <!--DEV_DOCS_WARNING-->
+<!--DEV_DOCS_WARNING_START-->
+The following documentation refers to the in-development version of `splunk-otel-java`. Docs for the latest version ([v2.0.0-alpha](https://github.com/signalfx/splunk-otel-java/releases/latest)) can be found [here](https://github.com/signalfx/splunk-otel-java/blob/v2.0.0-alpha/README.md).
```
breedx-splk, line 77:
```suggestion
The following documentation refers to the in-development version of `splunk-otel-java`. Docs for the latest version ([v1.30.1](https://github.com/signalfx/splunk-otel-java/releases/latest)) can be found [here](https://github.com/signalfx/splunk-otel-java/blob/v1.30.1/README.md).
```

deployments/cloudfoundry/index.yml
```diff
@@ -18,3 +18,4 @@
 1.28.0: https://github.com/signalfx/splunk-otel-java/releases/download/v1.28.0/splunk-otel-javaagent.jar
 1.29.0: https://github.com/signalfx/splunk-otel-java/releases/download/v1.29.0/splunk-otel-javaagent.jar
 1.30.0: https://github.com/signalfx/splunk-otel-java/releases/download/v1.30.0/splunk-otel-javaagent.jar
+2.0.0-alpha: https://github.com/signalfx/splunk-otel-java/releases/download/v2.0.0-alpha/splunk-otel-javaagent.jar
```
breedx-splk, line 21:
```suggestion
```

Prefer not to mess with CF while in alpha phase here.

version.gradle.kts
laurit: I think it should be `alpha-SNAPSHOT`. Going with `2.1.0` would let us be in sync with upstream, do we want that?

README.md
laurit: I'd leave this as it is, the 1.x branch will point to 1.x docs. Though if you think this is needed then we can do it this way too.

breedx-splk: Nah I think you're right. Let main be main and 2.0 even tho deemed alpha. Thanks for talking me down.

version.gradle.kts
breedx-splk: I mean, I guess that's fine. The problem has been that people then expect them to be in lockstep and when they diverge they freak out. :)

deployments/cloudfoundry/index.yml
laurit: the 2.0.0-alpha version is also used in https://github.com/signalfx/splunk-otel-java/blob/main/deployments/cloudfoundry/buildpack/README.md#configuration might be easier to let users decide which version they want

version.gradle.kts
laurit: imo, if possible, we should keep them in lockstep so that upstream major.minor.0 release would correspond to our major.minor.0, patch releases may differ.

deployments/cloudfoundry/index.yml
breedx-splk: Bah ok, I think you're right. I honestly am not sure if cloudfoundary (or at least the way our buildpack works) is compatible with multiple branches/major releases, but it does probably make sense to keep all the versions on main. I'll add it back in another PR that also adds the 1.30.1 patch.

breedx-splk (approved):

laurit (approved):

PR #1671, opened 2024-02-23
Remove python dependency for CloudFoundry supply

breedx-splk:
I could not find a reasonable way to use python in the supply script, nor could I find any references that describe how to install python or any existing art that shows other buildpacks using python in `supply`.....so I think it's just easier to rewrite the thing in `bash` and eliminate the dependency entirely.

Also seems to be more consistent with other buildpacks.

### DIFF
--- a/deployments/cloudfoundry/buildpack/bin/supply
+++ b/deployments/cloudfoundry/buildpack/bin/supply
@@ -1,79 +1,86 @@
-#!/usr/bin/env python3
+#!/bin/bash
 
-import os
-import shutil
-from os import makedirs, environ as ENV
-from pathlib import Path
-from sys import argv as ARGV
-from urllib.request import build_opener, HTTPRedirectHandler
+LATEST_VERSION="latest"
+JAVA_AGENT_VERSION="$SPLUNK_OTEL_JAVA_VERSION"
+if [ "$JAVA_AGENT_VERSION" == "" ]; then
+  JAVA_AGENT_VERSION="$LATEST_VERSION"
+fi
+JAVA_AGENT_JAR_NAME="splunk-otel-javaagent-${JAVA_AGENT_VERSION}.jar"
+JAVA_AGENT_URL="https://github.com/signalfx/splunk-otel-java/releases/download/v${JAVA_AGENT_VERSION}/splunk-otel-javaagent.jar"
+JAVA_AGENT_LATEST_URL="https://github.com/signalfx/splunk-otel-java/releases/latest/download/splunk-otel-javaagent.jar"
+RET=""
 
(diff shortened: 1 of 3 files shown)


deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -1,79 +1,86 @@
...
+  local AGENT_JAR="${CACHE_DIR}/${JAVA_AGENT_JAR_NAME}"
+  disable_cache_for_latest_version $AGENT_JAR
+  if [ !  -f "$AGENT_JAR" ]; then
+    echo "Agent jar ${AGENT_JAR} does not exist, downloading ..."
+    local URL="${JAVA_AGENT_URL}"
+    if [ "$JAVA_AGENT_VERSION" == "$LATEST_VERSION" ] ; then
+      URL="$JAVA_AGENT_LATEST_URL"
+    fi
+    echo "$URL"
+    curl -L -o "${AGENT_JAR}" "$URL"
```
agoallikmaa, line 36: I assume we can be sure that `curl` is installed on whatever this runs in?

agoallikmaa (approved):

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -1,79 +1,86 @@
-#!/usr/bin/env python3
+#!/bin/bash
```
laurit, line 1: would it make sense to use `/bin/sh` instead of `/bin/bash`? Perhaps not worth the effort.

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -1,79 +1,86 @@
...
 
-import os
-import shutil
-from os import makedirs, environ as ENV
-from pathlib import Path
-from sys import argv as ARGV
-from urllib.request import build_opener, HTTPRedirectHandler
+LATEST_VERSION="latest"
+JAVA_AGENT_VERSION="$SPLUNK_OTEL_JAVA_VERSION"
+if [ "$JAVA_AGENT_VERSION" == "" ]; then
```
laurit, line 5: I think it is more common to use something like `if [ "x$JAVA_AGENT_VERSION = "x" ]; then` or `if [ -z "$JAVA_AGENT_VERSION" ]; then` Idk, might be a bash vs sh thing.

laurit (approved):

deployments/cloudfoundry/buildpack/bin/supply
breedx-splk: Possibly, but when what happens if/when the default shell isn't `bash`? Seems risky doesn't it?

PR #1676, opened 2024-02-26
Update upstream version to snapshot

laurit:


laurit:
@breedx-splk maybe comment driven automation that would let you update licenses on the pr like we used to have in the upstream?

PR #1772, opened 2024-04-08
Switch renovate to v1.32.x branch

laurit:


breedx-splk (approved):
Leaving open so you can coordinate the merge/release foo.

PR #1763, opened 2024-04-05
Remove khttp instrumentation

breedx-splk:
Resolves https://github.com/signalfx/splunk-otel-java/issues/1769
Resolves https://github.com/signalfx/splunk-otel-java/issues/1768
Resolves https://github.com/signalfx/splunk-otel-java/issues/1767
Resolves https://github.com/signalfx/splunk-otel-java/issues/1764
Resolves https://github.com/signalfx/splunk-otel-java/issues/1762
Resolves https://github.com/signalfx/splunk-otel-java/issues/1759

Welp, it looks like khttp has attempted to disappear completely...so if that's the case, let's stop providing instrumentation for it. The thing was only at 0.1.0 and hadn't seen a release in ~6 years.

Evidence: 
* our failing build: https://github.com/signalfx/splunk-otel-java/actions/runs/8549750100/job/23465401587#step:5:43
* the author's github account is 404: https://github.com/jkcclemens
* jar artifact link from mvnrepository: https://repository.mulesoft.org/nexus/content/repositories/public/com/github/jkcclemens/khttp/0.1.0/khttp-0.1.0.jar

I wish I knew the full story....maybe someone who does can comment.

PR #1776, opened 2024-04-08
Use a fork of khttp

laurit:


breedx-splk (approved):
Yeah I think this is good. We can continue supporting this until things need to change. I'll look at reverting #1763 and going this route instead.

PR #1777, opened 2024-04-08
Restore khttp

breedx-splk:
This reverts #1763 and switches to using the semi-maintained fork of khttp, just like #1776.


laurit (approved):

laurit:
@theletterf we added back khttp instrumentation

PR #1785, opened 2024-04-09
Replace gradle enterprise with develocity

laurit:


breedx-splk:
Sorry, I think I missed the motivation behind this. I saw some stuff in upstream maybe changing around this, but wasn't sure if/how it might impact our distro. What's up?

laurit:
> Sorry, I think I missed the motivation behind this. I saw some stuff in upstream maybe changing around this, but wasn't sure if/how it might impact our distro. What's up?

gradle enterprise is now called develocity, gradle enterprise plugin is deprecated

breedx-splk:
[Got it](https://plugins.gradle.org/plugin/com.gradle.enterprise). Thanks!

breedx-splk (approved):

PR #1804, opened 2024-04-16
Update the use of gradle wrapper validation to the current form.

breedx-splk:
The release notes call this out:

https://github.com/gradle/wrapper-validation-action/releases/tag/v3.3.0

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -14,4 +14,4 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.2
 
-      - uses: gradle/wrapper-validation-action@v3.3.0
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -14,4 +14,4 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.2
 
-      - uses: gradle/wrapper-validation-action@v3.3.0
\ No newline at end of file
+      - uses: gradle/actions/wrapper-validation@v3
```
laurit, line 17: I think gdi spec requires using `v3.3.0` instead of `v3`

breedx-splk: ok I'll update

laurit (approved):

PR #1812, opened 2024-04-19
Remove dep on signalfx-metrics

breedx-splk:
Remove dependency on `signalfx-metrics`. Is this actually needed any more in the 1.x line?


laurit:
`signalfx-metrics` is used to send metrics from micrometer to splunk, it is requires in 1.x and can not be removed.

PR #1813, opened 2024-04-19
Remove commons lang

breedx-splk:
Looks like this dep was only used in 1 test and was just a silly string join that is easily accomplished in an alternative fashion.


custom/src/test/java/com/splunk/opentelemetry/TruncateCommandLineWhenMetricsEnabledTest.java
```diff
@@ -91,7 +91,7 @@ void truncatesWhenTooLong() {
     assertThat(resultCmd.length()).isEqualTo(255);
     assertThat(resultCmd.endsWith("...")).isTrue();
     var cmdArgs = result.getAttribute(PROCESS_COMMAND_ARGS);
-    String joinedArgs = StringUtils.joinWith(", ", cmdArgs.toArray());
+    String joinedArgs = getJoinedArgs(cmdArgs);
```
laurit, line 94: I'd try `String joinedArgs = String.join(", ", cmdArgs.toArray(new String[0]));`

laurit (approved):

PR #1817, opened 2024-04-19
Update commons-codec version

laurit:


breedx-splk:
When I build this locally I still see `commons-codec` 1.11 in the agent jar tho.

breedx-splk:
This was due to local license artifacts not being removed during a `clean`. Seems like this will work, let's go with it.

breedx-splk (approved):

PR #1858, opened 2024-05-21
Prepare release 1.32.1

laurit:
signalfx-java version should also be updated once it is released

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,20 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Unreleased
 
+## v1.32.1 - 2024-05-22
+
+### General
+
+- OpenTelemetry Java SDK has been updated to version 1.38.0.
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -8,12 +8,20 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Unreleased
 
+## v1.32.1 - 2024-05-22
+
```
breedx-splk, line 12: Since I think this is the version we're going to drop `-alpha` on, should we call this out as breaking from 1.x?

breedx-splk (approved):

CHANGELOG.md
laurit: I don't think we need to do that. For 1.x releases this is just a regular patch release as we have been doing for a while. Even for the next 2.x where we are going to drop `-alpha` we aren't really going to do any breaking changes compared to the previous 2.x release. We could add a notice that refers to the breaking changes in 2.0.

PR #1928, opened 2024-07-03
Warn on potentially wrong endpoint configuration

laurit:
Warn when the port number in OTLP endpoint doesn't agree with configured protocol. Hopefully this will help users more easily diagnose issues caused by changing the default protocol to `http/protobuf`.

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/EndpointProtocolValidator.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/EndpointProtocolValidator.java
@@ -35,9 +35,7 @@ class EndpointProtocolValidator {
     this.warn = warn;
   }
 
-  /**
-   * Warn when the port number in OTLP endpoint doesn't agree with configured protocol.
-   */
+  /** Warn when the port number in OTLP endpoint doesn't agree with configured protocol. */
   static void validate(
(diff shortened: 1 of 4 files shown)


custom/src/main/java/com/splunk/opentelemetry/EndpointProtocolValidator.java
```diff
@@ -0,0 +1,149 @@
...
+    for (String signal : new String[] {"traces", "metrics", "logs"}) {
+      verifySignalEndpointProtocol(signal, protocol, endpoint);
+    }
+  }
+
+  private boolean verifyEndpointProtocol(
+      String protocol, String endpoint, String protocolKey, String endpointKey) {
+    int port = getEndpointPort(endpoint);
+    if ("http/protobuf".equals(protocol)) {
+      if (port == 4317) {
```
agoallikmaa, line 67: Perhaps `String.valueOf(port).equals(getPort("grpc"))` and same for `"http/protobuf"` in the next block? Some extra work, but would only have one instance of 4318 and 4317 literals in this file then, plus it is clearer what this checks for when you don't remember which port number was for which protocol.

agoallikmaa (approved):

breedx-splk (approved):
This is awesome, I love it! Thank you. ❤️

PR #2007, opened 2024-08-16
Add windows to docker image platforms

breedx-splk:
@atoulme requested this. Seems fine.

### DIFF
--- a/.github/workflows/publish-docker-image-ghcr.yml
+++ b/.github/workflows/publish-docker-image-ghcr.yml
@@ -41,12 +41,12 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: build and publish container
+      - name: build and publish image
         uses: docker/build-push-action@v6.7.0
         with:
           push: true
           file: Dockerfile-ghcr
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker-image-ghcr.yml
```diff
@@ -41,12 +41,12 @@ jobs:
...
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: build and publish container
+      - name: build and publish image
         uses: docker/build-push-action@v6.7.0
         with:
           push: true
           file: Dockerfile-ghcr
-          platforms: linux/amd64,linux/arm64,linux/ppc64le
+          platforms: linux/amd64,linux/arm64,linux/ppc64le,windows/amd64
```
laurit, line 49: does it actually work? As far as I can tell the base image used in https://github.com/signalfx/splunk-otel-java/blob/82043a5e67c54f7d7fa6d688c22a615fe6668c92/Dockerfile-ghcr#L7 does not support windows.

breedx-splk: @atoulme need your opinion here thanks.

breedx-splk:
Yeah thanks @laurit this won't work as intended and we'll have to take another approach to support windows docker images. Thanks for the assist.

PR #2051, opened 2024-09-19
Let opentelemetry instrumentation bom configure the semconv version

laurit:
Currently because we set the semconv version we get renovate updates for it that we don't want (can't update semconv anyway before upstream updates).


breedx-splk (approved):
Oh good catch. 👍🏻

PR #2062, opened 2024-09-26
Update shadow plugin

laurit:


settings.gradle.kts
```diff
@@ -3,18 +3,7 @@ pluginManagement {
...
-}
-
-// We are using the following instead of having id("com.github.johnrengelman.shadow") version "8.1.1"
-// in pluginManagement to make shadow plugin use a newer version of asm so that it wouldn't fail on
-// classes compiled for jdk 21 (jackson-core 2.16.1 has such classes). This can be removed when
-// shadow plugin is updated.
-buildscript {
-  dependencies {
-    classpath("com.github.johnrengelman:shadow:8.1.1")
-    classpath("org.ow2.asm:asm:9.7")
```
robsunday, line 16: If package org.ow2.asm has been removed from dependencies then most probably it should also be removed from renovate.json5 file

breedx-splk (approved):

robsunday (commented):
It is fine

PR #2166, opened 2025-01-22
Prepare release 2.12.0

laurit:


dependencyManagement/build.gradle.kts
```diff
@@ -2,11 +2,11 @@ plugins {
...
 }
 
-val otelVersion = "1.45.0"
+val otelVersion = "1.46.0"
 val otelAlphaVersion = otelVersion.replaceFirst("(-SNAPSHOT)?$".toRegex(), "-alpha$1")
-val otelInstrumentationVersion = "2.11.0"
+val otelInstrumentationVersion = "2.12.0"
 val otelInstrumentationAlphaVersion =  otelInstrumentationVersion.replaceFirst("(-SNAPSHOT)?$".toRegex(), "-alpha$1")
-val otelContribAlphaVersion = "1.42.0-alpha"
+val otelContribAlphaVersion = "1.43.0-alpha"
```
robsunday, line 9: Should this also be mentioned in README.md?

laurit: We only list the sdk and upstream instrumentation version in the readme. An additional complication is that this version does not necessarily represent the version of all bundled contrib components. It only applies to components that are included by splunk agent, components included by the upstream agent may have an earlier version.

PR #2181, opened 2025-02-04
Register Traces for Snapshot Profiling

tduncan:
Adds a `SpanProcessor` with logic for identifying root spans and registering them for snapshot profiling. This is change #2 in a sequence of changes for porting the snapshot profiling feature in this repository. An additional PR will immediately follow to add trace thread profiling (along with yet more PRs for other necessary capabilities)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
tduncan, line 1: This `SpanProcessor` will expand in complexity a fair bit in the near future. Ultimately it acts as the coordination layer between distributed systems by deciding if a trace should be profiled (the root span logic in this PR), whether a down stream service should profile an already started trace (future change), and finally requesting the start of the profiling process.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
tduncan, line 1: Temporary activation notification code removed. Snapshot profiling extension activation can now be inferred by asking whether a trace has been registered for profiling with the `TraceRegistry` instance.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceRegistry.java
tduncan, line 1: The `TraceRegistry` is a store of traces that have been selected for profiling. Later changes will reference the state held in `TraceRegistry` instances for profiling callstack trace ID and span ID correlation as well profiling signal propagation, but for now it is more or less a facade over a `Map`.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/OpenTelemetrySdkExtension.java
```diff
@@ -16,28 +16,32 @@
...
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolver;
 
-public class OpenTelemetrySdkExtension implements AfterEachCallback {
+public class OpenTelemetrySdkExtension implements AfterEachCallback, ParameterResolver {
```
tduncan, line 44: The OpenTelemetry SDK extension has been updated to support providing `Tracer` parameters to individual tests. This is used to create and start spans to exercise the newly added `SpanProcessor` is a manner as similar as possible to how the OpenTelemetry runtime.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/OpenTelemetrySdkExtension.java
```diff
@@ -72,36 +89,52 @@ public Builder with(AutoConfigurationCustomizerProvider provider) {
...
      * Simplified re-implementation of AutoConfiguredOpenTelemetrySdkBuilder's build method. The
      * OpenTelemetry SDK is only configured with features necessary to pass existing test use cases.
      */
     public OpenTelemetrySdkExtension build() {
-      overrideProperties();
-
-      OpenTelemetrySdkBuilder sdkBuilder = OpenTelemetrySdk.builder();
-      OpenTelemetrySdk sdk = sdkBuilder.build();
+      ConfigProperties configProperties = customizeProperties();
+      SdkTracerProvider tracerProvider = customizeTracerProvider(configProperties);
```
tduncan, line 103: `SpanProcessors` are added to the OpenTelemetry SDK via `SdkTracerProvider`s so adding support for that SDK customization into the JUnit extension.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingFeatureFlagTest.java
tduncan, line 1: Renamed and slightly reworked the original SnapshotProfilingSdkCustomizerTest. While these tests look very similar to the other newly added tests their intent is to only consider whether the feature flag is honored.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpanSamplingTest.java
tduncan, line 1: While `SpanProcessor`s are not invoked by the OpenTelemetry runtime when a span is not sampled I felt that knowledge was a bit too far on the implementation detail side of things that it made sense to formally state the intent that the snapshot profiling SHOULD NOT activate when a span is not sampled despite not requiring explicit code changes to prevent that from happening.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  public void onStart(Context context, ReadWriteSpan span) {
+    if (isRoot(span) && isEntry(span)) {
+      registry.register(span.getSpanContext());
+    }
+  }
+
+  private boolean isRoot(ReadableSpan span) {
+    return SpanContext.getInvalid().equals(span.getParentSpanContext());
```
laurit, line 41: I'd use `span.getParentSpanContext().isValid()`

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  public boolean isStartRequired() {
+    return true;
+  }
+
+  @Override
+  public void onEnd(ReadableSpan span) {
+    if (isEntry(span)) {
+      registry.unregister(span.getSpanContext());
```
laurit, line 52: This method will not be called if there is a bug in the instrumentation that causes spans to be not ended which could result in a leak.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/OpenTelemetrySdkExtension.java
```diff
@@ -53,6 +57,18 @@ public void afterEach(ExtensionContext extensionContext) {
...
+  @Override
+  public boolean supportsParameter(
+      ParameterContext parameterContext, ExtensionContext extensionContext) {
+    return parameterContext.getParameter().getType() == Tracer.class;
+  }
+
+  @Override
+  public Object resolveParameter(
+      ParameterContext parameterContext, ExtensionContext extensionContext) {
+    return sdk.getTracer(extensionContext.getRequiredTestClass().getName(), "test'");
```
laurit, line 69: `"test'` looks like a typo

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
tduncan: At the moment the risk sounds very low given that we're not advertising this to customers yet, the extension is off by default, and it requires a bug to happen.

I'll follow this change up by replacing the simple `Map` in `TraceRegistry` with a more full featured cache. Maybe a `LoadingCache` from [`Caffeine`](https://github.com/ben-manes/caffeine)? Or do we have a preferred caching library already available?

laurit: I think it is fine to ignore this for now, if you so wish you can add a comment. 

> Maybe a LoadingCache from [Caffeine](https://github.com/ben-manes/caffeine)?

For this particular issue I'd guess a plain java WeakReference + RefereceQueue could work.

tduncan: Added a comment for now. Will address potential memory leak in a followup.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
tduncan: I forgot to negate the condition (is invalid vs. is valid). Updated to use `isValid`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
```diff
@@ -17,40 +17,42 @@
...
   @Override
   public void customize(AutoConfigurationCustomizer autoConfigurationCustomizer) {
-    autoConfigurationCustomizer.addPropertiesCustomizer(
-        config -> {
-          if (snapshotProfilingEnabled(config)) {
-            activationNotifier.run();
-          }
-          return emptyMap();
-        });
+    autoConfigurationCustomizer.addTracerProviderCustomizer(
```
breedx-splk, line 44: I think my preference here would be to check if the snapshot feature is enabled, and if so, then not even touch the `autoConfigurationCustomizer`. I think it's marginally safer but is also easier to understand.

breedx-splk (approved):
This looks good to me. Again, I deeply appreciate the care being given to submitting these in bite-sized reviewable pieces! 

I did have one small comment about when to check the config, and I also noticed that the test assertions aren't using the assertj style (preferred)...but since we're already using mixed assertions everywhere, it's probably not a big deal.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
tduncan: @breedx-splk I agree, though I'm unaware of how to access the `ConfigProperties` directly in the `customize` method.

tduncan:
> This looks good to me. Again, I deeply appreciate the care being given to submitting these in bite-sized reviewable pieces!
> 
> I did have one small comment about when to check the config, and I also noticed that the test assertions aren't using the assertj style (preferred)...but since we're already using mixed assertions everywhere, it's probably not a big deal.

Upcoming PRs will require new tests and modifications to the existing tests. I'll keep in mind the preference for `assertj` and update the existing assertions at the same time.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
laurit: See https://github.com/signalfx/splunk-otel-java/blob/0ccdc8339b803b72c69704098b69ca0aeda609d0/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java#L63 for the hack we use here and in the upstream instrumentation repro for getting access to `ConfigProperties`

tduncan: @laurit, unfortunately the `AgentListener` callback API appears accepts what appears to be a fully configured SDK where as `AutoConfigurationCustomizerProvider` callbacks are given a `AutoConfigurationCustomizer` which is more or less a builder of the eventual SDK.

So while the `AutoConfigurationCustomizer` implementation has a `ConfigProperties` instance field it's not directly accessible (there is a `private` `getConfig` method, but I don't think reflective access is the way forward) and, more importantly, the SDK is in an unreliable state of configuration when the `customize` method is called.

I don't think we can rely necessarily rely on the `ConfigProperties` being ready for use at the time this code executes.

laurit (approved):

PR #2184, opened 2025-02-13
Nocode prototype

johnbley:
This adds the ability to define
`SPLUNK_OTEL_INSTRUMENTATION_NOCODE_YML_FILE=/path/to/some.yml`
Where the yml looks like
```
- class: foo.Foo
  method: foo
  spanName: this.getName()
  attributes:
    - key: "business.context"
      value: this.getDetails().get("context")

- class: foo.Foo
  method: throwSomething
  spanKind: SERVER
  attributes:
    - key: "tostring"
      value: "this.toString()"
```
And then applies instrumentation to the given methods to create a span around them, with the given name/kind/attributes.  

(If this environment variable is not defined, there is no impact/change to the agent in any way).

It borrows heavily from the existing ["methods"](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/c22bb598e506bff8ad2dc518f4f73a3ca031d0fb/instrumentation/methods) instrumentation in upstream and could in theory be upstreamed as such, perhaps in reaction to requests like [wanting to name the span](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/11046).

It wound up being fairly easy to throw together - most of the code is pretty straightforward and the `Instrumentation` implementation is able to reuse/adhere to a lot of existing infrastructure for plugging in config-provided elements.  I augmented the instrumentation tests with more thorough unit testing of the JSPS expression "parser".

One large design topic to be explored is whether to continue using the "JSPS" Java-like expression interpreter or to "buy" an off-the-shelf library for such a thing.  This would be a future decision and merging this prototype shouldn't be gated on it - it should be easy enough to swap out that component without touching the rest of the code.

Bikeshedding of all chosen names is welcome!


bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public String toString() {
+      return className + "." + methodName + ":spanName=" + spanName + ",attrs=" + attributes;
+    }
+  }
+
+  // Unfortunately the particular sequence of who needs access to this and who can create (based on
+  // having the yaml parser library loaded) is awkward.  Would prefer this to be a simple
+  // static final immutable computed at startup...
+  private static final AtomicReference<List<Rule>> GlobalRules =
```
breedx-splk, line 56: Field should be named `globalRules` (camelCase not PascalCase)

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class NocodeRules {
+
+  public static class Rule {
```
breedx-splk, line 28:
```suggestion
  public final static class Rule {
```

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -0,0 +1,79 @@
...
+package com.splunk.opentelemetry.javaagent.bootstrap.nocode;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class NocodeRules {
```
breedx-splk, line 26:
```suggestion
public final class NocodeRules {
```

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
breedx-splk: same with `Name2Rule` -> `name2Rule` below

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -0,0 +1,79 @@
...
+      n2r.put(r.className + "." + r.methodName, r);
+    }
+    Name2Rule.set(Collections.unmodifiableMap(n2r));
+  }
+
+  public static List<Rule> getGlobalRules() {
+    return GlobalRules.get();
+  }
+
+  public static Rule findRuleByClassAndMethod(String className, String methodName) {
```
breedx-splk, line 75: nit: I think the "by class and method" is redundant because the arguments already convey that.....I'd just call it `findRule` or simply `find`. 🤷🏻

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/JSPS.java
```diff
@@ -0,0 +1,163 @@
...
+// As some illustrative examples:
+// this.getHeaders().get("X-Custom-Header").substring(5)
+// param0.getDetails().getCustomerAccount().getAccountType()
+// The limitations are:
+// no access to variables other than 'this' and 'paramN' (N indexed at 0)
+// no control flow (if), no local variables, basically nothing other than a single chain of method
+// calls
+// Methods calls are limited to either 0 or 1 parameters currently
+// Parameters must be literals and only integral (int/long), string, and boolean literals are
+// currently supported
```
breedx-splk, line 36: Kinda surprised the linter didn't call this out or fix it....should be 
```
/**
 * class comment style like this
 ...
 */
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/JSPS.java
```diff
@@ -0,0 +1,163 @@
...
+
+  // This sequence of methods is here because:
+  // - we want to try a variety of parameter types to match a literal
+  //      e.g., someMethod(5) could match someMethod(int) or someMethod(Object)
+  // - module rules around reflection make some "legal" things harder and require
+  //   looking for a public class/interface matching the method to call
+  //      e.g., a naive reflective call through
+  //         this.getSomeHashMap().entrySet().size()
+  //      would fail simply because the HashMap$EntrySet implementation
+  //      is not public, even though the interface it's being call through is.
```
breedx-splk, line 117: same comment about javadoc style

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/JSPS.java
```diff
@@ -0,0 +1,163 @@
...
+    Method m = findMatchingMethod(methodName, curObject.getClass(), actualParamTypes);
+    if (m == null) {
+      throw new NoSuchMethodException(
+          "Can't find matching method for " + methodName + " on " + curObject.getClass().getName());
+    }
+    return m;
+  }
+
+  // Returns null for none found
+  private static Method findMatchingMethod(String methodName, Class c, Class[] actualParamTypes) {
```
breedx-splk, line 129: nit: prefer `Class clazz` over just `c` for readability.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/JSPS.java
```diff
@@ -0,0 +1,163 @@
...
+// this.getHeaders().get("X-Custom-Header").substring(5)
+// param0.getDetails().getCustomerAccount().getAccountType()
+// The limitations are:
+// no access to variables other than 'this' and 'paramN' (N indexed at 0)
+// no control flow (if), no local variables, basically nothing other than a single chain of method
+// calls
+// Methods calls are limited to either 0 or 1 parameters currently
+// Parameters must be literals and only integral (int/long), string, and boolean literals are
+// currently supported
+public class JSPS {
```
breedx-splk, line 37:
```suggestion
public final class JSPS {
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
```diff
@@ -0,0 +1,64 @@
...
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.incubator.semconv.code.CodeAttributesExtractor;
+import io.opentelemetry.instrumentation.api.incubator.semconv.util.ClassAndMethod;
+import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
+import java.util.Collections;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+public class NocodeAttributesExtractor
```
breedx-splk, line 28:
```suggestion
public final class NocodeAttributesExtractor
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  public void onStart(
+      AttributesBuilder attributesBuilder, Context context, NocodeMethodInvocation mi) {
+    codeExtractor.onStart(attributesBuilder, context, mi.getClassAndMethod());
+
+    Map<String, String> attributes = Collections.EMPTY_MAP;
+    if (mi.getRule() != null) {
+      attributes = mi.getRule().attributes;
+    }
```
breedx-splk, line 44: There are a lot of opinions about direct field access, and yeah, with a lack of data classes available to us, I get the desire to not generate getters.....but this problem here would just disappear if it was encapsulated like`Map<String, String> attributes = methodInvocation.getRuleAttributes()`

Also reduces the law-of-demeter situation.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
```diff
@@ -0,0 +1,64 @@
...
+    for (String key : attributes.keySet()) {
```
breedx-splk, line 45: You can leverage `attributes.forEach((k,v) -> { ... })` rather than doing the iteration manually here.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInstrumentation.java
```diff
@@ -0,0 +1,98 @@
...
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.instrumentation.api.incubator.semconv.util.ClassAndMethod;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeTransformer;
+import java.lang.reflect.Method;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class NocodeInstrumentation implements TypeInstrumentation {
```
breedx-splk, line 35:
```suggestion
public final class NocodeInstrumentation implements TypeInstrumentation {
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInstrumentation.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return hasSuperType(named(rule.className));
+  }
+
+  @Override
+  public void transform(TypeTransformer transformer) {
+    transformer.applyAdviceToMethod(
+        named(rule.methodName), NocodeInstrumentation.class.getName() + "$NocodeAdvice");
```
breedx-splk, line 50: I can't remember if it's entirely necessary, but I worry about the reference, and this is conventional in upstream:
```suggestion
        named(rule.methodName), this.getClass().getName() + "$NocodeAdvice");
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInstrumentation.java
```diff
@@ -0,0 +1,98 @@
...
+        return;
+      }
+      context = instrumentor().start(parentContext, otelInvocation);
+      scope = context.makeCurrent();
+    }
+
+    public NocodeAdvice() {
+      super();
+    }
+
```
breedx-splk, line 81: Not needed.
```suggestion
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeMethodInvocation.java
```diff
@@ -0,0 +1,51 @@
...
+    return rule;
+  }
+
+  public Object getThiz() {
+    return thiz;
+  }
+
+  public Object[] getParameters() {
+    return parameters;
+  }
```
breedx-splk, line 46: I don't know how much we care, but I wanted to at least point out that this leaks a mutable array that a caller _could_ change out from under us....making this class not truly immutable.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeModule.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public NocodeModule() {
+    super("nocode");
+    YamlParser yp = new YamlParser();
+    NocodeRules.setGlobalRules(yp.getInstrumentationRules());
+  }
+
+  @Override
+  public List<TypeInstrumentation> typeInstrumentations() {
+    ArrayList<TypeInstrumentation> answer = new ArrayList<>();
```
breedx-splk, line 38:
```suggestion
    List<TypeInstrumentation> answer = new ArrayList<>();
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Logger;
+import org.snakeyaml.engine.v2.api.Load;
+import org.snakeyaml.engine.v2.api.LoadSettings;
+
+public class YamlParser {
```
breedx-splk, line 32:
```suggestion
public final class YamlParser {
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
```diff
@@ -0,0 +1,86 @@
...
+
+  public List<NocodeRules.Rule> getInstrumentationRules() {
+    return InstrumentationRules;
+  }
+
+  private static List<NocodeRules.Rule> load() {
+    try {
+      return loadUnsafe();
+    } catch (Exception e) {
+      logger.severe("Can't load configured yaml: " + e);
```
breedx-splk, line 51: Prefer to keep the exception details rather than `toString()`. Also added the word `nocode` to the message for context.

```suggestion
      logger.log(Level.SEVERE, "Can't load configured nocode yaml.", e)
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
```diff
@@ -0,0 +1,86 @@
...
+    if (yamlFileName == null || yamlFileName.trim().isEmpty()) {
+      return Collections.emptyList();
+    }
+    Reader yamlReader = new FileReader(yamlFileName.trim());
+
+    Load load = new Load(LoadSettings.builder().build());
+    Iterable<Object> parsedYaml = load.loadAllFromReader(yamlReader);
+    ArrayList<NocodeRules.Rule> answer = new ArrayList<>();
+    for (Object yamlBit : parsedYaml) {
+      List l = (List) yamlBit;
```
breedx-splk, line 67: Sorry, can't let `l` slide. 🙃 Also maybe we can cast to the generic up here and avoid it inside the loop?
```suggestion
      List<Map<String,Object>> rulesMap = (List<Map<String,Object>>) yamlBit;
```

instrumentation/nocode/src/test/config/nocode.yml
```diff
@@ -0,0 +1,24 @@
...
+      value: param0.toString().substring(0)
+
+- class: com.splunk.opentelemetry.instrumentation.nocode.NocodeInstrumentationTest$SampleClass
+  method: doInvalidRule
+  spanName: "this.thereIsNoSuchMethod()"
+  spanKind: INVALID
+  attributes:
+    - key: "notpresent"
+      value: "invalid.noSuchStatement()"
+
```
breedx-splk, line 24: It would be cool if we could be consistent in our use or lack of quotes. I'd prefer to drop them in all places where they're not needed (one of the few benefits if yaml, heh)

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JSPSTest.java
```diff
@@ -0,0 +1,205 @@
...
+          {"this.get(\"key\")", "value"},
+          {"this.get(\"key\").substring(1)", "alue"},
+          {"param0.isEmpty()", "false"},
+          {"param0.contains(\"present\")", "true"},
+          {"this.entrySet().size()", "1"},
+        };
+    for (String[] test : tests) {
+      assertEquals(test[1], JSPS.evaluate(test[0], thiz, new Object[] {param0}), test[0]);
+    }
+  }
```
breedx-splk, line 55: Use a proper [parameterized tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests) please! Also applies to methods below.

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JSPSTest.java
```diff
@@ -0,0 +1,205 @@
...
+        new String[] {
+          "this.get(\"key\").substring(1)",
+          " this.get(\"key\").substring(1)",
+          "this .get(\"key\").substring(1)",
+          "this. get(\"key\").substring(1)",
+          "this.get (\"key\").substring(1)",
+          "this.get( \"key\").substring(1)",
+          "this.get(\"key\" ).substring(1)",
+          "this.get(\"key\") .substring(1)",
+          "this.get(\"key\"). substring(1)",
```
breedx-splk, line 189: Makes me wonder if newlines and/or tabs would be allowed at this boundary. Might actually be a nice feature for longer blocks of code that is definitely not code. Maybe add test coverage for those?

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JSPSTest.java
```diff
@@ -0,0 +1,205 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+public class JSPSTest {
```
breedx-splk, line 29:
```suggestion
class JSPSTest {
```

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JSPSTest.java
```diff
@@ -0,0 +1,205 @@
...
+  private static final Map<String, String> thiz = new HashMap<>();
+  private static final Set<String> param0 = new HashSet<>();
+
+  static {
+    thiz.put("key", "value");
+    param0.add("present");
+  }
+
+  @Test
+  public void testBasicBehavior() {
```
breedx-splk, line 39: Same with test methods, prefer to omit `public` and for brevity.

```suggestion
 void testBasicBehavior() {
```

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JSPSTest.java
```diff
@@ -0,0 +1,205 @@
...
+    for (String test : tests) {
+      assertEquals("alue", JSPS.evaluate(test, thiz, new Object[] {param0}), test);
+    }
+  }
+
+  public void testManyParams() {
+    Object[] params = new Object[13];
+    Arrays.fill(params, new Object());
+    assertEquals(
+        "java.lang.Object", JSPS.evaluate("param12.getClass().getName()", new Object(), params));
```
breedx-splk, line 203: Makes me instantly wonder what `"param13.getClass().getName()"` would do....do we have coverage for going out of bounds? If not, can we add it?

breedx-splk (commented):
It's a great start for sure! It's still wild to me that folks have asked for this stuff, but here we are. Nice work turning this around so quickly. 😎

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  // Unfortunately the particular sequence of who needs access to this and who can create (based on
+  // having the yaml parser library loaded) is awkward.  Would prefer this to be a simple
+  // static final immutable computed at startup...
+  private static final AtomicReference<List<Rule>> GlobalRules =
+      new AtomicReference<>(Collections.EMPTY_LIST);
+  // Using className.methodName as the key
+  private static final AtomicReference<Map<String, Rule>> Name2Rule =
+      new AtomicReference<>(Collections.emptyMap());
```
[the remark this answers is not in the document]
johnbley: That cleaned it up a lot!

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
johnbley: Because this is used from inside an `@Advice` this triggered some sort of classloader/helper class problem for me with the closure and rather than fight it I left it this way.

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JSPSTest.java
johnbley: Yeah, that's covered several times over in `testInvalidJspsReturnNull`.

instrumentation/nocode/src/test/config/nocode.yml
johnbley: I'm specifically testing this behavior of yaml for our intended use!

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JSPSTest.java
johnbley: The implementation uses `Character.isWhitespace`; threw in a few other whitespace characters for good measure!

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
johnbley: Yeah, this was the code that was previously using reflection due to a classloader/gradle issue.  Went through and cleaned up types, which allow me to drop several local variables/casts/etc.

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JSPSTest.java
johnbley: TIL, thanks!  My style was JUnit circa 1999.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInstrumentation.java
```diff
@@ -0,0 +1,93 @@
...
+        @Advice.Origin Method method,
+        @Advice.Local("otelInvocation") NocodeMethodInvocation otelInvocation,
+        @Advice.Local("otelContext") Context context,
+        @Advice.Local("otelScope") Scope scope,
+        @Advice.Thrown Throwable error) {
+      if (scope == null) {
+        return;
+      }
+      scope.close();
+      // I wonder why the original methods instrumentation had support for modifying the return
```
laurit, line 88: this is because the original instrumentation has special handling for some async types such as `CompletableFuture` where span is ended when the future ends not when the method returns

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeMethodInvocation.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public ClassAndMethod getClassAndMethod() {
+    return classAndMethod;
+  }
+
+  public Map<String, String> getRuleAttributes() {
+    return rule == null ? Collections.EMPTY_MAP : rule.attributes;
```
laurit, line 60: I would guess that you get a compiler warning where, using `emptyMap()` instead should help.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeModule.java
```diff
@@ -0,0 +1,60 @@
...
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeAttributesExtractor",
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeMethodInvocation",
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeSpanKindExtractor",
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeSpanNameExtractor");
+  }
+
+  @Override
+  public int order() {
+    return Integer.MAX_VALUE;
+  }
```
laurit, line 59: any reason why you added this? usually we just use the default order.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeModule.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  public List<String> getAdditionalHelperClassNames() {
+    return Arrays.asList(
+        "com.splunk.opentelemetry.instrumentation.nocode.JSPS",
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeSingletons",
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeAttributesExtractor",
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeMethodInvocation",
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeSpanKindExtractor",
+        "com.splunk.opentelemetry.instrumentation.nocode.NocodeSpanNameExtractor");
+  }
```
laurit, line 54: try adding `id("splunk.muzzle-conventions")` to `build.gradle.kts` then you shouldn't need to add this

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeSpanKindExtractor.java
```diff
@@ -0,0 +1,34 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.SpanKindExtractor;
+
+public class NocodeSpanKindExtractor implements SpanKindExtractor<NocodeMethodInvocation> {
+  @Override
+  public SpanKind extract(NocodeMethodInvocation mi) {
+    if (mi.getRule() == null || mi.getRule().spanKind == null) {
+      return SpanKind.INTERNAL;
+    }
+    try {
+      return SpanKind.valueOf(mi.getRule().spanKind.toUpperCase());
```
laurit, line 29: usually we use `toUpper` with `Locale.ROOT`

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/JSPS.java
```diff
@@ -0,0 +1,173 @@
...
+ * <ul>
+ *   <li>no access to variables other than 'this' and 'paramN' (N indexed at 0)
+ *   <li>no control flow (if), no local variables, basically nothing other than a single chain of
+ *       method calls
+ *   <li>Method calls are limited to either 0 or 1 parameters currently
+ *   <li>Parameters must be literals and only integral (int/long), string, and boolean literals are
+ *       currently supported
+ * </ul>
+ */
+public final class JSPS {
```
laurit, line 46: I think it is better to replace this with an existing expression language implementation as the parsing will only get more complicated as features are added. For example parsing `this.get(")")` will already be a challenge.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeModule.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.auto.service.AutoService;
+import com.splunk.opentelemetry.javaagent.bootstrap.nocode.NocodeRules;
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public final class NocodeModule extends InstrumentationModule {
```
laurit, line 28: I think this module should be disabled by default for now

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInstrumentation.java
johnbley: Yes, on rereading that code I see that now.  I'll change the comment to reflect this.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeModule.java
johnbley: My thinking was that if somebody had, e.g., a servlet that they wanted some custom nocode instrumentation on, we would want our instrumentation to run *after* the "regular" servlet instrumentation (which would correctly handle context propagation, etc.).  If this is in fact the way this `order()` method works I'll add a comment to this effect, and if I'm wrong then I'll delete it.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeModule.java
johnbley:
```
java.lang.IllegalStateException: Failed to transform class com.splunk.opentelemetry.instrumentation.nocode.NocodeModule: [java.lang.IllegalStateException: java.lang.reflect.InvocationTargetException]
```

breedx-splk (approved):

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInstrumentationTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+// This test has "test/config/nocode.yml" applied to it by the gradle environment setting
+class NocodeInstrumentationTest {
+  @RegisterExtension
+  static final InstrumentationExtension testing = AgentInstrumentationExtension.create();
+
+  @Test
+  void testBasicMethod() {
+    new SampleClass().doSomething();
+    // FIXME what is scope version verification and why doesn't it pass?
```
laurit, line 38: for this to work you need to carefully choose the instrumentation name, fixed in https://github.com/johnbley/splunk-otel-java/pull/1

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -0,0 +1,68 @@
...
+      this.attributes = Collections.unmodifiableMap(new HashMap<>(attributes));
+    }
+
+    public String toString() {
+      return className + "." + methodName + ":spanName=" + spanName + ",attrs=" + attributes;
+    }
+  }
+
+  // Using className.methodName as the key
+  private static final ConcurrentHashMap<String, Rule> Name2Rule = new ConcurrentHashMap<>();
```
laurit, line 53: `Name2Rule` is an unusual name for a field, usually field names start with a lowercase letter.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInstrumentation.java
```diff
@@ -0,0 +1,94 @@
...
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(
+        @Advice.Origin("#t") Class<?> declaringClass,
+        @Advice.Origin("#m") String methodName,
+        @Advice.Local("otelInvocation") NocodeMethodInvocation otelInvocation,
+        @Advice.Local("otelContext") Context context,
+        @Advice.Local("otelScope") Scope scope,
+        @Advice.This Object thiz,
+        @Advice.AllArguments Object[] methodParams) {
+      NocodeRules.Rule rule = NocodeRules.find(declaringClass.getName(), methodName);
```
laurit, line 64: rule could be null here because matcher is registered with `hasSuperType(named(rule.className))`. Declaring class could be a subclass of the class used in the rule.

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
johnbley: Fixed and fixed.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInstrumentation.java
johnbley: Excellent catch, thank you!  I dropped it back to plain `named` and added a comment to the effect that we could support more complex logic by generating advice classes dynamically from the rule.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeModule.java
johnbley: @laurit thoughts on this?

laurit: I think your understanding is correct. A comment would be great.

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+package com.splunk.opentelemetry.javaagent.bootstrap.nocode;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class NocodeRules {
```
robsunday, line 24: Looks like you want this to be a singleton, so please add also a private parameterless constructor

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/JSPS.java
```diff
@@ -0,0 +1,173 @@
...
+      return null;
+    }
+    if (Modifier.isPublic(clazz.getModifiers())) {
+      try {
+        return clazz.getMethod(methodName, actualParamTypes);
+      } catch (NoSuchMethodException nsme) {
+        // keep trying
+      }
+    }
+    // not public, try interfaces and supertype
```
robsunday, line 150: It looks like Class#getMethod already is traversing up to Object class. For interfaces, appropriate method must be implemented in processed class or its superclass anyway.
So maybe this part is not needed?

johnbley: No, it definitely is because of where the method is declared and implemented.  When it is declared in a public interface but implemented (as a public method) in a non-public class, `getMethod` on the implementation will return a `Method` that isn't callable unless the module is open for reflection, while `getMethod` on the interface will return one that is callable.  This was borne out by unit testing `this.getSomeMap().entrySet().size()` where (e.g.) `HashMap$EntrySet` isn't public.

robsunday: Thanks for explanation!

robsunday (approved):

PR #2185, opened 2025-02-14
Add TextMapPropagator to Inform Downstream Services About Trace Snapshot Selection.

tduncan:
This PR adds a custom `TextMapPropagator` responsible for identifying when a trace has been selected for trace snapshotting and including a special header that downstream services can use to register the same trace for snapshotting.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
@@ -58,10 +58,13 @@ public class SnapshotProfilingSdkCustomizer implements AutoConfigurationCustomiz
     };
   }
 
-  private BiFunction<TextMapPropagator, ConfigProperties, TextMapPropagator> addProfilingSignalPropagator(TraceRegistry registry) {
+  private BiFunction<TextMapPropagator, ConfigProperties, TextMapPropagator>
+      addProfilingSignalPropagator(TraceRegistry registry) {
     return (textMapPropagator, properties) -> {
-      if (snapshotProfilingEnabled(properties) && textMapPropagator instanceof W3CTraceContextPropagator) {
-        return TextMapPropagator.composite(textMapPropagator, new SnapshotProfilingSignalPropagator(registry));
+      if (snapshotProfilingEnabled(properties)
+          && textMapPropagator instanceof W3CTraceContextPropagator) {
+        return TextMapPropagator.composite(
+            textMapPropagator, new SnapshotProfilingSignalPropagator(registry));
       }
       return textMapPropagator;
     };
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
@@ -20,6 +20,8 @@ import static com.splunk.opentelemetry.profiler.Configuration.CONFIG_KEY_ENABLE_
 
 import com.google.auto.service.AutoService;
 import com.google.common.annotations.VisibleForTesting;
+import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
+import io.opentelemetry.context.propagation.TextMapPropagator;
 import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizer;
 import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvider;
 import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
@@ -41,8 +43,9 @@ public class SnapshotProfilingSdkCustomizer implements AutoConfigurationCustomiz
 
   @Override
   public void customize(AutoConfigurationCustomizer autoConfigurationCustomizer) {
-    autoConfigurationCustomizer.addTracerProviderCustomizer(
-        snapshotProfilingSpanProcessor(registry));
+    autoConfigurationCustomizer
+        .addTracerProviderCustomizer(snapshotProfilingSpanProcessor(registry))
+        .addPropagatorCustomizer(addProfilingSignalPropagator(registry));
   }
 
   private BiFunction<SdkTracerProviderBuilder, ConfigProperties, SdkTracerProviderBuilder>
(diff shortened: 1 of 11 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSignalPropagator.java
tduncan, line 1: Ideally we'd like to use the `Baggage` API but we haven't figured out how to make that work just yet.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/DistributedProfilingSignalTest.java
```diff
@@ -0,0 +1,115 @@
...
+          .withProperty("splunk.snapshot.profiler.enabled", "true")
+          .with(upstreamCustomizer)
+          .build();
+
+  @RegisterExtension
+  public final Server upstream =
+      Server.builder(upstreamSdk).named("upstream").performing(ExitCall.to(downstream)).build();
+
+  @Test
+  void traceSnapshotVolumePropagatesAcrossProcessBoundaries() {
```
tduncan, line 70: This test attempts to simulate a trace spanning two distributed services, asserting the downstream service registers the same trace as the upstream service.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/simulation/Server.java
tduncan, line 1: This class, as well as `ExitCall` and `Message`, are inspired by similar classes found in the OpenTelemetry Java SDK tests.

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/testbed/clientserver/Server.java

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/Volume.java
tduncan, line 1: The necessity of this enum may not be obvious from it's usage in `SnapshotProfilingSignalPropagator` as the decision is effectively a `true|false` situation. However the idea here is that snapshotting would be a dial rather than a switch with the different OpenTelemetry signals modifying how much data to collect based on the selected `Volume`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
```diff
@@ -55,6 +58,18 @@ public void customize(AutoConfigurationCustomizer autoConfigurationCustomizer) {
     };
   }
 
+  private BiFunction<TextMapPropagator, ConfigProperties, TextMapPropagator>
+      addProfilingSignalPropagator(TraceRegistry registry) {
+    return (textMapPropagator, properties) -> {
+      if (snapshotProfilingEnabled(properties)
+          && textMapPropagator instanceof W3CTraceContextPropagator) {
```
laurit, line 65: Relying on the presence of the w3c tracecontext propagator isn't ideal. Perhaps this could also be achieved by modifying the `otel.propagators` propagator. If it is not set set it to `"tracecontext", "baggage", "splunk-spansphtot-profiler"` otherwise append `"splunk-spansphtot-profiler"`. Now when we add a `ConfigurablePropagatorProvider` implementation with name `"splunk-spansphtot-profiler"` it should get added as the last propagator.
Wdyt? If you believe this could work and is beneficial you can work on this in a later PR if that is more convenient for you.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSignalPropagator.java
```diff
@@ -0,0 +1,59 @@
...
+    SpanContext spanContext = Span.fromContext(context).getSpanContext();
+    if (registry.isRegistered(spanContext)) {
+      setter.set(carrier, PROFILING_SIGNAL, Volume.HIGHEST.toString());
+    }
+  }
+
+  @Override
+  public <C> Context extract(Context context, C carrier, TextMapGetter<C> getter) {
+    Volume volume = Volume.fromString(getter.get(carrier, PROFILING_SIGNAL));
+    SpanContext spanContext = Span.fromContext(context).getSpanContext();
```
laurit, line 53: I guess this could end up returning the constant for invalid span context when there is no span. Idk if that is an issue.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/Volume.java
```diff
@@ -0,0 +1,39 @@
...
+  OFF,
+  HIGHEST;
+
+  static Volume fromString(String value) {
+    if (value == null) {
+      return OFF;
+    }
+
+    try {
+      return Volume.valueOf(value.toUpperCase());
```
laurit, line 29: We usually use `toUpperCase` and `toLowerCase` with locale like `toUpperCase(Locale.ROOT)`. Using explicit locale helps against surprises where in the user locale characters are converted in an unexpected way. For example in Turkish language the upper case of `i` is not `I` but `İ` which here could result in a difficult to debug issue.

laurit (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSignalPropagator.java
tduncan: I don't think that is an issue but I'll add a test. I also think the sequence here can change a bit in that it's not necessary to even retrieve a `Span` from the `Context` when the snapshot `Volume` is `OFF`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
tduncan: I agree this approach is potentially brittle in that the extension no longer really works if an agent is configured without the default `tracecontext` propagator. We ended up here for a few reasons.

1. The extension propagator only needs to execute once so making a composite for every configured propagator is unnecessary.
2. The other default propagator (`baggage`) didn't have the trace context loaded so special code was required to work around that situation. Ignoring that propagator removed the need for this special code.
3. Requiring manual configuration of the propagator seemed very brittle, specifically the knowledge to not forget about the `baggage` and `tracecontext` propagators.
4. Using the provider approach also required the classic singleton design of classes like `TraceRegistry` as the SDK customization mechanisms are disconnected from each other but still need to share the same object instances. That introduced its own set of challenges for both testing and object lifecycle management that aren't present when the SDK customization is centralized.

laurit: > Requiring manual configuration of the propagator seemed very brittle, specifically the knowledge to not forget about the baggage and tracecontext propagators.

Not forgetting about the baggage and tracecontext is imo not an issue as you have to get it right only once. We already have similar stuff in https://github.com/signalfx/splunk-otel-java/blob/main/custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java You could just override the default value in `defaultProperties` to `tacecontext,baggage,profiler` and in `customize` append `profiler` if it is not present.

> Using the provider approach also required the classic singleton design of classes like TraceRegistry as the SDK customization mechanisms are disconnected from each other but still need to share the same object instances. That introduced its own set of challenges for both testing and object lifecycle management that aren't present when the SDK customization is centralized.

I guess an alternative would be to ask for an another extension point. I personally would prefer it to work reliably even if this requires using a singleton.

PR #2195, opened 2025-02-19
Use OpenTelemetry Baggage For Snapshot Volume Propagation

tduncan:
This PR converts the snapshot volume propagation mechanism to use OpenTelemetry's `Baggage` instead of HTTP headers. Using `Baggage` better integrates with existing OpenTelemetry data propagation mechanism specifically leaning on language implementations for baggage propagation enabling the snapshot volume signal to cross process boundaries of any OpenTelemetry instrumented service.

This new implementation also uses the Java Agent's `ConfigurablePropagatorProvider` which does introduce the additional requirement for users to manually add the propagator to the list specified in the `otel.propagators` configuration property.

I addition the propagator is dependent on Java's `W3CBaggagePropagator` and MUST be configured to executed after it.


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
```diff
@@ -31,15 +31,14 @@ public class SnapshotProfilingSpanProcessor implements SpanProcessor {
 
   @Override
   public void onStart(Context context, ReadWriteSpan span) {
-    if (isRoot(span) && isEntry(span)) {
-      registry.register(span.getSpanContext());
+    if (isEntry(span)) {
```
laurit, line 34: idk if this affects anything but messaging instrumentations can create nested `CONSUMER`

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotVolumePropagator.java
```diff
@@ -25,39 +25,25 @@
...
-    }
-  }
+  public <C> void inject(Context context, C carrier, TextMapSetter<C> setter) {}
 
   @Override
   public <C> Context extract(Context context, C carrier, TextMapGetter<C> getter) {
-    Volume volume = Volume.fromString(getter.get(carrier, PROFILING_SIGNAL));
-    if (volume == Volume.OFF) {
-      return context;
+    if (isTraceRoot(context)) {
```
laurit, line 39: I find this hard to understand, a comment that describes what is going on here could be helpful. As far as I can tell if this propagator is run before other propagators it will set the volume to highest. If this propagator is run after other propagators it will set the volume to highest only when there were no trace propagation headers.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotVolumePropagator.java
```diff
@@ -25,39 +25,25 @@
...
+    return context;
+  }
 
+  private boolean isTraceRoot(Context context) {
     SpanContext spanContext = Span.fromContext(context).getSpanContext();
-    if (spanContext.isValid()) {
-      registry.register(spanContext);
-    }
-    return context;
+    return !spanContext.isValid();
```
laurit, line 47: If you wish to make this shorter you could use `Span.fromContextOrNull(context) == null`

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotVolumePropagator.java
tduncan: Sure. I'll add an explanatory comment in the code.

The trace root is the decision point. Previously this was detected in the span processor where we could inspect the span directly but in the propagator we have to infer indirectly via the context. When the trace root is inferred the code decides whether or not to snapshot or not (at the moment the answer is always `YES!`).

tduncan:
> > This new implementation also uses the Java Agent's `ConfigurablePropagatorProvider` which does introduce the additional requirement for users to manually add the propagator to the list specified in the `otel.propagators` configuration property.
> > I addition the propagator is dependent on Java's `W3CBaggagePropagator` and MUST be configured to executed after it.
> 
> Any reason you didn't configure this automatically. You can change the value for `otel.propagators` in agent code.

I wasn't aware that was possible, but now that you bring it up I suppose that is possible using a property customizer?

laurit:
> I wasn't aware that was possible, but now that you bring it up I suppose that is possible using a property customizer?

yes, see https://github.com/signalfx/splunk-otel-java/blob/2185148bf1bb0960857a3e764f92c4f88e655796/custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java for ideas

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
tduncan: I don't think it does, but it could pose an issue in the near future when the profiling code is plugged in.

tduncan:
> > This PR converts the snapshot volume propagation mechanism to use OpenTelemetry's `Baggage` instead of HTTP headers. Using `Baggage` better integrates with existing OpenTelemetry data propagation mechanism specifically leaning on language implementations for baggage propagation enabling the snapshot volume signal to cross process boundaries of any OpenTelemetry instrumented service.
> 
> One potential issue with using baggage is that users can affect it. For example consider a user application that wishes to propagate baggage to a downstream service. It could use
> 
> ```java
> Baggage baggage = Baggage.builder().put("test", "test").build();
> try (Scope ignored = baggage.makeCurrent()) {
>   // call another service  
> }
> ```
> 
> which will clear our baggage. If the user application does not expect baggage from upstream or wish to propagate it this could be fine. I suspect we can't expect everybody to use `Baggage.current().toBuilder().put("key", "value").build()`.


I was not aware of this foot gun! I made this exact mistake in `Volume.storeInContext`!

tduncan:
> > I wasn't aware that was possible, but now that you bring it up I suppose that is possible using a property customizer?
> 
> yes, see https://github.com/signalfx/splunk-otel-java/blob/2185148bf1bb0960857a3e764f92c4f88e655796/custom/src/main/java/com/splunk/opentelemetry/SplunkConfiguration.java for ideas

After exploring a bit I think it makes sense to address auto configuring the propagator in a followup PR.

laurit (approved):

PR #2197, opened 2025-02-20
Mark Spans With Snapshot Attribute

tduncan:
This PR adds an attribute to all entry spans in a trace marked for snapshotting. The attribute indicates whether the snapshot profiler was initiated and is intended to help the UI differentiate between traces with and without callgraphs (an icon will appear next to the trace ID).

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
@@ -16,6 +16,8 @@
 
 package com.splunk.opentelemetry.profiler.snapshot;
 
+import static com.splunk.opentelemetry.profiler.ProfilingSemanticAttributes.SNAPSHOT_PROFILING;
+
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.sdk.trace.ReadWriteSpan;
@@ -36,7 +38,7 @@ public class SnapshotProfilingSpanProcessor implements SpanProcessor {
     }
 
     if (isEntry(span) && registry.isRegistered(span.getSpanContext())) {
-      span.setAttribute("splunk.snapshot.profiling", true);
+      span.setAttribute(SNAPSHOT_PROFILING, true);
     }
   }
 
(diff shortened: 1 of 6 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
```diff
@@ -34,6 +34,10 @@ public void onStart(Context context, ReadWriteSpan span) {
     if (isRoot(span) && isEntry(span)) {
       registry.register(span.getSpanContext());
     }
+
+    if (isEntry(span) && registry.isRegistered(span.getSpanContext())) {
+      span.setAttribute("splunk.snapshot.profiling", true);
```
tduncan, line 39: I was also considering whether the attribute should be less specific to profiling and allow for potentially other types of data to be included. For example by defining the attribute as a string array where valid values might include `callstacks` or `logs`. Not sure if that is practical or not.

laurit, line 39: It is preferable to create a constant for the attribute key `AttributeKey.booleanKey("splunk.snapshot.profiling")` and use that constant instead of the string.
`splunk.snapshot.profiling` should probably be also described in the gdi spec.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotSnapAttributeTest.java
```diff
@@ -0,0 +1,119 @@
...
+      OpenTelemetrySdkExtension.builder()
+          .with(customizer)
+          .withProperty("splunk.snapshot.profiler.enabled", "true")
+          .build();
+
+  @ParameterizedTest
+  @SpanKinds.Entry
+  void addSnapshotSpanAttributeToEntrySpans(SpanKind kind, Tracer tracer) {
+    var span = (ReadWriteSpan) tracer.spanBuilder("root").setSpanKind(kind).startSpan();
+    try (var ignored = span.makeCurrent()) {
```
laurit, line 46: making the span current shouldn't really be needed for these tests

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
tduncan: Re: the GDI spec

Yes. I was exploring that more recently and came to the conclusion that we need to add quite a bit, including this attribute.

breedx-splk: I like keeping separate things separate, and see very little benefit it combing features/flags under one attribute. 👍🏻

breedx-splk: Keeping the attribute key as a constant should also yield performance improvements.

breedx-splk (approved):
Looks good to me.

laurit (approved):

PR #2198, opened 2025-02-20
Add ability to add additional context for appd integrations.

breedx-splk:
Relates to this [gdi-specification PR](https://github.com/signalfx/gdi-specification/pull/335). 

This change adds two main features, both behind a config flag named `cisco.ctx.enabled` (default = `false`):

* Transfer 4 AppD context headers into corresponding span attributes
* Set two outbound headers containing the otel service name and deployment env.


custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusCustomizer.java
```diff
@@ -0,0 +1,68 @@
...
+          propagator.setServiceName(resource.getAttribute(SERVICE_NAME));
+          return resource;
+        });
+    customizer.addPropagatorCustomizer(
+        (existing, config) -> {
+          if (featureEnabled(config)) {
+            return TextMapPropagator.composite(existing, propagator);
+          }
+          return existing;
+        });
```
laurit, line 62: I suspect this isn't what you want. The customizer gets applied to all configured propagators so you may end up inserting the appd propagator multiple times.

breedx-splk: Are you sure? I could possibly see this being a problem (as you point out) if the `AppdBonusPropagator` was autoconfigured....but it's not even an `AutoService`. How would it get inserted then? Maybe I'm missing something.

breedx-splk: Thanks yeah good catch. I have reworked it...hopefully it makes more sense now.

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusConstants.java
```diff
@@ -0,0 +1,39 @@
...
+public final class AppdBonusConstants {
+  public static final String CONFIG_CISCO_CTX_ENABLED = "cisco.ctx.enabled";
+  public static final String PROPAGATOR_NAME = "appd-bonus";
+  public static final String CTX_KEY = "cisco-bonus-ctx";
+  public static final String CTX_HEADER_ACCT = "cisco-ctx-acct-id";
+  public static final String CTX_HEADER_APP = "cisco-ctx-app-id";
+  public static final String CTX_HEADER_BT = "cisco-ctx-bt-id";
+  public static final String CTX_HEADER_TIER = "cisco-ctx-tier-id";
+  public static final String CTX_HEADER_ENV = "cisco-ctx-env";
+  public static final String CTX_HEADER_SERVICE = "cisco-ctx-service";
```
laurit, line 32: Are these used outside of the propagator? If not, perhaps they should be moved there.

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusPropagator.java
```diff
@@ -0,0 +1,90 @@
...
+final class AppdBonusPropagator implements TextMapPropagator {
+
+  public static final ContextKey<AppdBonusContext> CONTEXT_KEY = ContextKey.named(CTX_KEY);
+  private static final AppdBonusPropagator INSTANCE = new AppdBonusPropagator();
+
+  @Nullable private String serviceName;
+  @Nullable private String environmentName;
+
+  @VisibleForTesting
+  AppdBonusPropagator() {}
```
laurit, line 46: at first glance couldn't see it used in the test (could be wrong)

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusSpanProcessor.java
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.appd;
+
+import static com.splunk.opentelemetry.appd.AppdBonusConstants.*;
```
laurit, line 19: usually we don't use wildcard imports

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusPropagator.java
breedx-splk: It's used in the `AppdBonusPropagatorTest`. :)

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusSpanProcessor.java
breedx-splk: Weird that spotless didn't catch it? Anyway, I'll fix it. I never look at imports. :)

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusConstants.java
breedx-splk: yeah that's fine. I thought it might be nice to have them in one place, but I'll move them to where they're used.

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusCustomizer.java
```diff
@@ -0,0 +1,87 @@
...
+    return config.getBoolean(CONFIG_CISCO_CTX_ENABLED, false);
+  }
+
+  /** Used to add the AppD propagator name to the otel.propagators list if configured. */
+  private Map<String, String> customizeProperties(ConfigProperties config) {
+    if (!config.getBoolean(CONFIG_CISCO_CTX_ENABLED, false)) {
+      return Collections.emptyMap();
+    }
+    List<String> existing = config.getList("otel.propagators", DEFAULT_PROPAGATORS);
+    if (existing.contains(PROPAGATOR_NAME)) {
```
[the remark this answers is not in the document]
breedx-splk: Thanks yeah good catch.

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusPropagator.java
```diff
@@ -0,0 +1,90 @@
...
+
+  static AppdBonusPropagator getInstance() {
```
robsunday, line 48: Does it have to be a singleton? The issue I see here is that it is not tested as a singleton, but is used as singleton in the production code.

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusPropagator.java
```diff
@@ -0,0 +1,90 @@
...
+    return INSTANCE;
+  }
+
+  @Override
+  public Collection<String> fields() {
+    return Arrays.asList(
```
robsunday, line 54: Could tis list be a final static field?

breedx-splk: Sure. 👍🏻

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusPropagator.java
breedx-splk: I thought it best, mostly to discourage public use but also because its stateful and if the state were somehow set on multiple instance it would be confusing and problematic. I've made the constructor private now and the one test that relies on state now explicitly resets the fields to null.

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusPropagator.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import javax.annotation.Nullable;
+
+final class AppdBonusPropagator implements TextMapPropagator {
+  public static final String CTX_HEADER_ACCT = "cisco-ctx-acct-id";
+  public static final String CTX_HEADER_APP = "cisco-ctx-app-id";
+  public static final String CTX_HEADER_BT = "cisco-ctx-bt-id";
+  public static final String CTX_HEADER_TIER = "cisco-ctx-tier-id";
```
laurit, line 37: Since the class is already package private might as well make these package private since they are only used in the tests.

laurit (approved):

custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusConstants.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.appd;
+
+import static io.opentelemetry.api.common.AttributeKey.stringKey;
+
+import io.opentelemetry.api.common.AttributeKey;
+
+public final class AppdBonusConstants {
```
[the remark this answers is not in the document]
breedx-splk: I'd put them in a header file if I could....:) I like the idea of having them in one place, so a reader can just ask "hey, what are the headers called? what is the context key? what is span attr?" but oh well, I moved them closer to the usage. I think it's harder to find the answer to the above questions, but oh well.

PR #2203, opened 2025-02-24
Use jexl for nocode expressions

johnbley:
Use jexl (Apache 2.0 license) for nocode expressions.  Very little syntax change from the first prototype.


instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeMethodInvocation.java
```diff
@@ -46,7 +46,7 @@ public Object getThiz() {
   /**
    * Please be careful with this, it's directly tied to @Advice.AllArguments.
    *
-   * @return
+   * @return @Advice.AllArguments - please be careful
```
laurit, line 49: just curious what concerns did you have to add this warning?

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/JexlEvaluator.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Override
+  public String evaluate(String expression, Object thiz, Object[] params) {
+    JexlContext context = new MapContext();
+    context.set("this", thiz);
+    for (int i = 0; i < params.length; i++) {
+      context.set("param" + i, params[i]);
+    }
+    try {
+      // could cache the Expression in the Rule if desired
```
laurit, line 63: I think this would be a good idea. Wouldn't need to parse the expression for each evaluation.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
```diff
@@ -33,7 +33,7 @@
 
 public final class YamlParser {
   private static final Logger logger = Logger.getLogger(YamlParser.class.getName());
-  // FIXME support method override selection - e.g., with classfile method signature or something
+  // FUTURE support method override selection - e.g., with classfile method signature or something
```
laurit, line 36: One issue with instrumenting methods from sub classes is that we currently can't easily tell from advice code which rule is tied to the method invocation because we can't just look up the rule based on the invoked class and method name. I haven't tried it out yet, but I believe we could make this work by adding an api in the upstream that would let us pass extra data, such as rule id, into the invoked advice method.

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JexlTest.java
```diff
@@ -41,10 +41,13 @@ class JSPSTest {
...
-  static Stream<Arguments> jspsToExpected() {
+  private static String evalJexl(String jexl, Object thiz, Object[] params) {
+    return new JexlEvaluator().evaluate(jexl, thiz, params);
+  }
+
+  static Stream<Arguments> jexlToExpected() {
     return Stream.of(
-        // Might be nice to support a bare "param0" or "this" but as a workaround can always use
-        // "this.toString()"
+        arguments("this", "{key=value}"),
```
laurit, line 50: an alternative to using expressions would be to use templates (I think this is what they are called in jexl) where you have something like `Hello ${this.name}`

laurit (approved):

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeMethodInvocation.java
johnbley: Jason pointed out that this is a mutable array and users of it should take care to know that it is tied to the given advice instrumentation.  https://github.com/signalfx/splunk-otel-java/pull/2184#discussion_r1955285508

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/JexlEvaluator.java
johnbley: I'll do a round of profiling for easy wins later, this will likely jump out as a pretty obvious one.

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JexlTest.java
johnbley: Yeah, jexl has a lot of features; I tried to keep things more java-y than expression-y.  When this gets more mature/field-tested I'll add docs for common scenarios, the names of the provided variables, etc.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeMethodInvocation.java
laurit: `@Advice.AllArguments` takes a parameter `readOnly` that defaults to `true`. This parameter controls whether method arguments can be mutated using the returned array. Reading https://github.com/signalfx/splunk-otel-java/pull/2184#discussion_r1955285508 I think that Jason was just concerned that array is a mutable type. I wouldn't be too concerned with this.

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JexlTest.java
laurit: Unfortunately the behavior is hard to change once this feature is made public so we'll have to consider that we are probably going to be stuck with whatever we come up here for a long time.

PR #2215, opened 2025-02-27
Probabilistically Select Traces For Snapshotting

tduncan:
This PR adds trace snapshot selection to the snapshot profiling extension. The included selection criteria is probability based with a default selection rate of 5%, though this is configurable up to a maximum selection rate of 10%.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
@@ -24,6 +24,7 @@ import com.google.auto.service.AutoService;
 import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizer;
 import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvider;
 import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.internal.DefaultConfigProperties;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
@@ -72,6 +73,7 @@ public class Configuration implements AutoConfigurationCustomizerProvider {
   public static final String CONFIG_KEY_ENABLE_SNAPSHOT_PROFILER =
       "splunk.snapshot.profiler.enabled";
   public static final String CONFIG_KEY_SNAPSHOT_SELECTION_RATE = "splunk.snapshot.selection.rate";
+  private static final double DEFAULT_SNAPSHOT_SELECTION_RATE = 0.01;
 
   @Override
   public void customize(AutoConfigurationCustomizer autoConfiguration) {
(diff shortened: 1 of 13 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -71,6 +71,9 @@ public class Configuration implements AutoConfigurationCustomizerProvider {
 
   public static final String CONFIG_KEY_ENABLE_SNAPSHOT_PROFILER =
       "splunk.snapshot.profiler.enabled";
+  public static final String CONFIG_KEY_SNAPSHOT_SELECTION_RATE = "splunk.snapshot.selection.rate";
+  private static final double DEFAULT_SNAPSHOT_SELECTION_RATE = 0.01;
```
laurit, line 75: pr description says `default selection rate of 5%`, here it looks like 1%

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -176,4 +180,31 @@ private static int getJavaVersion() {
...
   }
+
+  public static double getSnapshotSelectionRate(ConfigProperties properties) {
+    String selectionRatePropertyValue =
+        properties.getString(
+            CONFIG_KEY_SNAPSHOT_SELECTION_RATE, String.valueOf(DEFAULT_SNAPSHOT_SELECTION_RATE));
+    try {
+      double selectionRate = Double.parseDouble(selectionRatePropertyValue);
+      if (selectionRate > MAX_SNAPSHOT_SELECTION_RATE) {
+        logger.info(
```
laurit, line 191: perhaps `WARN` would be better?

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -176,4 +180,31 @@ private static int getJavaVersion() {
...
+            "Configured snapshot selection rate of '"
+                + selectionRatePropertyValue
+                + "' is higher than the maximum allowed rate. Using maximum allowed snapshot selection rate of '"
+                + MAX_SNAPSHOT_SELECTION_RATE
+                + "'");
+        return MAX_SNAPSHOT_SELECTION_RATE;
+      }
+      return selectionRate;
+    } catch (NumberFormatException e) {
+      logger.info(
```
laurit, line 201: perhaps `WARN` would be better?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ProbabilisticSnapshotSelector.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler.snapshot;
+
+import java.security.SecureRandom;
+
+class ProbabilisticSnapshotSelector implements SnapshotSelector {
+  private final SecureRandom random = new SecureRandom();
```
laurit, line 22: `SecureRandom` feels like overkill here, perhaps `Random` or `ThreadLocalRandom` would be sufficient

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
tduncan: Updating PR description!

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ProbabilisticSnapshotSelector.java
tduncan: I'll switch it to `ThreadLocalRandom`. I don't think `Random` is a good fit since multiple threads should be expected to call into the propagator.

laurit (approved):

PR #2216, opened 2025-02-28
Use type of expression result to decide span attribute type

johnbley:
Set span attribute type for a nocode expression based on what boxed type was returned from the expression:
- `boolean` from `Boolean`
- `double` from `Float` or `Double`
- `long` from `Long` or `Integer`
- `String` from everything else (via `toString()`)

This changed the types of some existing tests in desired ways, and I added additional yml-to-span instrumentation tests to cover all the cases above.

### DIFF
--- a/instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
+++ b/instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
@@ -41,9 +41,15 @@ public final class NocodeAttributesExtractor
     Map<String, String> attributes = mi.getRuleAttributes();
     for (String key : attributes.keySet()) {
       String expression = attributes.get(key);
-      String value = NocodeEvaluation.evaluate(expression, mi.getThiz(), mi.getParameters());
-      if (value != null) {
-        attributesBuilder.put(key, value);
+      Object value = NocodeEvaluation.evaluate(expression, mi.getThiz(), mi.getParameters());
+      if (value instanceof Long || value instanceof Integer) {
+        attributesBuilder.put(key, ((Number) value).longValue());
+      } else if (value instanceof Float || value instanceof Double) {
+        attributesBuilder.put(key, ((Number) value).doubleValue());
+      } else if (value instanceof Boolean) {
+        attributesBuilder.put(key, (Boolean) value);
+      } else if (value != null) {
+        attributesBuilder.put(key, value.toString());
       }
     }
   }
--- a/instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JexlTest.java
+++ b/instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JexlTest.java
@@ -41,7 +41,7 @@ class JexlTest {
     param0.add("present");
   }
 
-  private static String evalJexl(String jexl, Object thiz, Object[] params) {
+  private static Object evalJexl(String jexl, Object thiz, Object[] params) {
     return new JexlEvaluator().evaluate(jexl, thiz, params);
   }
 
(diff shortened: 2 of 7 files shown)


instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
```diff
@@ -41,9 +41,15 @@ public void onStart(
     Map<String, String> attributes = mi.getRuleAttributes();
     for (String key : attributes.keySet()) {
       String expression = attributes.get(key);
-      String value = NocodeEvaluation.evaluate(expression, mi.getThiz(), mi.getParameters());
-      if (value != null) {
-        attributesBuilder.put(key, value);
+      Object value = NocodeEvaluation.evaluate(expression, mi.getThiz(), mi.getParameters());
+      if (value instanceof Long || value instanceof Integer) {
```
robsunday, line 45: Can we get here also Short or Byte?
If so then there are few more types inheriting from Number, so maybe we could check:
1) if value is instance of Double || Float || BigDecimal
2) if value is instance of Number
3) if value is instance of Boolean
4) everything else

johnbley: `Number` is an abstract base class, and we wouldn't know best how to encode it (double or float) since the intent isn't there.  `BigDecimal` could exceed the range of `long||double` and I don't want to do strange things like "sometimes it's a double if the actual value is within range but sometimes it's a biiiiig string because it wouldn't fit in a double".  `Boolean` is already covered by the code.  

Good catch on `Short` and `Byte` though, will do.  `Character` will be `toString()`ed which is what we want anyway.

instrumentation/nocode/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/JexlTest.java
```diff
@@ -41,27 +41,31 @@ class JexlTest {
...
+        arguments("this.toString().length()", 11),
         arguments("this.get(\"key\")", "value"),
         arguments("this.get(\"key\").substring(1)", "alue"),
-        arguments("param0.isEmpty()", "false"),
-        arguments("param0.contains(\"present\")", "true"),
+        arguments("param0.isEmpty()", false),
+        arguments("param0.contains(\"present\")", true),
         arguments("\"prefix: \"+this.toString()+\" (suffix)\"", "prefix: {key=value} (suffix)"),
-        arguments("this.entrySet().size()", "1"));
+        arguments("this.entrySet().size()", 1));
```
breedx-splk, line 58: Can we add float/double coverage too please? 🙏🏻

breedx-splk: I do see coverage below in the other test, but it would still be nice here.

breedx-splk (approved):

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
robsunday: So we have two "families" of wrapper classes - floating point and integer. Both inherit from `Number`
if in step 1 we specifically handle floating point family wrapper classes (Double || Float || BigDecimal) then in step 2 we will automatically cover all integer family wrapper classes (there are many more than for floating point).
What do you think?

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
```diff
@@ -41,9 +41,18 @@ public void onStart(
     Map<String, String> attributes = mi.getRuleAttributes();
     for (String key : attributes.keySet()) {
       String expression = attributes.get(key);
-      String value = NocodeEvaluation.evaluate(expression, mi.getThiz(), mi.getParameters());
-      if (value != null) {
-        attributesBuilder.put(key, value);
+      Object value = NocodeEvaluation.evaluate(expression, mi.getThiz(), mi.getParameters());
+      if (value instanceof Long
```
laurit, line 45: Something like this has come up in a couple of places in upstream repository. Once you have the primitives and their wrappers covered the question is why stop there, otel also has attributes with array types. For example https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/fdcf0d264eb0a9b2e3a79cf01753e2869aafd295/instrumentation/logback/logback-appender-1.0/library/src/main/java/io/opentelemetry/instrumentation/logback/appender/v1_0/internal/LoggingEventMapper.java#L295 handles a bunch of array types. If you choose not to handle these you could add a comment.

PR #2217, opened 2025-02-28
Autoconfigure Snapshot Volume Propagator

tduncan:
This PR attempts to modify the `otel.propagators` value to autoconfigure the recently added `SnapshotVolumePropagatorProvider` as well as ensure the propagator is invoked after both the W3C baggage and trace context propagators.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
@@ -26,7 +26,6 @@ import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
 import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
 import java.util.Collections;
 import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.BiFunction;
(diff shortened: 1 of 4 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
```diff
@@ -55,6 +62,28 @@ public void customize(AutoConfigurationCustomizer autoConfigurationCustomizer) {
     };
   }
 
+  /**
+   * Attempt to autoconfigure the OpenTelemetry propagators to include the Splunk snapshot volume
+   * propagator and ensure it runs after both the W3C Baggage and Trace Context propagators. In
```
laurit, line 67: So if user has configured `b3` as the propagator we'll add w3c `tracecontext` propagator. Could you explain why this needs to be done. I can understand enabling the `baggage` propagator because in https://github.com/signalfx/splunk-otel-java/blob/f3561ddfff6888b962a962f9da38908f254adcdd/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/Volume.java#L32 volume is propagated through baggage. Imo it would be better to clearly state that baggage propagator is enabled because it is used to propagate the snapshot volume and that the splunk propagator just reads the baggage.

tduncan: I think I may not have had a good understanding on the purpose of the different propagators. My intent was to make sure the propagators were properly configured which from prior experience using the `otel.propagators` property meant both the trace context and baggage propagators needed to be explicitly mentioned if a custom propagator was included.

I think the appropriate logic here is to verify that the `B3` propagator isn't configured before additionally adding the trace context propagator.

laurit: Basically there are 3 cases:
1. user has not configured a propagator. In this case `otel.propagators` will be an empty list. You will have to return the default propagators used by otel sdk + spanpshot propagator so `tracecontext, baggage, splunk-snapshot` in that exact order.
2. user has configured propagators. You will have to add `splunk-snapshot` to the list of propagators, what exactly the current set of propagators is isn't really important. Since `splunk-snapshot` depends on `baggage` you may also need to add `baggage` before `splunk-snapshot`.
3. propagator list contains `none`, you should not add any propagators.
See https://github.com/open-telemetry/opentelemetry-java/blob/42056f57c87b5c1b979a0e0b6b74a132b9162245/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/PropagatorConfiguration.java#L27

tduncan: Got it. Will update.

laurit (approved):

PR #2231, opened 2025-03-06
Add ability to set span status through nocode expressions

johnbley:
Adds the ability to set `spanStatus` via nocode expressions like
```
   spanStatus: 'returnValue.statusCode() > 3 ? "OK" : "ERROR"'
```

Adds a barebones `README.md` to document some basic behavior.

### DIFF
--- a/bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
+++ b/bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
@@ -28,6 +28,8 @@ public final class NocodeRules {
     public final String methodName;
     public final String spanName; // may be null - use default of "class.method"
     public final String spanKind; // matches the SpanKind enum, null means default to INTERNAL
+    public final String spanStatus; // may be null, should return string from StatusCodes
+
     public final Map<String, String> attributes; // key name to jexl expression
 
     public Rule(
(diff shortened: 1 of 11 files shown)


instrumentation/nocode/README.md
```diff
@@ -0,0 +1,30 @@
+# "nocode" instrumentation
+
```
breedx-splk, line 2: I think there should be a section here that clarifies this featureset as "experimental" (or as otel now spells it, "development").

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -28,18 +28,22 @@ public static final class Rule {
...
+
     public final Map<String, String> attributes; // key name to jexl expression
 
     public Rule(
         String className,
         String methodName,
         String spanName,
         String spanKind,
+        String spanStatus,
         Map<String, String> attributes) {
```
breedx-splk, line 41: I get that we're hacking here, but this constructor is now pretty unwieldy. A builder could help.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeSpanStatusExtractor.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+    Object status =
+        NocodeEvaluation.evaluateAtEnd(
+            mi.getRule().spanStatus, mi.getThiz(), mi.getParameters(), returnValue, error);
+    if (status != null) {
+      try {
+        StatusCode code = StatusCode.valueOf(status.toString().toUpperCase(Locale.ROOT));
+        spanStatusBuilder.setStatus(code);
+      } catch (IllegalArgumentException noMatchingValue) {
+        // nop, should remain UNSET
```
breedx-splk, line 53: Consider logging at debug? If the user bungles an expression, they'll certainly never know....other than the status not being set.

breedx-splk (approved):
Had a couple small notes, but seems good to me.

Issue #2229, opened 2025-03-06
Fix flaky ScheduledExecutorStackTraceSamplerTest.onlyTakeStackTraceSamplesForOneThreadPerTrace


laurit:
https://scans.gradle.com/s/7xcx5n7uyubbc/tests/task/:profiler:test/details/com.splunk.opentelemetry.profiler.snapshot.ScheduledExecutorStackTraceSamplerTest/onlyTakeStackTraceSamplesForOneThreadPerTrace()?top-execution=1
@tduncan could you look at this

tduncan:
@laurit, address in https://github.com/signalfx/splunk-otel-java/pull/2233.

laurit:
resolved with https://github.com/signalfx/splunk-otel-java/pull/2233

Issue #2224, opened 2025-03-04
Address disabled "flaky test" in ProbabilisticSnapshotSelectorTest


breedx-splk:
See
https://github.com/signalfx/splunk-otel-java/blob/main/profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ProbabilisticSnapshotSelectorTest.java#L95

[based on this comment.](https://github.com/signalfx/splunk-otel-java/pull/2222/files/289549cd16545385fe0beeb9edc6e1c9f7522d60#r1979553986)

Testing things with random chance is always a challenge, but maybe a simple retry could help.

laurit:
> Testing things with random chance is always a challenge, but maybe a simple retry could help.

I'd prefer deleting the dest over adding a retry. In my opinion the problem in that test is with math. It assumes that when enough iterations are performed then the results for most of the iterations will be in +/- 10% of the expected value which doesn't seem to hold. I'm not skilled enough in math to tell whether the problem is in the number of iterations or whether the bounds set in the test leave too high of a probability for failure.

@breedx-splk could you figure out a way to assign this to @tduncan

PR #2233, opened 2025-03-06
Fix Flaky Test

tduncan:
Use a scheduler to stagger the simulated concurrent start span operations in the test that verifies only a single thread is profiled at a time.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
@@ -57,18 +57,17 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
 
   @Override
   public void start(SpanContext spanContext) {
-    ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
-    ScheduledExecutorService mappedScheduler =
-        samplers.computeIfAbsent(spanContext.getTraceId(), id -> scheduler);
-    if (scheduler == mappedScheduler) {
-      scheduler.scheduleAtFixedRate(
-          new StackTraceGatherer(spanContext.getTraceId(), Thread.currentThread().getId()),
-          SCHEDULER_INITIAL_DELAY,
-          samplingPeriod.toMillis(),
-          TimeUnit.MILLISECONDS);
-    } else {
-      scheduler.shutdown();
-    }
+    samplers.computeIfAbsent(
+        spanContext.getTraceId(),
+        traceId -> {
+          ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
+          scheduler.scheduleAtFixedRate(
+              new StackTraceGatherer(traceId, Thread.currentThread().getId()),
+              SCHEDULER_INITIAL_DELAY,
+              samplingPeriod.toMillis(),
+              TimeUnit.MILLISECONDS);
+          return scheduler;
+        });
   }
 
   @Override
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
@@ -57,17 +57,18 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
 
   @Override
   public void start(SpanContext spanContext) {
-    if (samplers.containsKey(spanContext.getTraceId())) {
-      return;
-    }
-
     ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
-    samplers.put(spanContext.getTraceId(), scheduler);
-    scheduler.scheduleAtFixedRate(
-        new StackTraceGatherer(spanContext.getTraceId(), Thread.currentThread().getId()),
-        SCHEDULER_INITIAL_DELAY,
-        samplingPeriod.toMillis(),
-        TimeUnit.MILLISECONDS);
+    ScheduledExecutorService mappedScheduler =
+        samplers.computeIfAbsent(spanContext.getTraceId(), id -> scheduler);
+    if (scheduler == mappedScheduler) {
+      scheduler.scheduleAtFixedRate(
+          new StackTraceGatherer(spanContext.getTraceId(), Thread.currentThread().getId()),
+          SCHEDULER_INITIAL_DELAY,
+          samplingPeriod.toMillis(),
+          TimeUnit.MILLISECONDS);
+    } else {
+      scheduler.shutdown();
+    }
   }
 
   @Override
--- a/profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
+++ b/profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
@@ -87,28 +87,29 @@ class ScheduledExecutorStackTraceSamplerTest {
     var scheduler = Executors.newScheduledThreadPool(2);
     var latch = new CountDownLatch(1);
     var traceId = idGenerator.generateTraceId();
+    var spanContext2 = randomSpanContext(traceId);
+    var spanContext1 = randomSpanContext(traceId);
 
-    var threadOne =
-        scheduler.schedule(
-            startSampling(randomSpanContext(traceId), latch), 0, TimeUnit.MILLISECONDS);
-    scheduler.schedule(startSampling(randomSpanContext(traceId), latch), 25, TimeUnit.MILLISECONDS);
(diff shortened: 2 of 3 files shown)


breedx-splk:
Resolves #2224

breedx-splk:
Resolves #2229

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
robsunday, line 1: For better test isolation `idGenerator`, `staging` and `sampler` should be instantiated in `@BeforeEach` annotated  method.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
```diff
@@ -80,33 +83,39 @@ void emptyStagingAreaAfterSamplingStops() {
...
-        });
+  private Callable<Thread> startSampling(SpanContext spanContext, CountDownLatch latch) {
+    return (() -> {
+      try {
+        System.out.println(
+            "Starting thread "
+                + Thread.currentThread().getName()
+                + ", id "
+                + Thread.currentThread().getId());
+        sampler.start(spanContext);
```
robsunday, line 112: sampler.stop(spanContext) is never called

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
```diff
@@ -80,33 +83,39 @@ void emptyStagingAreaAfterSamplingStops() {
...
-    var threadTwo = startAndStopSampler(randomSpanContext(traceId), latch);
 
-    threadOne.start();
-    threadTwo.start();
+    var threadOne =
+        scheduler.schedule(
+            startSampling(randomSpanContext(traceId), latch), 0, TimeUnit.MILLISECONDS);
+    scheduler.schedule(startSampling(randomSpanContext(traceId), latch), 25, TimeUnit.MILLISECONDS);
+
     await().atMost(HALF_SECOND).until(() -> staging.allStackTraces().size() > 5);
```
robsunday, line 96: Why 5?

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
```diff
@@ -80,33 +83,39 @@ void emptyStagingAreaAfterSamplingStops() {
...
     latch.countDown();
 
     var threadIds =
         staging.allStackTraces().stream().map(StackTrace::getThreadId).collect(Collectors.toSet());
```
robsunday, line 100: `staging` may not be initially empty in this test because its instance is reused among all test cases. That's why `staging` should be instantiated in `@BeforeEach` as I already mentioned

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
tduncan: There isn't anything special about the number 5, per say. The intent here is for multiple samples to be taken and for each thread to be running at the same time. Since the threads would potentially be on different sampling schedules it needs to be more than 2, but also not so high that the test must wait a long time to be certain the 2nd thread isn't being sampled. 5 seemed OK. I don't think a smaller number is a good idea but happy to increase it a bit.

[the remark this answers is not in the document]
tduncan: Sure, I can increase the wait time.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
tduncan: This is a misunderstanding about how JUnit isolates tests. By default a new instance of the test class is instantiated  for each test method (see [JUnit 5's Test Lifecycle](https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-instance-lifecycle)) so we can lean on this behavior to ensure that new instances of all test dependencies are also instantiated. We do have the option of changing the test lifecycle but we need to explicitly use the `@TestInstance(Lifecycle.PER_CLASS)` annotation. This is of course assuming we haven't overridden the default using the `junit.jupiter.testinstance.lifecycle.default` property which I don't see in the `build` check.

`@BeforeEach` setup methods are really best used to prime external resources (e.g. insert data in a database) that apply to every (or at least most) tests in the test class. Obviously we can instantiate test fields in setup methods but it doesn't add additional safety, only boilerplate.

If you've used TestNG in the past it explicitly does the opposite by reusing the test class instance therefore requiring the use of before and after test callback methods to manage state.

```java
public class LifecycleTest {
  private final List<UUID> uuids = new ArrayList<>();

  @Test
  void testOne() {
    uuids.add(UUID.randomUUID());
    System.out.println(uuids);
  }

  @Test
  void testTwo() {
    uuids.add(UUID.randomUUID());
    System.out.println(uuids);
  }
}

Output:
[3195ee77-46e6-4f79-ab3a-efc3f9816b47]
[75990fbf-23ae-4377-a0aa-da91fd5fa430]
```

```java
@TestInstance(TestInstance.Lifecycle.PER_CLASS)
public class LifecycleTest {
  private final List<UUID> uuids = new ArrayList<>();

  @Test
  void testOne() {
    uuids.add(UUID.randomUUID());
    System.out.println(uuids);
  }

  @Test
  void testTwo() {
    uuids.add(UUID.randomUUID());
    System.out.println(uuids);
  }
}

Output:
[f903466c-6037-4bb0-8232-f64b7995eb65]
[f903466c-6037-4bb0-8232-f64b7995eb65, 08203264-a72b-48f8-9de9-35327da0d3ff]
```

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
tduncan: See https://github.com/signalfx/splunk-otel-java/pull/2233#discussion_r1985548449

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
robsunday: Yeah, you are right. Too long break with Java coding...

laurit:
Test failure is still easily reproducible using a debugger.

tduncan:
> Test failure is still easily reproducible using a debugger.

I've updated the map writing to use `computeIfAbsent` which should protect against this race condition. It's worth noting, however, that I was unable to reproduce the failure without using a debugger despite running thousands of iterations.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
```diff
@@ -57,17 +57,18 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
...
+    ScheduledExecutorService mappedScheduler =
+        samplers.computeIfAbsent(spanContext.getTraceId(), id -> scheduler);
+    if (scheduler == mappedScheduler) {
+      scheduler.scheduleAtFixedRate(
+          new StackTraceGatherer(spanContext.getTraceId(), Thread.currentThread().getId()),
+          SCHEDULER_INITIAL_DELAY,
+          samplingPeriod.toMillis(),
+          TimeUnit.MILLISECONDS);
+    } else {
+      scheduler.shutdown();
```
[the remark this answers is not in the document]
tduncan: To ensure the thread allocated to this unused `ScheduledExecutorService` is released back to the JVM.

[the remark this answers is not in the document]
laurit: ok, makes sense. For some reason though that `scheduler` was what was already in the map. Did you consider writing this as
```java
    samplers.computeIfAbsent(spanContext.getTraceId(), traceId -> {
      ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
      scheduler.scheduleAtFixedRate(
          new StackTraceGatherer(traceId, Thread.currentThread().getId()),
          SCHEDULER_INITIAL_DELAY,
          samplingPeriod.toMillis(),
          TimeUnit.MILLISECONDS);
      return scheduler;
    });
```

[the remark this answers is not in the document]
tduncan: That is cleaner. Updated.

laurit (approved):

PR #2236, opened 2025-03-12
Pass through configured otlp headers for logs containing profiles.

breedx-splk:
This was a user request. They expect the profiling logs exporter to behave consistently with the other signals.

Seems like a reasonable ask, especially until/before we have native profiles ingest available.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/LogExporterBuilder.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/LogExporterBuilder.java
@@ -56,6 +56,10 @@ class LogExporterBuilder {
       ConfigProperties config, Supplier<OtlpHttpLogRecordExporterBuilder> makeBuilder) {
     OtlpHttpLogRecordExporterBuilder builder = makeBuilder.get();
     String ingestUrl = Configuration.getConfigUrl(config);
+
+    config.getMap("otel.exporter.otlp.headers").forEach(builder::addHeader);
+    config.getMap("otel.exporter.otlp.logs.headers").forEach(builder::addHeader);
+
     if (ingestUrl != null) {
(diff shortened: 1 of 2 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/LogExporterBuilder.java
```diff
@@ -56,6 +56,10 @@ static LogRecordExporter buildHttpExporter(
       ConfigProperties config, Supplier<OtlpHttpLogRecordExporterBuilder> makeBuilder) {
     OtlpHttpLogRecordExporterBuilder builder = makeBuilder.get();
     String ingestUrl = Configuration.getConfigUrl(config);
+
+    config.getMap("otel.exporter.otlp.headers").forEach(builder::addHeader);
+    config.getMap("otel.exporter.otlp.logs.headers").forEach(builder::addHeader);
```
laurit, line 61: imo this isn't ideal, we should follow the same logic as the sdk autoconfiguration uses. We could add
```
    OtlpConfigUtil.configureOtlpExporterBuilder(
        DATA_TYPE_LOGS,
        config,
        builder::setEndpoint,
        builder::addHeader,
        builder::setCompression,
        builder::setTimeout,
        builder::setTrustedCertificates,
        builder::setClientTls,
        builder::setRetryPolicy,
        builder::setMemoryMode);
```
to get the same conf as the autoconfigure uses and then override the endpoint and add the extra content type header.

breedx-splk: Yeah, good call. I'll make this change....I think it's more robust anyway.

laurit (approved):

PR #2237, opened 2025-03-13
Export Snapshot Profiling Stack Traces As OpenTelemetry Log Messages

tduncan:
This PR plugs in the log exporting functionality of the snapshot profiler which completes the end-to-end extension data flow. The full extension sequence is now:
1. Trace is selected for snapshotting
2. Start of root span of trace is detected
3. Profiling for trace thread starts
4. Collected stack traces are staged in the `AccumulatingStagingArea`
5. End of root span of trace is detected
6. Profiling of trace thread stops
7. Stack traces for trace are emptied from `StagingArea` to `StackTraceExporter`
8. Stack traces are converted into the `pprof` data format
9. Otel `LogRecordData` is created using `pprof` as log body
10. Otel `LogRecordData` is emitted to the configured Otel `LogRecordProcessor`

The configuration of the Otel `Logger` required adding an additional `AgentListener` to obtain the `Resource`, which in turn necessitated a 2-step initialization and an additional abstraction layer within `AccumulatingStagingArea`. 

The need to utilize the `AgentListener` callback also required further simulating of the agent startup process which was done by updating `OpenTelemetrySdkExtension` to support adding `AgentListeners` and calling into them prior to a test running. This did require `OpenTelemetrySdkExtension` to extend `AutoConfiguredOpenTelemetrySdk` and move into shadow package to actually implement the abstract methods along with renaming the `builder` function to `configure` so as to not conflict with `builder` function already defined.


profiler/build.gradle.kts
```diff
@@ -37,6 +37,7 @@ dependencies {
   annotationProcessor("com.google.auto.service:auto-service")
   compileOnly("com.google.auto.service:auto-service")
 
+  testImplementation(project(":custom"))
```
tduncan, line 40: I needed access to the `AutoConfigureUtil` class defined in this subproject. Used in `StackTraceExporterActivator` to access the `ConfigProperties` and `Resource` fields in the `AutoConfiguredOpenTelemetrySdk` when configuring the Otel `Logger`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -181,6 +181,10 @@ private static int getJavaVersion() {
     return Integer.parseInt(javaSpecVersion);
   }
 
+  public static boolean isSnapshotProfilingEnabled(ConfigProperties properties) {
+    return properties.getBoolean(CONFIG_KEY_ENABLE_SNAPSHOT_PROFILER, false);
+  }
```
tduncan, line 186: A 2nd use was added in `StackTraceExporterActivator` so I thought it made sense to roll this into the `Configuration` class.

profiler/src/main/java/com/splunk/opentelemetry/profiler/OtelLoggerFactory.java
tduncan, line 1: Adapted from code in `JfrActivator`. I intentionally didn't modify the existing profiling code to minimize the scope of this PR.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/Pprof.java
tduncan, line 1: Very similar to the existing `Pprof` class, mostly QoL/DX changes. I decided to port the code AS IN, duplications and all, due to timeline constraints. Once the extension functionality is in place my intention is to figure out how to unify the two slightly different strategies used by the continuous profiler and the snapshot profiler.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PprofTranslator.java
tduncan, line 1: Similar to `PprofCpuEventExporter` though slightly simpler due to guarantees that thread and trace information will be available.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/StackTraceExporterProvider.java
tduncan, line 1: This class acts a middleman between `AccumulatingStagingArea` and the `StackTraceExporter`. Its necessary because the `StagingArea` is instantiated during the SDK configuration phase (see updates in `SnapshotProfilingSdkCustomizer`) but the Otel `Logger` can't be configured until later in agent startup process (see newly added `StackTraceExporterActivator`).

To insulate the `AccumulatingStagingArea` from this 2-step initialization requirement it uses this `Supplier` to obtain a `StackTraceExporter`, always receiving a no-op implementation if the initialization process has yet to complete.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingLogExportingTest.java
tduncan, line 1: Component test of the extension's ability to export logs. Selects a trace for snapshotting, starts and stops a root span, and verifies that a log record lands in the configured Otel `LogRecordExporter`.

profiler/src/test/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkExtension.java
tduncan, line 1: The primary changes in this class is the ability to add attach `AgentListener` instances and simulating that portion of the agent startup process. To do this requires a `AutoConfiguredOpenTelemetrySdk` instance which is difficult to obtain on a per-test basis. The easiest approach was to subclass `AutoConfiguredOpenTelemetrySdk` which unfortunately required moving the entire JUnit extension into this shadow package so two of the abstract methods could be implemented.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
```diff
@@ -0,0 +1,135 @@
...
+  private Context createProfilingContext(List<StackTrace> stackTraces) {
+    String traceId = extractTraceId(stackTraces);
+    SpanContext spanContext =
+        ImmutableSpanContext.create(
+            traceId,
+            SpanId.getInvalid(),
+            TraceFlags.getDefault(),
+            TraceState.getDefault(),
+            false,
+            true);
```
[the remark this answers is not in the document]
tduncan: To which validation are you referring? My goal was to include the trace ID on the log message.

breedx-splk:
Building an implementation of `OpenTelemetry` for testing is wild. Without having read this code yet, my instinct is that it's a mistake or at least a misstep.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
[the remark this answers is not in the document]
laurit: See https://github.com/open-telemetry/opentelemetry-java/blob/9c1b9f609fe6ece64923c126990f57981b7171fc/api/all/src/main/java/io/opentelemetry/api/internal/ImmutableSpanContext.java#L67 The last `true` parameter to `ImmutableSpanContext.create` is named `skipIdValidation`. If validation was enabled then this would be deemed invalid because of the `SpanId.getInvalid()`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
```diff
@@ -97,7 +99,7 @@ private boolean includeTraceContextPropagator(Set<String> configuredPropagators)
     return configuredPropagators.isEmpty();
   }
 
-  private boolean snapshotProfilingEnabled(ConfigProperties config) {
-    return config.getBoolean(CONFIG_KEY_ENABLE_SNAPSHOT_PROFILER, false);
+  private boolean snapshotProfilingEnabled(ConfigProperties properties) {
```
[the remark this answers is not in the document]
tduncan: Addressed in a separate PR: https://github.com/signalfx/splunk-otel-java/pull/2243

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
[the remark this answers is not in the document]
tduncan: Taking a more critical look at this specifically I don't think its necessary to include the `SpanContext` on the log message. The Splunk backend doesn't do anything with it and it likely wouldn't make the conversion to the Otel profiling signal/format anyway.

Long story short: I'll remove this section entirely.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
```diff
@@ -0,0 +1,135 @@
...
+    Span span = Span.wrap(spanContext);
+    return span.storeInContext(Context.root());
+  }
+
+  private String extractTraceId(List<StackTrace> stackTraces) {
+    return stackTraces.stream().findFirst().map(StackTrace::getTraceId).orElse(null);
```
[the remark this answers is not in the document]
tduncan: See [my other comment](https://github.com/signalfx/splunk-otel-java/pull/2237#discussion_r2002815571). This code isn't necessary so I'll remove it.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/Pprof.java
```diff
@@ -0,0 +1,226 @@
...
+import java.util.Objects;
+import java.util.function.Consumer;
+
+/**
+ * Adapted from the Splunk OpenTelemetry profiler <a
+ * href=https://github.com/signalfx/splunk-otel-java/blob/main/profiler/src/main/java/com/splunk/opentelemetry/profiler/pprof/Pprof.java>here</a>,
+ * which is itself adapted from Google's Bazel build system <a
+ * href=https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/profiler/memory/AllocationTracker.java>here</a>.
+ */
+class Pprof {
```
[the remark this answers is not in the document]
tduncan: I tried to explain [here](https://github.com/signalfx/splunk-otel-java/pull/2237#discussion_r1994131648) and [here](https://github.com/signalfx/splunk-otel-java/pull/2237#issuecomment-2722468235) why I went this way.

The longer version is that I was tinkering for a few hours on how to reconcile the differences between the `StackTrace` defined in the `snapshot` package with the `StackTrace` defined in `StackTraceParser` used in `PprofCpuEventExporter`. The two data classes are very similar with the primary differences being the `osThreadId` field and how the stack frames are represented (`StackTraceLine` vs the built-in `StackTraceElement`). The approach I was exploring was to define `StackTrace` and `StackFrame` interfaces with the relevant accessor methods.

The next obstacle was around the building and serialization of the profiling data. `PprofCpuEventExporter` maintains a mutable `Pprof` instance that it resets after being flushed where as the design of the snapshot profiling extension does not maintain any mutable state during the log export process. The location of the serialization code also differs with it being found in the `Pprof` class itself vs in the `AsyncLogExporter` for snapshots.

There's also a minor DX change on the snapshot profiling `Pprof` version where the label creation and adding of labels to `Samples` is separated into two distinct operations and pushed up a level into the `PprofTranslator` as well the addition a few convenience methods to accept a `StackTraceElement`. Ex:

```java
// Always On
...
if (stackTrace.getThreadId() != 0) {
  pprof.addLabel(sample, THREAD_ID, stackTrace.getThreadId());
  pprof.addLabel(sample, THREAD_NAME, stackTrace.getThreadName());
}
...
for (StackTraceParser.StackTraceLine stl : stackTrace.getStackTraceLines()) {
  sample.addLocationId(
      pprof.getLocationId(
          stl.getLocation(), stl.getClassName(), stl.getMethod(), stl.getLineNumber()));
  pprof.incFrameCount();
}

// Snapshot Profiling
...
sample.addLabel(pprof.newLabel(THREAD_ID, stackTrace.getThreadId()));
sample.addLabel(pprof.newLabel(THREAD_NAME, stackTrace.getThreadName()));
...
for (StackTraceElement stackFrame : stackTrace.getStackFrames()) {
  sample.addLocationId(pprof.getLocationId(stackFrame));
  pprof.incFrameCount();
}
```

Finally, I also thought it was a better idea to tackle the unification of the two stack trace data models and their translation into the `pprof` data format separately so as to not unintentionally break the existing Always On profiling extension while the snapshot profiling extension is being ported.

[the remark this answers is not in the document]
tduncan: Also see [here](https://github.com/signalfx/splunk-otel-java/pull/2237#discussion_r1994127290) for another place where I intentionally did not modify the existing profiling code to minimize the scope of this PR.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
```diff
@@ -0,0 +1,53 @@
...
+
+package com.splunk.opentelemetry.profiler.snapshot;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Supplier;
+
+class AccumulatingStagingArea implements StagingArea {
```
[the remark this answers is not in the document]
tduncan: This sounds like a very good idea. Would you be against doing this as a followup (this PR is a partial blocker to adding span correlation)?

I don't think a time-based approach is necessarily the right approach however. What I'm envisioning is a `BufferingStagingArea` with some number of max stack traces before it automatically empties itself in addition to the forced empty upon detection of the entry span ending. I don't know what that number should be; maybe 20?

This would control how many stack traces are bundled into a single log record. The frequency of the log record exporting is separately controlled by the `LogRecordProcessor` configured in `OtelLoggerFactory`. Right now it wires up a `SimpleLogRecordProcessor` which I believe immediately exports the log record. We could use In the `BatchLogRecordProcessor` instead, though I've had issues getting this to work in the past.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/Pprof.java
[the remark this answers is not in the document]
laurit: why not something like https://github.com/tduncan/splunk-otel-java/pull/1

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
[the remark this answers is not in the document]
laurit: > Would you be against doing this as a followup

sure

> I don't think a time-based approach is necessarily the right approach however. What I'm envisioning is a BufferingStagingArea with some number of max stack traces before it automatically empties itself in addition to the forced empty upon detection of the entry span ending. I don't know what that number should be; maybe 20?

Flushing when there are no new stack traces in some time isn't that hard. When you write stacks from all ongoing traces to the same pprof then flushing on ending the entry span doesn't make sense. `20` is imo too low, `2000` or even larger would be better. I think that determining the batch size requires some experimentation.

> This would control how many stack traces are bundled into a single log record. The frequency of the log record exporting is separately controlled by the LogRecordProcessor configured in OtelLoggerFactory. Right now it wires up a SimpleLogRecordProcessor which I believe immediately exports the log record. We could use In the BatchLogRecordProcessor instead, though I've had issues getting this to work in the past.

Using `BatchLogRecordProcessor` here isn't ideal because pprof works best when you put lots of stacks into the same batch. When the same string is used multiple times then the shared string pool will help in reducing the transfer size.  We need to manage creating these batches ourself to not export too often so we really don't gain much from using the `BatchLogRecordProcessor`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/Pprof.java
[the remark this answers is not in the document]
tduncan: @laurit, thanks for doing this! I've merged that PR into the branch, added tests, and also added new configuration properties to allow control over both the stack depth and sampling period of the snapshot profiler.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.logging.Level;
+
+class AsyncStackTraceExporter implements StackTraceExporter {
+  private static final java.util.logging.Logger logger =
+      java.util.logging.Logger.getLogger(AsyncStackTraceExporter.class.getName());
+
+  private final ExecutorService executor = Executors.newSingleThreadScheduledExecutor();
```
laurit, line 33: I think you meant to use `newSingleThreadExecutor` here.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
```diff
@@ -0,0 +1,76 @@
...
+              stackTrace.getThreadName(),
+              stackTrace.getThreadState(),
+              stackTrace.getStackFrames(),
+              stackTrace.getTimestamp(),
+              stackTrace.getTraceId(),
+              null);
+        }
+        cpuEventExporter.flush();
+      } catch (Exception e) {
+        logger.log(Level.SEVERE, "an exception was thrown", e);
```
laurit, line 72: Perhaps change to something like `An exception was thrown while exporting profiling snapshots.`

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
```diff
@@ -0,0 +1,76 @@
...
+
+        for (StackTrace stackTrace : stackTraces) {
+          cpuEventExporter.export(
+              stackTrace.getThreadId(),
+              stackTrace.getThreadName(),
+              stackTrace.getThreadState(),
+              stackTrace.getStackFrames(),
+              stackTrace.getTimestamp(),
+              stackTrace.getTraceId(),
+              null);
```
laurit, line 68: I'm not sure how the profiling UI will handle the missing span id here. Perhaps it would be better to use the local root span id here?

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
```diff
@@ -108,19 +113,118 @@ void onlyTakeStackTraceSamplesForOneThreadPerTrace() {
...
 
-  private Runnable startSampling(
+  @Test
+  void includeTimestampOnStackTraces() {
+    var now = Instant.now();
+    var spanContext = randomSpanContext();
+
+    try {
+      sampler.start(spanContext);
+      await().atMost(HALF_SECOND).until(() -> !staging.allStackTraces().isEmpty());
```
laurit, line 123: it is best not to use short wait times like `atMost(HALF_SECOND)` as this can result in sporadic test failures

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
tduncan: Thanks. Updated!

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
tduncan: Updated!

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java
tduncan: I removed all of the explicit timeouts from these tests so the 10 second default is used.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AsyncStackTraceExporter.java
tduncan: The next PR adding span correlation will fill this detail in. To your concern a different set of UI screens is being created for snapshot profiling data and the two sources of callstacks will not overlap at any point in the UI.

laurit (approved):

PR #2249, opened 2025-03-25
Add smoke test for snapshot profiler

laurit:
Copied from the existing profiler smoke test. cc @tduncan

### DIFF
--- a/smoke-tests/src/test/java/com/splunk/opentelemetry/SnapshotProfilerSmokeTest.java
+++ b/smoke-tests/src/test/java/com/splunk/opentelemetry/SnapshotProfilerSmokeTest.java
@@ -127,7 +127,7 @@ public abstract class SnapshotProfilerSmokeTest {
   private void generateSomeSpans() throws Exception {
     logger.info("Generating some spans...");
     int port = containerManager.getTargetMappedPort(PETCLINIC_PORT);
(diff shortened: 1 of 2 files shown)


smoke-tests/src/test/java/com/splunk/opentelemetry/SnapshotProfilerSmokeTest.java
```diff
@@ -0,0 +1,173 @@
...
+                "-Dsplunk.profiler.logs-endpoint=http://collector:4318/v1/logs",
+                // uncomment to enable exporting traces
+                // "-Dotel.exporter.otlp.endpoint=http://collector:4318",
+                "-jar",
+                "/app/spring-petclinic-rest.jar"));
+
+    logger.info("Petclinic has been started.");
+
+    try {
+      generateSomeSpans();
```
tduncan, line 168: Given the random nature of selecting traces for snapshotting it may be necessary to continuously send requests. With this setup we should expect 2 traces to be selected, but there will definitely be iterations where 0 of the 200 traces are selected.

laurit: I increased the selection rate and iteration count.

tduncan (approved):

robsunday (approved):

PR #2251, opened 2025-03-27
Port span correlation

tduncan:
This PR add span correlation snapshot profiling stack traces. Span correlation is done by intercepting OpenTelemetry context changes via a `ContextStorage` that maps trace IDs to `SpanContext`s as new `Context`s are attached and detached.


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/InterceptingContextStorageSpanTrackingActivator.java
tduncan, line 1: This class primarily exists to ease testing since `ContextStorage.addWrapper` must be called prior to the `ContextStorage` being initialized. Putting this behind a `Supplier` offers tests a way to hook in (via the `SettableContextStorageProvider`) without trying to coordinate the `ContextStorage` initialization across every single test.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SpanTrackerProvider.java
tduncan, line 1: Early designs had `ActiveSpanTracker` as a singleton object. That worked but was quite intricate to use correctly. Pushing the singleton behavior into a provider removes a lot of the complexities with a minor inconvenience in `ScheduledExecutorStackTraceSampler` where the `SpanTracker` is needed.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
```diff
@@ -0,0 +1,75 @@
...
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextStorage;
+import io.opentelemetry.context.Scope;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import javax.annotation.Nullable;
+
+class ActiveSpanTracker implements ContextStorage, SpanTracker {
+  private final ConcurrentMap<String, SpanContext> activeSpans = new ConcurrentHashMap<>();
```
laurit, line 30: Are the traces ever removed from this map?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
```diff
@@ -0,0 +1,75 @@
...
+  @Override
+  public Scope attach(Context toAttach) {
+    Scope scope = delegate.attach(toAttach);
+    SpanContext spanContext = Span.fromContext(toAttach).getSpanContext();
+    if (doNotTrack(spanContext)) {
+      return scope;
+    }
+
+    String traceId = spanContext.getTraceId();
+    SpanContext current = activeSpans.get(traceId);
```
laurit, line 49: I think you would be better off rethinking this. Same trace could run on multiple threads even concurrently which could mess up the accounting. Perhaps it would make more sense to track the current trace/span per thread, then you wouldn't need to worry about concurrency etc. The stack trace you wish to associate with the trace/span is for a particular thread anyway. In https://github.com/signalfx/splunk-otel-java/pull/1306/files#diff-1c0fefdc9094d5abda5ae666e64d5955141bac620a75a4f0f16668d379c21314 there is my attempt at associating trace/span with a thread. Since it uses `ThreadLocal` (as does https://github.com/signalfx/splunk-otel-java/blob/main/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrContextStorage.java) it may cause issues with virtual threads. That code tracks all threads, for snapshot profiler you know which threads you need so you'll have to see whether it makes sense to track only the threads that are being profiled.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
tduncan: Yes. In the scope closing block `computeIfPresent` will remove a key from the map if the resolved value is `null`. There is also a test to confirm this behavior in `ActiveSpanTrackerTest`.

```java
  @Test
  void noActiveSpanForTraceAfterSpansScopeIsClosed() {
    var span = FakeSpan.newSpan(Snapshotting.spanContext());
    var spanContext = span.getSpanContext();
    var context = Context.root().with(span);
    registry.register(spanContext);

    var scope = spanTracker.attach(context);
    scope.close();

    assertEquals(Optional.empty(), spanTracker.getActiveSpan(spanContext.getTraceId()));
  }
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
tduncan: Yes, you are right. This really needs to be associating threads with spans and outsourcing the trace -> thread -> span correlation to another part of the system.

I've updated it to map threadId -> `SpanContext`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
laurit: thanks

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
```diff
@@ -0,0 +1,75 @@
...
+
+    SpanContext previous = activeSpansForThreads.put(threadId, spanContext);
+    return () -> {
+      activeSpansForThreads.computeIfPresent(threadId, (id, sc) -> previous);
+      scope.close();
+    };
+  }
+
+  private boolean doNotTrack(SpanContext spanContext) {
+    return !spanContext.isSampled() || !registry.isRegistered(spanContext);
```
laurit, line 62: perhaps should also add `!spanContext.isValid()`

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
```diff
@@ -0,0 +1,75 @@
...
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextStorage;
+import io.opentelemetry.context.Scope;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import javax.annotation.Nullable;
+
+class ActiveSpanTracker implements ContextStorage, SpanTracker {
+  private final ConcurrentMap<Long, SpanContext> activeSpansForThreads = new ConcurrentHashMap<>();
```
laurit, line 30: An alternative would be to use `Cache<Thread, SpanContext> activeSpansForThreads = Cache.weak()` which is a concurrent map with weak keys. This would offer automatic cleanup when a scope wasn't closed on a thread before the thread terminated. It does not have `computeIfPresent` but that should be easy to handle. Idk whether it is worth it.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTrackerTest.java
```diff
@@ -0,0 +1,307 @@
...
+    var executor = Executors.newFixedThreadPool(2);
+    try (var scope1 = executor.submit(attach(span1)).get();
+        var scope2 = executor.submit(attach(span2)).get()) {
+      assertEquals(
+          Optional.of(span1.getSpanContext()), spanTracker.getActiveSpan(scope1.getThreadId()));
+      assertEquals(
+          Optional.of(span2.getSpanContext()), spanTracker.getActiveSpan(scope2.getThreadId()));
+    } finally {
+      executor.shutdown();
+    }
```
laurit, line 130: I found it hard to reason why both actions should be executed on separate threads. Looks line an implementation detail of ThreadPoolExecutor. I guess it is fine as long as it works reliably.

tduncan: `ActiveSpanTracker` records the thread ID using `Thread.currentThread().getId()` so if both spans are attached to the OpenTelemetry `Context` on the same thread only one would be considered the active thread.

laurit: Let me rephrase. If you change `Executors.newFixedThreadPool(2)` to `Executors.newFixedThreadPool(1)` the test will fail. As far as I can tell the fact that when using `Executors.newFixedThreadPool(2)` both of the actions run on separate threads is an implementation detail of `ThreadPoolExecutor` not a specified behavior.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
tduncan: I've added a test for invalid spans but it did not require adding a explicit check. I think the `isSampled` check doubles as a check for `isValid`. I'll leave the test to make that scenario documented.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
tduncan: It took a bit of trial and error but I was able to write to expose this weakness.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
```diff
@@ -88,13 +91,19 @@ public void run() {
       try {
         Instant now = Instant.now();
         ThreadInfo threadInfo = threadMXBean.getThreadInfo(threadId, Integer.MAX_VALUE);
-        StackTrace stackTrace = StackTrace.from(now, samplingPeriod, traceId, threadInfo);
+        SpanContext spanContext = retrieveActiveSpan(threadId);
```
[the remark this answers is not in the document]
tduncan: Theoretically, yes, but in practice I would suspect the risk is likely quite low. Low enough that I don't think we would want to block the trace thread(s) while the every time the snapshot profiler wants to take sample.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTrackerTest.java
tduncan: I understand now. You are technically correct, there aren't any documented guarantees that the two operations will actually be performed on separate threads. I've used this technique quite a few times over the years and it hasn't caused issues but I can add an explicit release step so the test ensures that the first submitted operation doesn't end to quickly to ensure that both threads in the thread pool much be used.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
laurit: I missed that the trace registry only checks for the trace id.

laurit: https://github.com/signalfx/splunk-otel-java/blob/7bb828ab7599946757b3f8d8abbcfb8ddc5db398/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceRegistry.java#L24 could be a set as the map value is not used

PR #2256, opened 2025-04-02
Update TraceRegistry To Use A Set

tduncan:
This PR modifies `TraceRegistry` to use a `Set` instead of a `Map`. As noted by @laurit [here](https://github.com/signalfx/splunk-otel-java/pull/2251#discussion_r2024952775) only the keys of the `Map` are used.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceRegistry.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceRegistry.java
@@ -17,18 +17,18 @@
 package com.splunk.opentelemetry.profiler.snapshot;
 
 import io.opentelemetry.api.trace.SpanContext;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.Set;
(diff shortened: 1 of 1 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceRegistry.java
```diff
@@ -17,18 +17,18 @@
...
 
 import io.opentelemetry.api.trace.SpanContext;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 
 class TraceRegistry {
-  private final Map<String, SpanContext> traceIds = new ConcurrentHashMap<>();
+  private final Set<String> traceIds = new CopyOnWriteArraySet<>();
```
laurit, line 24: I'd use `private final Set<String> traceIds = Collections.newSetFromMap(new ConcurrentHashMap<>());`

laurit (approved):

PR #2271, opened 2025-04-15
Prepare release 2.15.0

laurit:


breedx-splk (approved):
Thanks for taking it on.

PR #2276, opened 2025-04-16
Take a snapshot sample on spanEnd

lo-jason:
Take additional sample when triggered for end of Span
- add unit tests


lo-jason:
I have read the CLA Document and I hereby sign the CLA

PR #2277, opened 2025-04-16
Take Final Stack Trace Sample When Trace Profiling Stops

tduncan:
This PR adds a final stack trace sample taken when stack trace sampling is stopped for a given trace. Upon stopping sampling for a given trace ID a find stack trace sample will be taken on demand with a duration matching more closely with the amount of time elapsed between when the previous sample was taken and when the service entry span ends.

In addition the mechanism by which the stack trace duration is calculated has been updated to compute the value every iteration rather than assume the configured sampling duration. This has the effect that at sampling start and stop a value likely much smaller than the configured sampling rate will be calculated, and normal executions of the sampler will report computed values rather than an assumed value of the sampling period.

Note: This PR and #2276 are addressing the same issue.


lo-jason (approved):
looks good. I had comment on slack https://splunk.slack.com/archives/C07MYK71B55/p1744845504377729?thread_ts=1744835907.680119&cid=C07MYK71B55

callgraph executor ending early is a bug we should look to address before release as well.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
```diff
@@ -54,52 +53,68 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
...
-          ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
-          scheduler.scheduleAtFixedRate(
-              new StackTraceGatherer(
-                  samplingPeriod, spanContext.getTraceId(), Thread.currentThread()),
-              SCHEDULER_INITIAL_DELAY,
-              samplingPeriod.toMillis(),
-              TimeUnit.MILLISECONDS);
-          return scheduler;
-        });
+        traceId -> new ThreadSampler(traceId, Thread.currentThread(), samplingPeriod));
```
breedx-splk, line 56: Why is this change in here?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
```diff
@@ -54,52 +53,68 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
...
+            StackTrace.from(Instant.now(), samplingPeriod, threadInfo, traceId, spanId);
         stagingArea.stage(traceId, stackTrace);
       } catch (Exception e) {
         logger.log(Level.SEVERE, e, samplerErrorMessage(traceId, thread.getId()));
+      } finally {
+        timestampNanos = currentSampleTimestamp;
       }
     }
 
+    private Duration samplingPeriod(long fromNanos, long toNanos) {
```
breedx-splk, line 114: I don't think this method provides too much -- I'd just inline it.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/InMemoryStagingArea.java
```diff
@@ -16,38 +16,85 @@
...
           }
-          stackTraces.add(stackTrace);
           return stackTraces;
         });
   }
 
   @Override
   public void empty(String traceId) {
-    stackTraces.remove(traceId);
+    stackTraces.get(traceId).empty();
```
breedx-splk, line 51: does this NPE if not found?

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/InMemoryStagingArea.java
```diff
@@ -16,38 +16,85 @@
...
 
   public List<StackTrace> allStackTraces() {
-    return stackTraces.values().stream().flatMap(List::stream).collect(Collectors.toList());
+    return stackTraces.values().stream()
+        .map(StagedStackTraces::stackTraces)
+        .flatMap(List::stream)
+        .collect(Collectors.toList());
+  }
+
+  static class StagedStackTraces extends AbstractList<StackTrace> {
```
breedx-splk, line 69: This feels bad to me.  It's also just a strange inner class that can potentially claim to be "emptied" even though it has content in the list and can have content in the list but still be "emptied". I don't follow. Even if you do still somehow want to keep this, it needs to be private.

tduncan: It can't be `private` as it is used in [ScheduledExecutorStackTraceSamplerTest](profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSamplerTest.java).

Testing the inclusion of a unscheduled thread snapshot being taken when trace sampling is stopped caused contradictory requirements of visibility into the `StagingArea`. Right now the expectations is that the `StagingArea` will be emptied when sampling stops and previously this was confirmed by checking that the `List` here was indeed an empty list. Now we have a new requirement that a final thread snapshot is taken when sampling is stopped, but because the `StagingArea` is also emptied there wasn't a reliable way to confirm an additional `StackTrace` instance was staged. This class is the compromise a came up with; it records the staged `StackTraces` (which is what most tests care about) as well as whether the `empty` method was called for a specific trace ID.

Also part of the context, and not at all possible to know from this PR, is a change that I have nearly ready (its partially blocked by 2262) to completely remove this part of the design. Instead of staging and emptying by trace ID the `StagingArea` will keep all `StackTrace`s in a single bucket and export all of the at the same time periodically rendering this class and the test that relies on it unnecessary.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
tduncan: The creation of and scheduling of `StackTraceGatherer` instances has moved into the construction of the new `ThreadSampler` which now assumes responsibility of managing thread profiling mechanism, including stopping the scheduled sampling and taking the final on demand sample when trace profiling is stopped.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/InMemoryStagingArea.java
tduncan: Yes, it would.

breedx-splk: Should we guard against that?

tduncan: Yes. The latest commit ensures a non-`null` value will always be returned.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
```diff
@@ -54,52 +53,68 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
...
+      sampler.stop();
     }
+
     stagingArea.empty(spanContext.getTraceId());
   }
 
-  class StackTraceGatherer implements Runnable {
-    private final Duration samplingPeriod;
+  private class ThreadSampler {
+    private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
```
laurit, line 71: imo it would be better to use daemon threads like in https://github.com/signalfx/splunk-otel-java/blob/main/profiler/src/main/java/com/splunk/opentelemetry/profiler/util/HelpfulExecutors.java

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
```diff
@@ -54,52 +53,68 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
...
+
+    private ThreadSampler(String traceId, Thread thread, Duration period) {
+      gatherer = new StackTraceGatherer(traceId, thread, System.nanoTime());
+      scheduler.scheduleAtFixedRate(
+          gatherer, SCHEDULER_INITIAL_DELAY, period.toMillis(), TimeUnit.MILLISECONDS);
+    }
+
+    void stop() {
+      scheduler.shutdown();
+      gatherer.run();
```
laurit, line 82: I don't get it why you'd want to get a sample on end. That sample doesn't have any application code running, what use does it have?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
```diff
@@ -54,52 +53,68 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
...
     }
 
     @Override
     public void run() {
+      long currentSampleTimestamp = System.nanoTime();
       try {
-        Instant now = Instant.now();
         ThreadInfo threadInfo = threadMXBean.getThreadInfo(thread.getId(), Integer.MAX_VALUE);
-        SpanContext spanContext = retrieveActiveSpan(thread);
+        Duration samplingPeriod = samplingPeriod(timestampNanos, currentSampleTimestamp);
```
laurit, line 102: I have heard that the period isn't really used for snapshot profiling events, you sure it is worth it to compute exact values? Secondly since this code is called from the scheduled executor and also manually on stop you have a bit of race possibility here. Wouldn't be surprised when with debugger you could create a situation where the computed period is negative.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/InMemoryStagingArea.java
```diff
@@ -16,38 +16,85 @@
...
           }
-          stackTraces.add(stackTrace);
           return stackTraces;
         });
   }
 
   @Override
   public void empty(String traceId) {
-    stackTraces.remove(traceId);
+    stackTraces.getOrDefault(traceId, new StagedStackTraces()).empty();
```
laurit, line 51: How are things removed from the `stackTraces` map now?

tduncan: For this test implementation they aren't. However the lifespan of an `InMemoryStagingArea` instance is only that of a single test so the entire `Map` will be made available for GC as each test has its own instance to work with.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
tduncan: Previously the sampling period was ignored (see changes in `AsyncStackTraceExporter`). Now the sampling period included in the profiling log message will be the one computed by the `StackTraceSampler`. This data is eventually stored in the profiling backend and used to calculated the estimated total time each stack frame was running.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
tduncan: It helps report more accurate estimated timing information. Prior to this change all the trace profiles would have timing data that was some multiple of 20 because every sample was submitted with a 20ms duration, and would almost always be more than 20ms longer than the reported span duration.

Computing each sample period and taking a final sample greatly reduces the difference between the report span duration and the reported profiling duration. They are still not the same, however. In fact with this change the profiling duration is consistently undercounting the duration by a few milliseconds when compared to the duration of the entry span duration.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/InMemoryStagingArea.java
laurit: didn't notice this was a test class

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/InMemoryStagingArea.java
breedx-splk: We tend to put a comment or use `@VisibleForTesting` so that the reader knows why things are visible.

tduncan: Sure, I'm familiar with that convention for production code, though this class exists in the `test` directory and is only available to tests (specifically those in the `com.splunk.opentelemetry.profiler.snapshot` package).

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
laurit: Are you sure that using the timing data this way makes sense at all for snapshot profiler? A stack sample proves that a method was executed at given time. It does not reveal anything about how long the method was executing. I get the feeling that you are deriving arbitrary meaning here for the period. As far as I remember profilers usually show percentages, like this method was observed in 52% of samples.
I guess the `SCHEDULER_INITIAL_DELAY` is set to `0` for similar reasons.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
```diff
@@ -54,52 +53,68 @@ class ScheduledExecutorStackTraceSampler implements StackTraceSampler {
...
+    void stop() {
+      scheduler.shutdown();
+      gatherer.run();
+    }
+  }
+
+  private class StackTraceGatherer implements Runnable {
     private final String traceId;
     private final Thread thread;
+    private long timestampNanos;
```
lo-jason, line 89: timestampnanos is accessed from multiple threads so should be made volatile

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
lo-jason: +1 @laurit this timestampnanos needs to be threadsafe

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
tduncan: Address in new PR I just opened: #2283

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
laurit: Imo the timestampnanos could just be the tip of the iceberg here. `ThreadSampler#shutdown` calls
```
scheduler.shutdown();
gatherer.run();
```
`ExecutorService.shutdown` javadoc says
> Initiates an orderly shutdown in which previously submitted tasks are executed, but no new tasks will be accepted. Invocation has no additional effect if already shut down.
This method does not wait for previously submitted tasks to complete execution. Use awaitTermination to do that.

I didn't try this out but hypothetically assuming that the executor was already executing `gatherer.run` when `shutdown` was called `gatherer.run` may run in parallel from the executor and from the explicit `gatherer.run()` call. Similarly `stagingArea.empty(spanContext.getTraceId())` called from `ScheduledExecutorStackTraceSampler#stop` could run in parallel with `gatherer.run` from the executor with some unlucky timing. I'd guess that when `gatherer.run` is run in parallel the previous timestamp computations could be off, also `AccumulatingStagingArea` internally uses `ArrayList` that probably could be corrupted with concurrent mutations.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
laurit: > This is the same strategy used for classic java profilers. Unless methods are being instrumented directly then method timing is extrapolated the time based on the samples that it receives.

The difference is that a classic profiler has usually lots of samples. Here even if you somehow combined samples from multiple requests to the same endpoint it still wouldn't help because all the samples are biased because they are taken at 10ms intervals.

> This extrapolation does lead to somewhat arbitrary inaccuracies

If you already know that this is not going to work correctly then perhaps you should considered alternatives. Whatever we choose to implement now is what we are going to be stuck with for the foreseeable future. If you believe that taking an extra sample at the start and end of the trace is the best option then you will need to get this into the gdi spec along with whatever magic you need to do to compute the period.
Imo it would be nice to see you moving from having a separate scheduler per profiled trace into something more standard like having a single scheduler that drives the sampling for all profiled traces.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
lo-jason: +1 accumulatingstagingarea needs to be threadsafe. Internal use of ArrayList won't work.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/InMemoryStagingArea.java
lo-jason: @tduncan is another option is to use a mock stagingarea. That is what i did in https://github.com/signalfx/splunk-otel-java/pull/2276/

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
lo-jason: Arbitrary inaccuracy is too harsh, the main point is that the strategy is consistent. We always say for the time since last sample it is spent in the method that we see. We are looking at smaller time periods here, but still it is for the user to understand approximate time taken.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
tduncan: Thread-safety of `AccumulatingStagingArea` addressed in #2288.

laurit (approved):

PR #2278, opened 2025-04-17
Publish CSA bundle

breedx-splk:
This adds a module that will publish a second jar file that contains the Cisco Secure Application extension bundled inside `splunk-otel-java`. The output jar is named `splunk-otel-javaagent-csa-<version>.jar`.

### DIFF
--- a/agent-csa-bundle/build.gradle.kts
+++ b/agent-csa-bundle/build.gradle.kts
@@ -1,0 +1,82 @@
+
+plugins {
+  id("maven-publish")
+  id("signing")
+}
+
+dependencies {
+  implementation(project(":agent"))
+}
+
+// This should be updated for every CSA release
+val csaVersion = "25.3.0-1321"
+base.archivesName.set("splunk-otel-javaagent-csa")
+
+tasks {
+  val bundleSecureApp by registering(Exec::class) {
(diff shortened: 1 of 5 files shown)


agent-csa-bundle/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+// This should be updated for every CSA release
+val csaVersion = "25.3.0-1321"
+base.archivesName.set("splunk-otel-javaagent-csa")
+
+tasks {
+  val bundleSecureApp by registering(Exec::class) {
+    description = "Bundle the Cisco SecureApp with splunk-otel-java"
+    dependsOn(":agent:assemble")
+    isIgnoreExitValue = false
+    commandLine("./build-agent-csa-fat-jar.sh", version, csaVersion)
```
laurit, line 20: invoking an shell script is an unusual choice, did you consider doing whatever the shell script does with gradle?

laurit (approved):

agent-csa-bundle/build.gradle.kts
breedx-splk: Thanks, yeah, I concede that this isn't the greatest. We do resort to scripts in a number of other places, but I think we could probably figure this out as pure gradle. Will take that as a follow-up item.

PR #2280, opened 2025-04-21
Only Stop Trace Profiling When Same Span Ends

tduncan:
This PR closes a bug that could cause the snapshot profiling of a trace to be stopped before the thread being profiled has actually stopped processing that trace. This is possible when multiple API calls are made to a service simultaneously the a thread that did not start the profiling is the first thread to finish processing its request.


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ScheduledExecutorStackTraceSampler.java
tduncan, line 1: The first draft mapped `SpanContext` to `ScheduledExecutorService` but that conflicted with the intent to only allow a single thread for a trace to be profiled. Settled on retaining the `SpanContext` so it can be verified that the same span is stopping the profiling.

laurit (approved):

PR #2282, opened 2025-04-22
Classify metrics in the metadata

laurit:
Classify metrics as either bundled or custom.

### DIFF
--- a/metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
+++ b/metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
@@ -3630,15 +3630,15 @@ public class MetadataGenerator {
     instrumentations.add(
         splunkInstrumentation("jvm-metrics.splunk")
             .component("Java Platform", null)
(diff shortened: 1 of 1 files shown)


metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -3626,15 +3630,15 @@ Some instrumentations (e.g. Spring Web MVC instrumentation) produce SpanKind.Int
     instrumentations.add(
         splunkInstrumentation("jvm-metrics.splunk")
             .component("Java Platform", null)
-            .metric(
+            .customMetric(
```
laurit, line 3633: @breedx-splk it seems like memory profiling metrics are currently custom

breedx-splk: I have no idea, but that's fine. Classification shouldn't be an instrumentation concern anyway.

breedx-splk (approved):

PR #2283, opened 2025-04-22
Use Daemon Threads In Trace Snapshot Profiler

tduncan:
This PR updates the `ExecutorService` and `ScheduledExecutorService` instances in the trace snapshot profiler to use implementations that create daemon threads.


tduncan:
Thoughts on renaming `HelpfulExecutors` to more clearly indicate it provides `ExecutorService`s and `ScheduledExecutorService`s that use daemon threads?

laurit:
> Thoughts on renaming `HelpfulExecutors` to more clearly indicate it provides `ExecutorService`s and `ScheduledExecutorService`s that use daemon threads?

Sure, I have nothing agains that. I think that @breedx-splk came up with that name.

laurit (approved):

PR #2286, opened 2025-04-23
Update csa version and ensure newline in extension service file

breedx-splk:

PR #2287, opened 2025-04-23
Switch CSA bundling to pure gradle

breedx-splk:
Supersedes #2286.

This rebuilds the CSA packaging as a native gradle build instead of hybrid gradle+shell. Thanks @laurit for the push.


laurit (approved):
For some reason I thought that this would be simpler, but I guess that was just me being foolish. I tried to improve this a bit in https://github.com/breedx-splk/splunk-otel-java/pull/102

agent-csa-bundle/build.gradle.kts
```diff
@@ -1,32 +1,132 @@
...
+    url = uri("https://github.com/")
+    metadataSources {
+      artifact()
+    }
+    patternLayout {
+      ivy("[organisation]/[module]/releases/download/[revision]/[artifact].[ext]")
+      artifact("[organisation]/[module]/releases/download/[revision]/[artifact].[ext]")
+    }
+  }
+}
```
laurit, line 34: amazing

agent-csa-bundle/build.gradle.kts
```diff
@@ -1,32 +1,132 @@
...
+
 tasks {
-  val bundleSecureApp by registering(Exec::class) {
-    description = "Bundle the Cisco SecureApp with splunk-otel-java"
-    dependsOn(":agent:assemble")
-    isIgnoreExitValue = false
-    commandLine("./build-agent-csa-fat-jar.sh", version, csaVersion)
-    outputs.file(File("build/splunk-otel-javaagent-csa-$version.jar"))
+
+  // This exists purely as a hack to get the extension jar into our build dir
```
laurit, line 48: I have also been struggling with this. I think the problem is in our mental model where we think that how hard could a simple copy be. From gradle perspective I think archives and files with directories are the same thing so instead of copy we have to tell it to take the files from this archive and place them into another archive. We can hope that internally it is smart enough to somehow translate it to just copying the archive.

PR #2288, opened 2025-04-23
Make AccumulatingStagingArea Thread Safe

tduncan:
Stores staged `StackTrace` instances in a thread-safe `Queue` while exporting a "view" of that `Queue` when `empty` is called.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
@@ -16,7 +16,6 @@
 
 package com.splunk.opentelemetry.profiler.snapshot;
 
-import java.util.ArrayList;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -48,7 +47,7 @@ class AccumulatingStagingArea implements StagingArea {
   public void empty(String traceId) {
     Queue<StackTrace> stackTraces = this.stackTraces.remove(traceId);
     if (stackTraces != null) {
-      exporter.get().export(new ArrayList<>(stackTraces));
+      exporter.get().export(stackTraces);
     }
   }
 }
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
@@ -17,13 +17,14 @@
 package com.splunk.opentelemetry.profiler.snapshot;
 
 import java.util.ArrayList;
-import java.util.List;
+import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Supplier;
 
 class AccumulatingStagingArea implements StagingArea {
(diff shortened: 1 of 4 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
```diff
@@ -45,9 +46,9 @@ public void stage(String traceId, StackTrace stackTrace) {
 
   @Override
   public void empty(String traceId) {
-    List<StackTrace> stackTraces = this.stackTraces.remove(traceId);
+    Queue<StackTrace> stackTraces = this.stackTraces.remove(traceId);
     if (stackTraces != null) {
-      exporter.get().export(stackTraces);
+      exporter.get().export(new ArrayList<>(stackTraces));
```
laurit, line 51: if you made the `StackTraceExporter` accept a `Collection` than you could skip the copy.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
```diff
@@ -36,7 +37,7 @@ public void stage(String traceId, StackTrace stackTrace) {
         traceId,
         (id, stackTraces) -> {
           if (stackTraces == null) {
-            stackTraces = new ArrayList<>();
+            stackTraces = new ConcurrentLinkedQueue<>();
```
laurit, line 40: When `empty` is called before the last `stage` from the schedule then you will end up leaking a map entry.

lo-jason (changes requested):
I see two issues with this class. Agree with Lauri that the leaking of map objects is an issue, as is needing to access 2 concurrent datastructures to add samples. Wouldn't we be better off using ScheduledExecutorStackTraceSampler as a stagingArea since that is already per Thread and would solve above two issues?

If there is a concern in future we move to multi-threaded stack traces then the solution should be implemented in a way that won't make calling stage so expensive.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/AccumulatingStagingArea.java
tduncan: @laurit yes, though not a thread-safety issue and not introduced by this PR.

tduncan:
> I see two issues with this class. Agree with Lauri that the leaking of map objects is an issue, as is needing to access 2 concurrent datastructures to add samples. Wouldn't we be better off using ScheduledExecutorStackTraceSampler as a stagingArea since that is already per Thread and would solve above two issues?
> 
> If there is a concern in future we move to multi-threaded stack traces then the solution should be implemented in a way that won't make calling stage so expensive.

@lo-jason The primary purpose of this PR is to both address concerns raised in #2277 and #2262 without the scope of those PRs growing beyond the specific goal of each. In both of those PRs the thread-safety problems of `AccumulatingStagingArea` were brought up, which I believe is addressed.

Further #2262 is preventing the replacement of `AccumulatingStagingArea` with an implementation that groups all `StackTrace`s regardless of associated trace ID and periodically exports them as one batch (see [here](https://github.com/tduncan/splunk-otel-java/blob/batch-export-snapshot-profiling-stacktraces/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java)). That implementation is ready to submit as a PR accept for a single test that fails sporadically because the `StagingArea` isn't closed (hence being blocked by #2262).

I'm reluctant to expend too much effort perfecting this implementation when I intend to immediately replace it (pending the merging of #2262)

laurit (approved):

PR #2290, opened 2025-04-24
nocode docs: fix typos and add stern warning

johnbley:
Fix a few syntax typos in the nocode docs and add a stern warning that you should know what you're doing or things will break.

### DIFF
--- a/instrumentation/nocode/README.md
+++ b/instrumentation/nocode/README.md
@@ -4,7 +4,7 @@ Stability: under active development; breaking changes can occur
 
 Please don't use this if you have the ability to edit the code being instrumented.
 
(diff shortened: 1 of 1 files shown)


instrumentation/nocode/README.md
```diff
@@ -4,6 +4,16 @@ Stability: under active development; breaking changes can occur
 
 Please don't use this if you have the ability to edit the code being instrumented.
 
+By using this feature you are essentially changing your application's code.  Doing so without understanding
```
laurit, line 7:
```suggestion
By using this feature you are essentially changing your application's code. Doing so without understanding
```

laurit (approved):

PR #2298, opened 2025-04-28
Periodically Export Stacktraces From StagingArea

tduncan:
This PR swaps out the `AccumulatingStagingArea` for a `StagingArea` implementation that will automatically export empty itself when 1) a certain amount of has passed or 2) it has reached its capacity. Both the time interval between exports and capacity are configurable.


profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -81,6 +81,15 @@ public class Configuration implements AutoConfigurationCustomizerProvider {
...
   private static final Duration DEFAULT_SNAPSHOT_PROFILER_SAMPLING_INTERVAL = Duration.ofMillis(10);
 
+  private static final String CONFIG_KEY_SNAPSHOT_PROFILER_STAGING_EMPTY_INTERVAL =
+      "splunk.snapshot.profiler.staging.empty.interval";
+  private static final Duration DEFAULT_SNAPSHOT_PROFILER_STAGING_EMPTY_INTERVAL =
+      Duration.ofSeconds(5);
+
+  private static final String CONFIG_KEY_SNAPSHOT_PROFILER_STAGING_CAPACITY =
+      "splunk.snapshot.profiler.staging.capacity";
+  private static final int DEFAULT_SNAPSHOT_PROFILER_STAGING_CAPACITY = 2000;
```
tduncan, line 91: I don't have a reason for the chosen defaults. Happy to change them.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/StackTrace.java
tduncan, line 1: Auto-generated equals and hashCode so `StackTrace` can be safely used in a `Set`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java
```diff
@@ -0,0 +1,88 @@
...
+        this::empty, emptyDuration.toMillis(), emptyDuration.toMillis(), TimeUnit.MILLISECONDS);
+  }
+
+  @Override
+  public void stage(StackTrace stackTrace) {
+    if (closed) {
+      return;
+    }
+    stackTraces.add(stackTrace);
+    if (stackTraces.size() >= capacity) {
```
tduncan, line 53: Its possible that multiple threads could trigger an `empty` at roughly the same time but I think that is OK. I don't think its a huge deal the `stackTraces` collection is exactly the expected size when exporting, nor do I think its much of a problem if a 2nd export is done with a very small number of `StackTrace`s.

I suppose we could lock this down but I don't think the complexity of coordinating all of the potential threads in this way to super helpful. I think its more important that the `StackTrace` be staged and control returned to the caller ASAP.

laurit: Lets assume that 2 threads call `stage` at the same time, both observe `stackTraces.size() >= capacity` and call `empty` there both threads could copy the `stackTraces` collection and end up exporting the same data twice.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java
```diff
@@ -0,0 +1,88 @@
...
+  private void stopScheduledExecutor() {
+    try {
+      scheduler.shutdown();
+      if (!scheduler.awaitTermination(1, TimeUnit.SECONDS)) {
+        scheduler.shutdownNow();
+      }
+    } catch (InterruptedException e) {
+      scheduler.shutdownNow();
+      Thread.currentThread().interrupt();
+    }
```
laurit, line 86: Imo calling `shutdownNow` does not make sense. If you really care whether it manages to export or not I think you should be using a more generous timeout than 1s.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/StackTrace.java
laurit: I don't really get how having `equals` and `hashCode` here helps. As far as I can tell every `StackTrace` is unique and identity comparison would work fine.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java
tduncan: Yes. I'll model this in a test.

We have 2 semi-competing interests here: 
1) Export `StackTrace`s in a timely manner
2) Prevent too many `StackTrace`s from building up

I think its OK if the capacity limit triggers and, shortly after, a scheduled export also triggers. The 2nd `export` will be much smaller and not as space efficient but I don't think its super important to micro-manage those priorities to maximize the efficiency of the exported `pprof` package further downstream.

I do think having two workers here makes sense to simplify only allowing a single export invocation at a time.

tduncan: @laurit, I did take some cues from `BatchSpanProcessor` in the latest revision by both using a single background worker thread and coordinating the thread using a wake up signal.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.DelayQueue;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
+
+class PeriodicallyExportingStagingArea implements StagingArea {
```
laurit, line 32: this is more complicated than what I would have expected. I tried to simplify it in https://github.com/tduncan/splunk-otel-java/pull/2 WDYT?

tduncan: I like not having to maintain a separate field for the number of stack traces. I also liked the shutdown approach better. I don't like having to directly manage as much detail around when wake up (that's why I opted for the `DelayedQueue`).

I think that's the biggest difference; I pushed scheduling complexity into the queue where as [here](https://github.com/tduncan/splunk-otel-java/pull/2) its directly in the worker method. Most of the LoC difference is not needing a companion `Delayed` implementation class.

tduncan: @laurit, I realized I had essentially recreated a very simple `ScheduledExecutorService` without actually solving the many of the race conditions.

I've merged your suggestions and made a few mostly superficial changes.

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -81,6 +81,15 @@ public class Configuration implements AutoConfigurationCustomizerProvider {
       "splunk.snapshot.profiler.sampling.interval";
   private static final Duration DEFAULT_SNAPSHOT_PROFILER_SAMPLING_INTERVAL = Duration.ofMillis(10);
 
+  private static final String CONFIG_KEY_SNAPSHOT_PROFILER_STAGING_EMPTY_INTERVAL =
+      "splunk.snapshot.profiler.staging.empty.interval";
+  private static final Duration DEFAULT_SNAPSHOT_PROFILER_STAGING_EMPTY_INTERVAL =
+      Duration.ofSeconds(5);
```
laurit, line 87: Perhaps we should go with a larger interval, like 30s, to allow for more stacks to accumulate. Are there any downsides for this?

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -81,6 +81,15 @@ public class Configuration implements AutoConfigurationCustomizerProvider {
...
+
+  private static final String CONFIG_KEY_SNAPSHOT_PROFILER_STAGING_CAPACITY =
+      "splunk.snapshot.profiler.staging.capacity";
```
laurit, line 90: sdk uses `otel.bsp.max.queue.size` for similar purpose

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    void add(StackTrace stackTrace) {
+      // If queue is full drop the stack trace, not much we can do.
+      queue.offer(stackTrace);
+    }
+
+    @Override
+    public void run() {
+      List<StackTrace> stackTracesToExport = new ArrayList<>();
```
laurit, line 99: We could also consider whether we need the `AsyncStackTraceExporter` at all. Now we could use a single `CpuEventExporter` and just write the stacks there and `flush` to trigger export.

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
laurit: I think 2000 is reasonable for start.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java
laurit: I think that limit doesn't really apply here. Sdk waits 10s for the `CompletableResultCode` to produce a result, but it does not apply a limit on how long producing that `CompletableResultCode` could take. Unfortunately the shutdown handling is still incomplete as we don't wait for the actual export to complete.

laurit (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
tduncan: Longer delay means more likely that a trace is fully ingested and available in the UI but the trace profiling callgraph is incomplete/not yet available which is poor UX. Is 30 seconds often enough for profiling data to win the race against trace ingestion?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java
tduncan: Yes. Lets explore that option once this PR is merged.

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
tduncan: Poor UX because we indicate next to a trace ID that a call graph is available when in fact the call graph hasn't been (fully) ingested yet. This edge case is unavoidable in some circumstances but we want to avoid it as much as possible.

This is an advantage of the "export on trace end" approach. Assuming there isn't a breakdown during the profiling ingestion process the call graph is nearly always persisted before a trace.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicallyExportingStagingArea.java
tduncan: Updated the shutdown process to wait for thread exit. Now the question is what happens if the thread itself has stalled during shutdown?

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
laurit: > This is an advantage of the "export on trace end" approach. Assuming there isn't a breakdown during the profiling ingestion process the call graph is nearly always persisted before a trace.

This would only be true if the profiling ingestion operates faster than the trace ingestion. There could also be a collector between the app and ingest that further batches spans. I don't think we need to worry about this too much, there is inevitably a delay before the data appears in the apm. We can leave this to 5s for now.

PR #2302, opened 2025-05-08
Log Snapshot Profiling Properties On Agent Startup

tduncan:
This PR prints snapshot profiling configuration properties to the agent logs when the agent starts when snapshot profiling is enabled.


profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
tduncan, line 1: All deleted code has been moved to `SnapshotProfilingConfiguration`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingConfiguration.java
tduncan, line 1: Similar to the existing `Configuration` class with the added inclusion of the `log` method found in `ConfigurationLogger`.

tduncan:
recheck

laurit (commented):
If it were up to me I'd move the logging of profiler arguments to debug level in the next major release. What makes profiling and snapshot profiling special so that we should display the configuration for them but we don't display it for any of the other functionality? If the intent is to give some sort of indication that the profiler was enabled then we could do this without logging all the obscure options. @breedx-splk probably disagrees with me

lo-jason:
@laurit approach that we have debug for options and info for snapshot enabled looks good to me. I would also like to see we have info log that shows the log endpoint?

It would be easier to review if re-factor of is in a separate PR

breedx-splk:
> If it were up to me I'd move the logging of profiler arguments to debug level in the next major release. What makes profiling and snapshot profiling special so that we should display the configuration for them but we don't display it for any of the other functionality? If the intent is to give some sort of indication that the profiler was enabled then we could do this without logging all the obscure options. @breedx-splk probably disagrees with me

I've backed down on my incessant desire for config logging at startup. :) Logging it at debug and allowing `-Dotel.javaagent.debug=true` to govern seems like the reasonable way.

tduncan:
Updated the snapshot profiling property logging to `debug` level.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSdkCustomizer.java
```diff
@@ -46,16 +45,18 @@ public SnapshotProfilingSdkCustomizer() {
...
       return new ScheduledExecutorStackTraceSampler(
           StagingArea.SUPPLIER, SpanTracker.SUPPLIER, samplingPeriod);
     };
   }
 
   private static StagingArea createStagingArea(ConfigProperties properties) {
-    Duration interval = Configuration.getSnapshotProfilerExportInterval(properties);
-    int capacity = Configuration.getSnapshotProfilerStagingCapacity(properties);
+    Duration interval =
+        SnapshotProfilingConfiguration.getSnapshotProfilerExportInterval(properties);
```
breedx-splk, line 58: nit: these long names are good candidates for static imports.

tduncan: @breedx-splk, I've gone ahead and removed the `SnapshotProfiler` part of most of the function names. Seemed redundant given it's scoped to the `SnapshotProfilingConfiguration` class.

laurit (approved):

PR #2304, opened 2025-05-13
Nocode class and method selectors

johnbley:
Add logic for more complex class/method selectors than just exact-name-matching: 
  - and
  - or
  - not
  - hasSuperType
  - named
  - nameMatches (regex)
  - hasParameterCount
  - hasParameterOfType (index, type)

 I tried to keep the list/complexity as small as possible while still accounting for 95%+ of real-world usage.  I had to add a separate `-testing` module to work around a classloader issue in order to properly test all the invalid parse logic. 

A larger classloader problem I encountered was that the yaml is clearly specifying bytebuddy `ElementMatcher`s (and the logic for producing and using them is pretty clean since the feature set offered by bytebuddy is so complete), but bytebuddy classes are not visible to the bootstrap-located `Rule` class.  ~~Rather than wrap the `ElementMatcher` logic in a bootstrap-only shim class, for now I'm exposing them as simple `Object`s (with some comments) since they are never used or invoked at the bootstrap layer.   It's not a great design, and I'm happy to code up anything better that anybody else comes up with.~~   Lauri suggested a fix separating a bootstrap interface for `Rule` from a `RuleImpl` in agent space which can safely reference the `ElementMatcher` directly.  Thanks, Lauri!

### DIFF
--- a/instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
+++ b/instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
@@ -136,6 +136,7 @@ class YamlParser {
     return answer;
   }
 
+  @SuppressWarnings("unchecked")
   private static <E extends NamedElement> List<ElementMatcher<E>> matcherListFromYaml(
       Object yamlObject, Map<String, MatcherParser> parsers) {
     if (!(yamlObject instanceof List)) {
@@ -154,6 +155,7 @@ class YamlParser {
     if (!(yamlObject instanceof Map)) {
       throw new IllegalArgumentException("Bare yaml value not expected: " + yamlObject);
     }
+    @SuppressWarnings("unchecked")
     Map<String, Object> yaml = (Map<String, Object>) yamlObject;
     if (yaml.size() != 1) {
       throw new IllegalArgumentException("Multiple yaml elements not allowed without and:/or:");
@@ -163,26 +165,27 @@ class YamlParser {
     return matcherFromKeyAndYamlValue(key, value, parsers);
   }
 
+  @SuppressWarnings("unchecked")
   private static <E extends NamedElement> ElementMatcher<E> matcherFromKeyAndYamlValue(
       String key, Object value, Map<String, MatcherParser> parsers) {
     MatcherParser parser = parsers.get(key);
     if (parser == null) {
       throw new IllegalArgumentException("Unknown yaml element: " + key);
     }
-    return parser.parse(value, parsers);
+    return (ElementMatcher<E>) parser.parse(value, parsers);
   }
 
   enum MatcherParser {
     not {
-      <E extends NamedElement> ElementMatcher<E> parse(
+      ElementMatcher<? extends NamedElement> parse(
           Object value, Map<String, MatcherParser> parsers) {
         return ElementMatchers.not(matcherFromYaml(value, parsers));
       }
     },
     and {
-      <E extends NamedElement> ElementMatcher<E> parse(
+      ElementMatcher<? extends NamedElement> parse(
           Object value, Map<String, MatcherParser> parsers) {
-        ElementMatcher.Junction<E> matcher = ElementMatchers.any();
+        ElementMatcher.Junction<NamedElement> matcher = ElementMatchers.any();
         for (ElementMatcher<NamedElement> sub : matcherListFromYaml(value, parsers)) {
           matcher = matcher.and(sub);
         }
@@ -190,9 +193,9 @@ class YamlParser {
       }
     },
     or {
-      <E extends NamedElement> ElementMatcher<E> parse(
+      ElementMatcher<? extends NamedElement> parse(
           Object value, Map<String, MatcherParser> parsers) {
-        ElementMatcher.Junction<E> matcher = ElementMatchers.none();
+        ElementMatcher.Junction<NamedElement> matcher = ElementMatchers.none();
         for (ElementMatcher<NamedElement> sub : matcherListFromYaml(value, parsers)) {
           matcher = matcher.or(sub);
         }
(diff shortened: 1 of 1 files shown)


bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
```diff
@@ -29,23 +29,23 @@ public static final class Rule {
     private static final AtomicInteger counter = new AtomicInteger();
 
     private final int id = counter.incrementAndGet();
-    private final String className;
-    private final String methodName;
+    private final Object classMatcher; // ElementMatcher
```
laurit, line 32: The reason we need this class in boot loader is for it to be usable from the advice. `NocodeInstrumentation` that uses the matchers has access to agent code. Perhaps it would help if we somehow took advantage of it, we could split the rule into stuff that is needed from the advice and from the agent code.

instrumentation/nocode/README.md
```diff
@@ -40,3 +40,28 @@ the following variables:
...
+        - named: some.specific.Name
+        - nameMatches: some.*Regex.*
+      - hasSuperType: some.superclass.or.Interface
+  method:
+    and:
+      - nameMatches: methodName.*Regex.*
+      - not:
+          nameMatches: specificMethodNameToNotMatch
+      - hasParameterCount: 2
+      - hasParameterOfType: 0 int
```
laurit, line 61: this doesn't look like a typical yaml solution

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
```diff
@@ -19,93 +19,233 @@
...
+      try (Reader reader = Files.newBufferedReader(Paths.get(yamlFileName.trim()))) {
+        return new YamlParser(reader).getInstrumentationRules();
+      }
+    } catch (IOException e) {
+      logger.log(Level.SEVERE, "Can't load configured nocode yaml.", e);
+      return Collections.emptyList();
+    }
+  }
+
+  public static List<NocodeRules.Rule> parseFromString(String yaml) {
```
laurit, line 63: this method is only for testing?

instrumentation/nocode/README.md
johnbley: What would you recommend?
```yaml
  - hasParameter: 0
     ofType: int
```

```yaml
  - hasParameter:
        index: 0
        ofType: int
```

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
johnbley: Yes, is there a convention to follow for that?

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
```diff
@@ -19,93 +19,233 @@
...
+      }
+    },
+    nameMatches {
+      ElementMatcher parse(Object value, Map<String, MatcherParser> parsers) {
+        return ElementMatchers.nameMatches(value.toString());
+      }
+    },
+    hasSuperType {
+      ElementMatcher parse(Object value, Map<String, MatcherParser> parsers) {
+        return ElementMatchers.hasSuperType(ElementMatchers.named(value.toString()));
```
laurit, line 226: we usually use `hasSuperType` from `io.opentelemetry.javaagent.extension.matcher.AgentElementMatchers`. I think it behaves better when super type of the tested type can't be found

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
johnbley: Not sure I follow the idea here?  Are you suggesting something like a`bootstrap.NocodeRules` and `instrumentation.NocodeRules` with the same index used for both?

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
```diff
@@ -19,93 +19,233 @@
...
+                MatcherParser.named,
+                MatcherParser.nameMatches));
+    parsers.addAll(Arrays.asList(extras));
+    for (MatcherParser mp : parsers) {
+      answer.put(mp.name(), mp);
+    }
+    return answer;
+  }
+
+  private static List<ElementMatcher> matcherListFromYaml(
```
laurit, line 154: you could add a type parameter to `ElementMatcher` to reduce the amount of warnings

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
laurit: could make package private (if it is in the same package as tests) and add a comment that it is only used from tests

bootstrap/src/main/java/com/splunk/opentelemetry/javaagent/bootstrap/nocode/NocodeRules.java
laurit: What if you make `NocodeRules.Rule` an interface that includes only the methods needed by the advice. In the non-advice code you could use `RuleImpl` that implements `Rule` and provides more methods.

instrumentation/nocode/README.md
laurit: I think I'd go with the second one. @breedx-splk @robsunday could you review the yaml syntax. We can't really change it later so we should strive to make it as good as possible from the start. One change we could make would be to use nouns as the keys `ofType` -> `type`, `hasParameterCount` -> `parameterCount`, `named` -> `name`, `nameMatches` -> ? wdyt?

breedx-splk: I dunno, yaml isn't the most expressive syntax to begin with. I like what @laurit suggested, to nounify the keys, something like:

```yaml
not:
  name: specificMethodNameToNotMatch
parameterCount: 2
parameter:
  index: 0
  type: int
parameter:
  name: swizzle
```

johnbley: OK, I reworked to 

```
- name: specificName
- nameRegex: whatever.*
- parameter:
    index: 0
    type: int
- superType: someclass.or.Interface
- parameterCount: 99
```

More bikeshedding welcome as changing names is easy at this point (and saves time on a future `-contrib` PR)!

robsunday: +1,

Alternatively:
```yaml
not:
  name: specificMethodNameToNotMatch
parameters:
  - parameter:
    type: int
  - parameter:
    name: swizzle
  - parameter:
    name: userNames
    type: java.util.List
```

robsunday:
And one more question. YAML files for declarative config use snake case naming convention, rather than camelCase. Maybe we should use the same?

johnbley:
> And one more question. YAML files for declarative config use snake case naming convention, rather than camelCase. Maybe we should use the same?

Good call.  I updated all the old stuff to `snake_case` too (e.g., `span_status` instead of `spanStatus`, and touched up the docs as well.

instrumentation/nocode-testing/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParserTest.java
```diff
@@ -0,0 +1,97 @@
...
+              + "  method:\n"
+              + "    parameter: bareValueNotSupported\n",
+          "- class: someClassName\n"
+              + "  method:\n"
+              + "    parameter:\n"
+              + "      noIndex: 1\n"
+              + "      noType: int\n",
+      })
+  // spotless:on
+  void invalidYamlIsInvalid(String yaml) {
```
robsunday, line 94: This test should have more cases, including the valid yaml

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeInitializer.java
```diff
@@ -23,14 +23,27 @@
...
+  private static final String NOCODE_YMLFILE = "splunk.otel.instrumentation.nocode.yml.file";
+  private static final Logger logger = Logger.getLogger(NocodeInitializer.class.getName());
 
   @Override
   public void beforeAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk) {
     ConfigProperties config = getConfig(autoConfiguredOpenTelemetrySdk);
-    YamlParser yp = new YamlParser(config);
-    NocodeRules.setGlobalRules(yp.getInstrumentationRules());
+    String yamlFileName = config.getString(NOCODE_YMLFILE);
+    List<NocodeRules.Rule> instrumentationRules = Collections.EMPTY_LIST;
```
laurit, line 40: you could avoid the unchecked assignment warning with
```suggestion
    List<NocodeRules.Rule> instrumentationRules = Collections.emptyList();
```

instrumentation/nocode-testing/src/test/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParserTest.java
```diff
@@ -0,0 +1,103 @@
...
+class YamlParserTest {
+  @Test
+  void testBasicRuleParsesOK() throws Exception {
+    String yaml = "- class: someClass\n" + "  method: someMethod\n";
+    assertEquals(1, YamlParser.parseFromString(yaml).size());
+  }
+
+  // formatting of the strings with newlines makes it easier to read and
+  // reason about the test.
+  // spotless:off
```
laurit, line 35: an alternative that could be considered would be to use text blocks since the tests run with java 17

johnbley: I'll see what's available when I reformat this whole thing for `-contrib`.

PR #2305, opened 2025-05-14
Add Snapshot Selection Strategy Based On TraceID

tduncan:
This PR adds a second strategy for selecting a trace for snapshotting based on that trace's ID.

The intent of this new selection strategy is to increase the number of traces that are selected for snapshotting by allowing participating services downstream of the root service to locally and deterministically make a snapshotting decision if the trace has not already considered.

This is helpful when the trace root is not capable of making the decision -- likely because the installed agent does not yet have the ability to snapshot a trace -- but downstream services are. Specific use cases are a RUM initiated trace, traces starting from a language without a snapshotting capable agent (e.g. golang, .NET, javascript), or traces started from a Java service with an older splunk-otel-java agent installation.


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceIdBasedSnapshotSelector.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Override
+  public boolean select(Context context) {
+    SpanContext spanContext = Span.fromContext(context).getSpanContext();
+    if (!spanContext.isValid()) {
+      return false;
+    }
+
+    int hash = Math.abs(spanContext.getTraceId().hashCode());
```
rnterbush, line 40: Do we want to align the strategy here with some of the algorithms used for sampling in the collector? 

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/probabilisticsamplerprocessor/README.md

tduncan: Possibly.

I think the main criteria, though, is to use an algorithm that can be easily replicated across many languages. Admittedly Java's `String` hashing algorithm may not be a great choice.

rnterbush: Agreed, maybe at least using a more specific hashing algorithm for which there are libraries in the other languages we're interested in would help with consistency. I was just calling it out now because it's a difficult thing to change later. If we decide to change the approach, the customer could end up with an environment where some agents have been upgraded to use the new algorithm, while others are still using the old algorithm and making different sampling decisions

tduncan: @rnterbush, do you have a suggestion for which algorithm to use?

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotSelectorTest.java
```diff
@@ -0,0 +1,45 @@
...
+    assertThat(selector.select(context)).isEqualTo(expected);
+  }
+
+  private static Stream<Arguments> oneOrTheOther() {
+    return Stream.of(
+        Arguments.of(TRUE.or(TRUE), true),
+        Arguments.of(TRUE.or(FALSE), true),
+        Arguments.of(FALSE.or(TRUE), true),
+        Arguments.of(FALSE.or(FALSE), false));
+  }
```
breedx-splk, line 44: can we add another case that confirms that the other isn't evaluated when this one is true?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ProbabilisticSnapshotSelector.java
```diff
@@ -26,8 +28,16 @@ public ProbabilisticSnapshotSelector(double selectionRate) {
   }
 
   @Override
-  public boolean select() {
+  public boolean select(Context context) {
+    if (!isTraceRoot(context)) {
+      return false;
+    }
```
breedx-splk, line 34: Heh.
```suggestion
    return isTraceRoot(context);
```

tduncan: This would not execute the random selection below.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotSelectorTest.java
tduncan: Yes, I can add a test verifying the short circuit behavior.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceIdBasedSnapshotSelector.java
laurit: Did you consider using `TraceIdRatioBasedSampler`? Algorithm is described in https://github.com/open-telemetry/opentelemetry-specification/blob/9eee5293f95b9fd74f6f1c280b97f87aaec872d7/specification/trace/sdk.md#traceidratiobased-sampler-algorithm

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotVolumePropagator.java
```diff
@@ -40,22 +39,24 @@ public Collection<String> fields() {
...
-   * placing an entry in OpenTelemetry's {@link io.opentelemetry.api.baggage.Baggage}. <br>
-   * <br>
-   * If we're somewhere downstream of the trace root we leave the {@link Context} unchanged.
+   * Make a decision whether a trace should be selected for snapshotting if --and only if -- that
+   * decision has not yet been made by an upstream service. Ideally a snapshotting decision will
+   * have been made only at trace root, however there are many scenarios (e.g. RUM, or language
+   * agents without snapshot support) where a trace will begin without a decision. In those
+   * instances we would like to snapshot as much of the trace as possible.
+   *
+   * <p>Not seen here in the introduction of {@link TraceIdBasedSnapshotSelector} which will
```
laurit, line 48: Idk whether this is relevant but one side effect of using `TraceIdBasedSnapshotSelector` is that it might produce a surprising result with sampling. Say that you are using the same thread id strategy for sampling traces. If you select 10% traces and use the same strategy to select 10% for snapshotting then you will still end up with 100% of traces that were sampled also be selected for snapshotting.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,309 @@
...
+
+package com.splunk.opentelemetry.profiler.snapshot;
+
+import io.opentelemetry.api.internal.OtelEncodingUtils;
+import io.opentelemetry.sdk.trace.IdGenerator;
+
+import java.util.ArrayDeque;
+import java.util.Map;
+
+class SpecialTraceIds {
```
laurit, line 25: is this class still relevant?

tduncan: Yes. I changed the percentile calculation code to match the new approach and replaced all of the pre-computed trace IDs. The trace IDs themselves are used in a of the newly added tests.

laurit: That is a shame, I hoped I wouldn't need to review this. I have a hard time believing that what it does makes sense. How do negative percentiles make sense? Since the algorithm here does not align with `TraceIdRatioBasedSampler` what do these numbers mean at all? Is it a coincidence that the tests pass with these numbers? I guess I'll try to figure this out on Monday.

laurit: After spending some time I think I figured out how this works. Since I thought that `TraceIdRatioBasedSampler` works a bit differently from how it really works I had to first look at `TraceIdRatioBasedSampler` for this to start making sense. It would be nice if you could add comments that point out that these trace ids are chosen based on how `TraceIdRatioBasedSampler` selects traces, this would make it a lot easier for those who might look at this class later. Also you could elaborate that the negative percentiles are also based on how `TraceIdRatioBasedSampler` splits the number range.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,309 @@
...
+          Map.entry(-9, "1022610fdec29ffef4f85a3ce56bbdc7"),
+          Map.entry(-8, "c44f365084aab186f68841e3e1208b38"),
+          Map.entry(-7, "ea6cbd5179d37f3ff7d0afe2a209fa68"),
+          Map.entry(-6, "48037208dec71569f8cb0b67c5e01381"),
+          Map.entry(-5, "986b203a02447d5df9bb619e3666421d"),
+          Map.entry(-4, "947035a119bd5455fb861c001125dc25"),
+          Map.entry(-3, "4cd88ee12d42abe5fc417cca5a01baac"),
+          Map.entry(-2, "436cdf1964184276fdbce3f148296994"),
+          Map.entry(-1, "d3d251f275071212fec6ac6aa3e880fa"),
+          Map.entry(0, "a28341e910992ea0fedb40ef4a39b11a"),
```
laurit, line 147: This is a bit weird, I'd expect the value at 0 percentile to be never sampled. The second long in the trace id should be `Long.MIN_VALUE`. Here the 0 range seems to be the same as the -1 range.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,309 @@
...
+class SpecialTraceIds {
+  /**
+   * Get trace ID known to have a computed percentile equal to the requested value. Negative values
+   * are accepted.
+   *
+   * @return trace ID with matching computed percentile
+   * @throws IllegalArgumentException when provided percentile's absolute value of is less than 0 or
+   *     more than 100
+   */
+  static String forPercentile(int percentile) {
```
laurit, line 34: did you consider going with something like
```java
    if (!isValidPercentile(percentile)) {
      throw new IllegalArgumentException("Invalid percentile: " + percentile);
    }
    if (percentile < 0) {
      // imo negative percentiles should also be invalid
      percentile = -percentile;
    }
    long idHi = 1;
    long idLo;
    if (percentile == 0) {
      idLo = Long.MIN_VALUE;
    } else if (percentile == 100) {
      idLo = Long.MAX_VALUE;
    } else {
      long idUpperBound = (long) (percentile / 100.0 * Long.MAX_VALUE);
      idLo = new Random().nextLong(-idUpperBound + 1, idUpperBound);
    }
    return TraceId.fromLongs(idHi, idLo);
```

tduncan: I'd like to avoid dynamically generating trace IDs on demand if possible. If a test unexpectedly begins to fail having a static list of trace IDs should help in recreating and diagnosing the failure.

laurit: if you don't like randomness then you could pick the trace id based on the upper or lower bound or a midway point or use a fixed seed for the `Random`.

lo-jason:
looks good to me

lo-jason (approved):
looks good, just comment on testing

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
tduncan: @lo-jason, that we use `TraceIdRatioBasedSampler` internally is an implementation detail. We could choose to recreate the hashing algorithm and probability range computation ourselves.

Its very important to have tests fail if the the upstream OpenTelemetry behavior were to ever change; it's a clear signal that the previous behavior needs to be recreated locally in order to maintain compatibility between agents. So the tests are verifying our assumptions about how our dependency behaves and if those assumptions ever change these tests should protect against shipping a broken version of the agent.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,256 @@
...
+ * <pre>
+ * Ex. For percentile 5 the range is considered to be:
+ * lower = {@link Long#MAX_VALUE} * 0.04 = 368934881474191040
+ * upper = {@link Long#MAX_VALUE} * 0.05 = 461168601842738816
+ * </pre>
+ *
+ * <p>A trace ID is considered to be in the percentile range when the absolute value of its hash
+ * value (as calculated using {@link
+ * io.opentelemetry.api.internal.OtelEncodingUtils#longFromBase16String(traceId, 16)}) is within the
+ * range.
```
laurit, line 45: It isn't really a hash. Trace id is 128 bits, which is 2 longs. A long is 16 hex chars. `longFromBase16String` transforms the hex string to long. `16` is the offset to the start of the second long.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
laurit: The chances of the behavior of this changes are very small, after all it is described in the otel spec. Otel takes backwards compatibility quite seriously.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * @see io.opentelemetry.sdk.trace.samplers.TraceIdRatioBasedSampler
+ * @see io.opentelemetry.api.internal.OtelEncodingUtils#longFromBase16String(CharSequence, int)
+ */
+class SpecialTraceIds {
```
laurit, line 50: Imo `SpecialTraceIds` isn't a suitable name for this class, these trace ids aren't that special.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,256 @@
...
+  /**
+   * Get trace IDs known to be within the requested percentile. At least two traces will be returned
+   * for each percentile representing both positive and negative hash value calculations.
```
laurit, line 53: > representing both positive and negative hash value calculations

perhaps rephrase to `representing second long in the trace id being both positive and negative`

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+    return ranges;
+  }
+
+  private static Map<Integer, List<String>> generateTraceIds(Queue<PercentileRange> ranges) {
+    var map = new LinkedHashMap<Integer, List<String>>();
+    while (!ranges.isEmpty()) {
+      var range = ranges.remove();
+      boolean foundTraceIdInRange = false;
+      while (!foundTraceIdInRange) {
```
laurit, line 196: as far as I can tell this loop always does only 1 iteration

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,256 @@
...
+        var lower = new Random().nextLong(range.lower + 1, range.upper);
+        var traceId = TraceId.fromLongs(range.upper, lower);
```
laurit, line 198: might help to have a comment explaining that using `range.upper` here doesn't have a deeper meaning, `TraceIdRatioBasedSampler` doesn't use the first long.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,256 @@
...
+    private final long upper;
+
+    private PercentileRange(int value, boolean negative) {
+      this.value = value;
+      if (!negative) {
+        this.lower = (long) (Long.MAX_VALUE * ((double) (value - 1) / 100));
+        this.upper = (long) (Long.MAX_VALUE * ((double) (value) / 100));
+      } else {
+        this.upper = (long) (Long.MIN_VALUE * ((double) (value - 1) / 100));
+        this.lower = (long) (Long.MIN_VALUE * ((double) (value) / 100));
```
laurit, line 245: since `Long.MAX_VALUE != -Long.MIN_VALUE` and the `TraceIdRatioBasedSampler` only uses `Long.MAX_VALUE` it is difficult to tell whether this recreates the same value range as the code in `TraceIdRatioBasedSampler`

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,256 @@
...
+    for (int percentile = 1; percentile <= 99; percentile++) {
+      ranges.add(new PercentileRange(percentile, false));
+      ranges.add(new PercentileRange(percentile, true));
+    }
+    return ranges;
+  }
+
+  private static Map<Integer, List<String>> generateTraceIds(Queue<PercentileRange> ranges) {
+    var map = new LinkedHashMap<Integer, List<String>>();
+    while (!ranges.isEmpty()) {
```
laurit, line 193: Putting the ranges into a queue and iterating over them feels like a roundabout way. You could get it done with less code.
```java
  public static void main(String[] args) {
    var map = new LinkedHashMap<Integer, List<String>>();
    for (int percentile = 1; percentile <= 99; percentile++) {
      long lower = (long) (Long.MAX_VALUE * ((double) (percentile - 1) / 100));
      long upper = (long) (Long.MAX_VALUE * ((double) (percentile) / 100));

      var traceIds = new ArrayList<String>();
      var random = new Random().nextLong(lower + 1, upper);
      traceIds.add(TraceId.fromLongs(0, random));
      traceIds.add(TraceId.fromLongs(0, -random));
      map.put(percentile, traceIds);
    }
    System.out.println(asJavaCode(map));
  }
```
hope I got the logic correctly

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
tduncan: Yes. Artifact left over from generating random trace IDs and checking if they fell within the range. Removing.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
tduncan: It does work, yes.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,256 @@
...
+import java.util.Map;
+import java.util.Queue;
+import java.util.Random;
+
+/**
+ * Maintains a collection of trace IDs mapped to percentiles matching the probability calculation in
+ * {@link io.opentelemetry.sdk.trace.samplers.TraceIdRatioBasedSampler}.
+ *
+ * <p>A trace ID percentile is defined as the range of values existing between {@link
+ * Long#MAX_VALUE} * (percentile / 100) and {@link Long#MAX_VALUE} * (percentile + 1 / 100)
```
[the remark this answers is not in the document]
tduncan: Tried to describe how negative values are considered.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
tduncan: Removed the word hash to hopefully be more precise in the language.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
tduncan: @laurit, no doubt. Though our need to maintain backwards compatibility with the snapshotting decision is not inherently linked with the desire for Otel to also maintain backwards compatibility of the trace-based sampling. Either us or Otel may decide to break compatibility for reasons entirely unrelated to the other.

The primary difference is that Otel doesn't care about our snapshotting decision strategy where as we are (currently) dependent on Otel's trace-based sampling decision.

It probably won't happen but the tests are inexpensive and should catch it (assuming the human process doesn't first).

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
tduncan: Suggestion for a better name?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/Volume.java
```diff
@@ -23,7 +23,8 @@
 
 public enum Volume implements ImplicitContextKeyed {
   OFF,
-  HIGHEST;
+  HIGHEST,
+  UNSPECIFIED;
```
lo-jason, line 27: I don't see the volume is part of the spec https://github.com/signalfx/gdi-specification/blob/main/specification/behaviors.md That should be added as this needs to function the same across all agents. UNSPECIFIED is not part of spec as this is internal behavior to agent (it will never be sent over the wire) would be good to add a comment to that effect.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
lo-jason: could it be called PercentileSamplingTraceIdGenerator?

lo-jason (approved):
looks good in general. For the comment on spec we can file follow-up no action needed in this PR.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SpecialTraceIds.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+
+/**
+ * Maintains a collection of trace ids mapped to percentiles matching the probability calculation in
+ * {@link io.opentelemetry.sdk.trace.samplers.TraceIdRatioBasedSampler}.
+ *
+ * <p>A trace ID percentile is defined as the range of values existing between {@link
+ * Long#MAX_VALUE} * (percentile / 100) and {@link Long#MAX_VALUE} * (percentile + 1 / 100). A trace
```
laurit, line 31: code uses `percentile - 1` to `percentile` as the bounds

laurit (approved):

PR #2320, opened 2025-05-23
Unregister Stalled Traces 

tduncan:
This PR attempts to detect and automatically unregister stalled traces from snapshotting.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
@@ -29,9 +29,9 @@ class ActiveSpanTracker implements ContextStorage, SpanTracker {
   private final Cache<Thread, SpanContext> cache = Cache.weak();
 
   private final ContextStorage delegate;
-  private final ITraceRegistry registry;
+  private final TraceRegistry registry;
 
-  ActiveSpanTracker(ContextStorage delegate, ITraceRegistry registry) {
+  ActiveSpanTracker(ContextStorage delegate, TraceRegistry registry) {
     this.delegate = delegate;
     this.registry = registry;
   }
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
@@ -29,9 +29,9 @@ class ActiveSpanTracker implements ContextStorage, SpanTracker {
   private final Cache<Thread, SpanContext> cache = Cache.weak();
 
   private final ContextStorage delegate;
-  private final TraceRegistry registry;
+  private final ITraceRegistry registry;
 
-  ActiveSpanTracker(ContextStorage delegate, TraceRegistry registry) {
+  ActiveSpanTracker(ContextStorage delegate, ITraceRegistry registry) {
     this.delegate = delegate;
     this.registry = registry;
   }
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
@@ -23,15 +23,16 @@ import io.opentelemetry.context.ContextStorage;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.instrumentation.api.internal.cache.Cache;
 import java.util.Optional;
+import java.util.function.Supplier;
 import javax.annotation.Nullable;
 
 class ActiveSpanTracker implements ContextStorage, SpanTracker {
   private final Cache<Thread, SpanContext> cache = Cache.weak();
 
   private final ContextStorage delegate;
-  private final TraceRegistry registry;
+  private final Supplier<TraceRegistry> registry;
 
-  ActiveSpanTracker(ContextStorage delegate, TraceRegistry registry) {
+  ActiveSpanTracker(ContextStorage delegate, Supplier<TraceRegistry> registry) {
     this.delegate = delegate;
     this.registry = registry;
   }
@@ -62,7 +63,7 @@ class ActiveSpanTracker implements ContextStorage, SpanTracker {
   }
 
   private boolean doNotTrack(SpanContext spanContext) {
-    return !spanContext.isSampled() || !registry.isRegistered(spanContext);
+    return !spanContext.isSampled() || !registry.get().isRegistered(spanContext);
   }
 
   @Nullable
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/StalledTraceDetectingTraceRegistry.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/StalledTraceDetectingTraceRegistry.java
@@ -26,7 +26,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
-class StalledTraceDetectingTraceRegistry implements TraceRegistry, AutoCloseable {
+class StalledTraceDetectingTraceRegistry implements TraceRegistry {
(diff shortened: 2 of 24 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/StalledTraceDetectingTraceRegistry.java
```diff
@@ -0,0 +1,160 @@
...
+
+import io.opentelemetry.api.trace.SpanContext;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+class StalledTraceDetectingTraceRegistry implements TraceRegistry {
```
tduncan, line 27: @laurit, originally you suggested using a [`WeakReference`s with a `ReferenceQueue`](([url](https://github.com/signalfx/splunk-otel-java/pull/2181#discussion_r1945181945)) but I wasn't able to figure out how to figure out which traces were removed so the thread sampling could be automatically stopped as well.

tduncan: I think I've answered my 2nd question myself. These references do prevent the `SpanContext` instances from being GC'd. Here's my test:

```java
@Test
void orphanedTrace() throws Exception {
  var spanContext = Snapshotting.spanContext().build();
  registry.register(spanContext);
  sampler.start(spanContext);
  var spanContextCopy = SpanContext.create(
                  spanContext.getTraceId(),
                  spanContext.getSpanId(),
                  spanContext.getTraceFlags(),
                  spanContext.getTraceState());

  var spanContextReference = new WeakReference<>(spanContext);
  spanContext = null;
  GcUtils.awaitGc(spanContextReference, Duration.ofSeconds(10));

  await().untilAsserted(() -> assertFalse(registry.isRegistered(spanContextCopy)));
  await().untilAsserted(() -> assertThat(sampler.sampledTraces).isEmpty());
}

private static class Sampler implements StackTraceSampler {
  private final Set<SpanContext> sampledTraces = new HashSet<>();

  @Override
  public void start(SpanContext spanContext) {
    sampledTraces.add(spanContext);
  }

  @Override
  public void stop(SpanContext spanContext) {
    sampledTraces.remove(spanContext);
  }
}
```

```
reference was not cleared in time
java.util.concurrent.TimeoutException: reference was not cleared in time
	at io.opentelemetry.instrumentation.test.utils.GcUtils.awaitGc(GcUtils.java:44)
	at com.splunk.opentelemetry.profiler.snapshot.StalledTraceDetectingTraceRegistryTest.orphanedTrace(StalledTraceDetectingTraceRegistryTest.java:94)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

laurit: > 1. What is the importance of `thread.setContextClassLoader(null)` in these examples? The documentation says that `null` indicates we want to use the system class loader but its not clear to me why that is important in this context?

I think it isn't really important here since the thread would get created during agent start. It is often a good idea to set context loader to null because the new thread inherits the context class loader from the thread that create it. Imagine that you create a thread at the start of servlet request processing. Since the servlet container sets the context loader to webapp loader your thread will now have a reference to the webapp loader. When the webapp is undeployed your thread will continue to keep the webapp loader alive. Probably would be a good idea to add `setContextClassLoader(null)` to `HelpfulExecutors` since `ScheduledExecutorStackTraceSampler` creates new threads. It is possible to also leak access control context the same way, but that isn't that common.

> When you originally suggested this approach we hadn't added most of the trace profiling extension. Now that we have (and have deviated from the reference extension I was porting from) we also have the same potential memory leak in ScheduledStackTraceSampler (this is why the timeout implementation in this PR stops sampling) which maintains references to SpanContexts. Won't those references prevent the SpanContext instances from being GC'd?

Yes they will. I didn't really look too much into this PR. Just wanted to give you an idea how this could be implemented so you could check whether this approach could be useful or not. If you wish to keep a reference to spancontext you could use span instead for the weak ref target, but as you noted whatever you choose it must not be kept alive.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/OrphanedTraceDetectingTraceRegistry.java
tduncan, line 1: @laurit, converted the previous timer-based implementation for a `WeakReference`-based one very similar to your suggestion.

I'm curious as to you thoughts about implementing this as a `TraceRegistry`? Maybe a `SpanProcessor` would be more appropriate? Or maybe have it be its own thing and obtain the `SpanContext` references in some other way?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ActiveSpanTracker.java
```diff
@@ -22,31 +22,34 @@
...
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.instrumentation.api.internal.cache.Cache;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Supplier;
 import javax.annotation.Nullable;
 
 class ActiveSpanTracker implements ContextStorage, SpanTracker {
-  private final Cache<Thread, SpanContext> cache = Cache.weak();
+  private final Cache<Thread, ProfilingSpanContext> cache = Cache.weak();
```
tduncan, line 31: I couldn't find a way to scan the cache values so if a `SpanContext` is GC'd the trace will be unregistered and sampling will be stopped but the `Thread`->`ProfilingSpanContext` mapping will remain until a new trace is selected and registered on that same thread.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/OrphanedTraceDetectingTraceRegistry.java
laurit: The way you have built it now doesn't seem to benefit much from being a `TraceRegistry` and could be merged into the span processor or made a separate class.

tduncan: @laurit, I've explored some other options but this option still felt the cleanest.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/OrphanedTraceDetectingTraceRegistryTest.java
```diff
@@ -0,0 +1,110 @@
...
+            spanContext.getTraceId(),
+            spanContext.getSpanId(),
+            spanContext.getTraceFlags(),
+            spanContext.getTraceState());
+
+    var spanContextReference = new WeakReference<>(spanContext);
+    spanContext = null;
+    GcUtils.awaitGc(spanContextReference, Duration.ofSeconds(10));
+
+    await().untilAsserted(() -> assertThat(sampler.isBeingSampled(spanContextCopy)).isFalse());
```
laurit, line 108: you should also test that this evaluates to true before the gc

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/OrphanedTraceDetectingTraceRegistry.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @Override
+  public boolean isRegistered(SpanContext spanContext) {
+    return delegate.isRegistered(spanContext);
+  }
+
+  @Override
+  public void unregister(SpanContext spanContext) {
+    delegate.unregister(spanContext);
+    traces.remove(new LookupKey(spanContext));
```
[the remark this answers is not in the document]
tduncan: You are correct. I neglected to correctly implement `.equals`. Fixed.

PR #2331, opened 2025-06-06
Update repos for sonatype central

breedx-splk:
See this annoucement: https://central.sonatype.org/news/20250326_ossrh_sunset/

and see this for the docs on this change https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-central

Once a new token is generated and put into GitLab releaser repo, we can merge this to get the new urls into place.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,7 +9,10 @@ nexusPublishing {
   packageGroup.set("com.splunk")
 
   repositories {
+    // see https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#configuration
     sonatype {
+      nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
+      snapshotRepositoryUrl.set(uri("https://central.sonatype.com/repository/maven-snapshots/"))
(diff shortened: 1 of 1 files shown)


laurit (approved):

build.gradle.kts
```diff
@@ -9,7 +9,10 @@ nexusPublishing {
   packageGroup.set("com.splunk")
 
   repositories {
+    // see https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#configuration
     sonatype {
+      nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
+      snapshotRepositoryUrl.set(uri("https://central.sonatype.com/repository/maven-snapshots/"))
```
laurit, line 15: we may need to update https://github.com/signalfx/splunk-otel-java?tab=readme-ov-file#snapshot-builds if snapshots are now in new location

breedx-splk: agreed. I guess we'll know for sure after the first one is published...because even tho I migrated us, the `com/splunk` snapshots don't yet exist. Maybe/probably just new stuff.

PR #2336, opened 2025-06-10
update windows backend for smoke tests

breedx-splk:
Related to #2335. Not super confident that this will fix the nightly build failure, but it looked from the build scans that they were all failing to stand up the windows backend.

image

Our reference was to a 3 year old image...so maybe they aged out? Let's see what this does anyway...even if it doesn't fix it, maybe it's a reasonable upgrade.


laurit:
closing are reopening to trigger checks

breedx-splk:
Yeah, it was a bit of a hail mary. Should we just do this anyway tho?

laurit:
> Yeah, it was a bit of a hail mary. Should we just do this anyway tho?

we could, but we'd first need to get the tests passing with the new image https://scans.gradle.com/s/da4puo2ie2p5c Could be as easy as `windows-2019` -> `windows-2022` though may also need to update the version for other images, which could bring out other issues.

PR #2341, opened 2025-06-16
Introduce a Single Threaded Stack Trace Sampler

tduncan:
This PR introduces a `StackTraceSampler` that utilizes a single thread for taking samples of all threads associated with trace ids that have been selected for snapshotting. This new `StackTraceSampler` replaces the existing `ScheduledExecutorStackTraceSampler` implementation which used a 1:1 ratio of sampling thread to trace id approach.


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/DaemonThreadStackTraceSampler.java
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  private static class ThreadSampler extends Thread {
+    private final ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
+    private final BlockingQueue<Command> queue = new LinkedBlockingQueue<>();
+    private final Supplier<StagingArea> staging;
+    private final Supplier<SpanTracker> spanTracker;
+    private final Duration delay;
+
+    private long nextSampleTime;
```
laurit, line 76: could be a local variable

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/DaemonThreadStackTraceSampler.java
```diff
@@ -0,0 +1,250 @@
...
+          if (command == null) {
+            sample(traceThreads.values());
+          } else {
+            if (command.action == Action.SHUTDOWN) {
+              return;
+            } else if (command.action == Action.START) {
+              startSampling(command, traceThreads);
+            } else if (command.action == Action.STOP) {
+              stopSampling(command, traceThreads);
+            }
```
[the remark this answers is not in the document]
tduncan: This is how I understand the the hypothetical:

1. Trace 1 starts
2. Thread 1 is requested to be sampled for Trace 1 (`SnapshotProfilingSpanProcessor`)
3. Thread 1 is associated with Trace 1:Span 1 (`ActiveSpanTraker`)
4. Trace 1 ends
5. Thread 1 is unassociated with Trace1:Span 1 ('ActiveSpanTracker`)
6. Thread 1 is requested to no longer be sampled (`SnapshotProfilingSpanProcessor`)
7. Thread 1 added for sampling (`PeriodicStackTraceSampler`)
8. Thread 1 is removed from sampling (`PeriodicStackTraceSampler`)

In the above a sample would be taken at both steps 7 and 8, at which point Thread 1 is doing whatever else (maybe idle, maybe processing a new request). Once the sample is taken a `StackTrace` instance is created using Trace 1's trace id and either an invalid span id or a span id from the next trace.

This is a problem. I'll figure out how to model it in a test.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -0,0 +1,224 @@
...
+    @Override
+    public void run() {
+      long nextSampleTime = System.nanoTime() + delay.toNanos();
+      try {
+        while (!Thread.currentThread().isInterrupted()) {
+          Object object = queue.poll(nextSampleTime - System.nanoTime(), TimeUnit.NANOSECONDS);
+          if (object == SHUTDOWN_SIGNAL) {
+            return;
+          }
+          sample(traceSamplingContexts.values());
```
laurit, line 138: It is possible that by the time the periodic sampling completes `remove` has already been called for some thread, is that ok?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -0,0 +1,224 @@
...
+        logger.log(Level.SEVERE, e, () -> "Unexpected error during callstack sampling");
+      }
+    }
+
+    private List<StackTrace> toStackTraces(
+        ThreadInfo[] threadInfos, Map<Long, SamplingContext> contexts, long currentTimestamp) {
+      List<StackTrace> stackTraces = new ArrayList<>(threadInfos.length);
+      for (ThreadInfo threadInfo : threadInfos) {
+        SamplingContext context = contexts.get(threadInfo.getThreadId());
+        if (traceSamplingContexts.containsKey(context.traceId)) {
```
laurit, line 168: I guess this check is here to remove threads for which `remove` was already called? If it is so then a comment would be nice.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -0,0 +1,224 @@
...
+    private SpanContext retrieveActiveSpan(Thread thread) {
+      return spanTracker.get().getActiveSpan(thread).orElse(SpanContext.getInvalid());
+    }
+  }
+
+  private static class SamplingContext {
+    private final Thread thread;
+    private final String traceId;
+    private final String spanId;
+    private long timestamp;
```
laurit, line 211: should be volatile?

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
tduncan: Its possible, yes. And no, it's not OK. I've attempted to model this in the `doNotStageStackTraceWhenThreadNoLongerAssociatedWithSameTraceId` test.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
tduncan: Yes, this is protecting against reporting stack traces for traces that have been removed after the periodic sample but before the stack trace is staged. I'll add a comment.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -0,0 +1,242 @@
...
+      this.spanId = spanId;
+      this.sampleTime = new AtomicLong(sampleTime);
+    }
+
+    void updateSampleTime(long timestamp) {
+      this.sampleTime.updateAndGet(
+          operand -> {
+            // Prevent sample time being updated to a "past" value due to race condition between
+            // periodic samples
+            // and stop profiling samples
```
laurit, line 234: these lines could be merged

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -0,0 +1,224 @@
...
+          threadInfo, context, context.traceId, spanContext.getSpanId(), currentTimestamp);
+    }
+
+    private StackTrace toStackTrace(
+        ThreadInfo threadInfo,
+        SamplingContext context,
+        String traceId,
+        String spanId,
+        long currentTimestamp) {
+      Duration samplingPeriod = Duration.ofNanos(currentTimestamp - context.timestamp);
```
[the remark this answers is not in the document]
laurit: > What would you suggest as an alternative? Ultimately the source.event.period is used as the stack frame execution time.

In https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md#logrecord-message-text-data-format-specific-attributes the description for `source.event.period` says

> MUST contain the sampling period in milliseconds if this LogRecord represents a periodic event

now here we have a combination of periodic and non periodic activity which imo does not align with the existing description. Both `source.event.period` and `source.event.time` are with millisecond precision, is this enough for your use cases? Perhaps you should consider adding a new attribute. Or you could consider altering the description for the existing attribute.

> I've switched the underlying field to an AtomicLong and added a check to prevent timestamps from the past from overwriting newer ones.

Lets assume that context timestamp is set to `T0`. When `stop` and periodic timer run at the same time they'll capture time as `T1` and `T2` respectively so that `T1 < T2`. Now lets assume that the periodic timer completes first for some reason and updates the context timestamp. Now the sampling period for stop is `T2 - T1` which is a negative value.
I believe you could resolve this with some locking.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
tduncan: Yeah. I think this is from spotless code formatting. I'll try to get them onto a single line.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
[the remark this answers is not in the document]
tduncan: I've been able to reproduce this inconsistently with the existing test and indeed the sampling time is negative. What's interesting to is I've only been able to reproduce it in with the test when using a lock. Without locks (using the same code as in this diff) the test doesn't fail.

I have been able to force the scenario with breakpoints and manually advancing the two competing threads so it is possible.

I'm curious how you think a lock may be used here. I've experimented with a couple of solutions. I'd like to avoid locking the sampling mechanism; it doesn't seem ideal to hold up the periodic sampling thread because a new thread is being added but maybe that's the only way?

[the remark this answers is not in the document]
laurit: > I'm curious how you think a lock may be used here.

If you lock around taking the end sample and taking the periodic sample then they can't run concurrently and see inconsistent state. For locking you could use a `ReentrantReadWriteLock` where end sample uses the shaded lock (read lock) and periodic sample use the exclusive lock (write lock). Or you could synchronize on the sampling context or add a lock into it if you wish to make it more fine grained. My understanding is that you need to ensure is that the periodic sample is finalized before the end sample is added or that the periodic sample is not added at all.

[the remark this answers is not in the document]
laurit: If you wish to avoid locking I guess you could keep state (e.g sample time) in `SamplingContext`. Before you take the sample you read the state and after you use cas to update it. If cas succeeds all is fine. If it does not when you are taking a periodic sample you can assume that end sample was already added and discard the periodic sample. If the cas fails for end sample you can retry taking the end sample (or recompute the sample period) or perhaps when the periodic sample is later than the end sample discard the end sample.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -0,0 +1,234 @@
...
+      return spanTracker.get().getActiveSpan(thread).orElse(SpanContext.getInvalid());
+    }
+  }
+
+  private static class SamplingContext {
+    private final Lock lock = new ReentrantLock();
+    private final Thread thread;
+    private final String traceId;
+    private final String spanId;
+    private volatile long sampleTime;
```
laurit, line 221: as far as I can tell this is now always updated when holding a lock, so doesn't need to be volatile

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -0,0 +1,234 @@
...
+        logger.info("Unexpected error during callstack sampling");
+      }
+    }
+
+    private List<StackTrace> toStackTraces(
+        ThreadInfo[] threadInfos, Map<Long, SamplingContext> contexts, long currentSampleTime) {
+      List<StackTrace> stackTraces = new ArrayList<>(threadInfos.length);
+      for (ThreadInfo threadInfo : threadInfos) {
+        SamplingContext context = contexts.get(threadInfo.getThreadId());
+        // When the context is locked and on demand sample is being taken. No need to report
```
laurit, line 174: should be `an` instead of `and`?
instead of `skip the on-demand sample` should be `skip the periodic sample`?

laurit (approved):

PR #2344, opened 2025-06-17
Run smoke tests on windows 2022

laurit:
Since windows 2019 runners are going to be retired soon we have to migrate to 2022. This PR uses the same versions for testing as the upstream. Note that openj9 isn't available for windows tests any more (docker base image is not available for windows).
This PR also fixes Tomee detection for newer Tomee versions.


breedx-splk (approved):
thanks for getting ahead of this.

PR #2347, opened 2025-06-18
Add build scan links

laurit:


breedx-splk (approved):
So if `build-scan.txt` is missing, apparently `hashFiles()` will return an empty string? I have no idea how you know that, but I'm sure it's fine. 🙈

PR #2348, opened 2025-06-18
Context propagation for sql server

laurit:
Based on https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2709


instrumentation/jdbc/src/main/java/com/splunk/opentelemetry/instrumentation/jdbc/sqlserver/SqlServerUtil.java
```diff
@@ -0,0 +1,58 @@
...
+              if ("traceparent".equals(key)) {
+                carrier.set(value);
+              }
+            });
+    String traceparent = state.get();
+    String existingTraceparent = connectionState.get(connection);
+    if (traceparent == null && existingTraceparent != null) {
+      // we need to clear existing tracing state from the connection
+      connectionState.set(connection, null);
+      PreparedStatement statement = connection.prepareStatement("set context_info 0x");
```
robsunday, line 49: I think it could reuse the same string definition `set context_info ?`, so only one PS would be cached, but this is not a big deal.

robsunday (approved):

instrumentation/jdbc/src/test/java/com/splunk/opentelemetry/instrumentation/jdbc/sqlserver/SqlServerTest.java
```diff
@@ -0,0 +1,366 @@
...
+    createTable.execute("CREATE TABLE test_table (value INT NOT NULL)");
+    createTable.close();
+
+    testing.waitForTraces(1);
+    testing.clearData();
+  }
+
+  @AfterAll
+  static void cleanup() throws Exception {
+    connection.close();
```
breedx-splk, line 80: It looks like the tests create their own connection. Is there a reason to hold it open for the duration? Would it be bad to close the connection after creating the table in the `@BeforeAll`?

instrumentation/jdbc/src/test/java/com/splunk/opentelemetry/instrumentation/jdbc/sqlserver/SqlServerTest.java
```diff
@@ -0,0 +1,366 @@
...
+      callDepthSplunk.decrementAndGet();
+    }
+  }
+
+  private static void assertNoContext(Connection connection) throws SQLException {
+    CallDepth callDepthJdbc = CallDepth.forClass(Statement.class);
+    // disable jdbc instrumentation, so it wouldn't create a span for the statement execution
+    callDepthJdbc.getAndIncrement();
+    try {
+      connection.createStatement().execute("SELECT 1");
```
breedx-splk, line 355: It'll get GC'd I guess, but shouldn't we also close the statement here?

breedx-splk (approved):
Looks good. Had a couple small things, but nothing to hold this up. The use of `CallDepth` took a little bit to get my head around, especially how it's manipulated by the tests...but I _think_ I grok it. I appreciate the thoroughness in the tests!
 
Let's give it a shot! 🚀

instrumentation/jdbc/build.gradle.kts
```diff
@@ -0,0 +1,27 @@
...
+  testInstrumentation("io.opentelemetry.javaagent.instrumentation:opentelemetry-javaagent-jdbc")
+
+  testLibrary("com.microsoft.sqlserver:mssql-jdbc:6.1.0.jre8")
+  testCompileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
+  testImplementation("org.testcontainers:testcontainers")
+  testImplementation("org.testcontainers:mssqlserver")
+}
+
+tasks.withType<Test>().configureEach {
+  jvmArgs("-Dotel.instrumentation.splunk-jdbc.enabled=true")
```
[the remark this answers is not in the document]
breedx-splk: oh yeah I wasn't considering the fact that is fits the generic instrumentation enable/disable naming pattern. Sorry. Let's ignore that for now then. :)

PR #2383, opened 2025-07-18
Profile Trace Across Background Tasks

tduncan:
This PR attempts to have the trace snapshot profiler follow a trace across thread boundaries by hooking into the `Context` attaching process. When a new `Context` is attached the profiler verifies that the `SpanContext` belongs to a trace that has been registered for snapshotting. If yes and sampling will be started for the current `Thread` if it is not already being sampled.

Upon the entry span ending all sampling for the trace id and span id combination will also be stopped.

There is still a gap where a long running background task that outlives its associated `SpanContext` may not be profiled for the duration of that task's execution. Ideally the profiler would continue to sample the background task it's unclear how to reconcile the need to stop sampling for a specific `Thread` associated with a `SpanContext` while also allowing sampling to be stopped from another `Thread` (e.g. a span being started and stopped from different `Threads`). For now this scenario is being deferred to a future change.


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -90,8 +104,7 @@ public void close() {
   }
 
   private static class ThreadSampler extends Thread {
-    private final Map<SpanContext, SamplingContext> traceSamplingContexts =
-        new ConcurrentHashMap<>();
+    private final Map<Thread, SamplingContext> threadSamplingContexts = new ConcurrentHashMap<>();
```
laurit, line 107: Wondering whether we should use a weak map to allow collecting threads even when they weren't removed from the map. I guess that would mean that `stop` wasn't called and there would be other issues as well, so probably not worth the effort.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -233,12 +271,12 @@ private SpanContext retrieveActiveSpan(Thread thread) {
   private static class SamplingContext {
     private final Lock lock = new ReentrantLock();
     private final Thread thread;
-    private final String traceId;
+    private final SpanContext spanContext;
     private long sampleTime;
 
-    private SamplingContext(Thread thread, String traceId, long sampleTime) {
+    private SamplingContext(Thread thread, SpanContext spanContext, long sampleTime) {
```
laurit, line 277: Imo using span context is a bit confusing. If it is a random span context then only the trace id of it is usable, if it is the context of the root span than the span id could also be usable. Here it seems that sometimes it is the root span and sometimes not. You should add a comment explaining that or always make it contain the root span and add a comment or rename the variable.

tduncan: The intent is for the `SpanContext` to be the one associated with the `Thread` when the sampling starts so the sampling is only stopped when both the same `Thread` and `SpanContext` are sent. Its an attempt to prevent sampling from being stopped too early.

laurit: To me the confusing part is not the usage of `SpanContext` per se , but rather that the trace can contain multiple span contexts (every span has its own span context). If you don't take care to use only the root span context here then might as well just use the trace id since the contexts would have different span ids which could lead to bugs.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -110,33 +123,58 @@ private ThreadSampler(
...
+      takeBulkSample(threads);
+    }
+
+    private List<SamplingContext> findAndStopSamplingAssociatedThreads(SpanContext spanContext) {
+      List<SamplingContext> threadsToStopSampling = new ArrayList<>();
+      Iterator<Map.Entry<Thread, SamplingContext>> iterator =
+          threadSamplingContexts.entrySet().iterator();
+      while (iterator.hasNext()) {
+        Map.Entry<Thread, SamplingContext> entry = iterator.next();
+        if (entry.getValue().spanContext.equals(spanContext)) {
```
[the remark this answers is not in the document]
tduncan: The implementation has changed to only use the `SpanContext` when reporting the stack trace. Sampling will be started and stopped when the `Context` is opened and closed respectively for a `Span` which is expected to happen within the same `Thread`. As such only the `Thread` is accepted when stopping sampling.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/OrphanedTraceCleaner.java
```diff
@@ -73,7 +73,7 @@ private void unregisterOrphanedTraces() {
           Key key = (Key) reference;
           if (traces.remove(key)) {
             traceRegistry.unregister(key.getSpanContext().getTraceId());
-            sampler.get().stop(key.getSpanContext());
+            sampler.get().stopAllSampling(key.getSpanContext());
```
tduncan, line 76: @laurit, curious about your thoughts here. When a `Span` is GC'd do you think only the original `Thread` should have sampling stopped? Stop all `Thread`s associated with the `SpanContext` (current behavior)? Stop all profiling for the trace id? Assume the `Context` will be closed appropriately and do nothing?

laurit: The span being GC'd before sampling is stopped indicates an error where span end was not called. We don't expect to have such bugs so it isn't super important. If we are going to handle it then imo it would make sense for it to behave the same way as span end was called.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
tduncan: Without the need to clean up sampling when a Span is unexpectedly GC'd its only necessary to retain the trace id.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceThreadChangeDetector.java
```diff
@@ -0,0 +1,68 @@
...
+    this.delegate = delegate;
+    this.registry = registry;
+    this.sampler = sampler;
+  }
+
+  @Override
+  public Scope attach(Context toAttach) {
+    Scope scope = delegate.attach(toAttach);
+    SpanContext newSpanContext = Span.fromContext(toAttach).getSpanContext();
+    if (doNotTrack(newSpanContext)) {
```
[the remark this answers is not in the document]
tduncan: Yes. I think the most recent commit to this extension added support for profiling multiple trace service entry threads within a single service instance. Changing from mapping `SpanContext` to `Thread` in `PeriodicStackTraceSampler` makes that use case easier to support.

Tests for that scenario are in `PeriodicStackTraceSamplerTest`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceThreadChangeDetector.java
```diff
@@ -0,0 +1,68 @@
...
+      return scope;
+    }
+
+    Thread thread = Thread.currentThread();
+    if (sampler.get().isBeingSampled(thread)) {
+      return scope;
+    }
+
+    sampler.get().start(thread, newSpanContext);
```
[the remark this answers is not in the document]
tduncan: Trying to think if this is a concern.

A sample should only be taken if 1) the trace is registered for snapshotting and 2) the thread isn't already being sampled. If those other scopes are for unregistered traces a sample won't be taken. If the scopes are for a registered trace that is already being sampled a new sample won't be taken.

So I suppose the danger is a profiled trace spawns hundreds (or more) background tasks at roughly the same time causing a flood of on demand samples to be taken.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceThreadChangeDetector.java
```diff
@@ -0,0 +1,68 @@
...
+  @Override
+  public Scope attach(Context toAttach) {
+    Scope scope = delegate.attach(toAttach);
+    SpanContext newSpanContext = Span.fromContext(toAttach).getSpanContext();
+    if (doNotTrack(newSpanContext)) {
+      return scope;
+    }
+
+    Thread thread = Thread.currentThread();
+    if (sampler.get().isBeingSampled(thread)) {
```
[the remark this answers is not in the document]
tduncan: I don't believe this is an issue. As long as the scope corresponding to the profiled trace is opened and closed appropriately any other scopes (whether higher or lower in the scope stack) shouldn't matter.

PR #2483, opened 2025-10-15
Update description for splunk.realm

laurit:
updated description was provided by the docs team


breedx-splk (approved):
Arguably worse, but ok.

PR #2498, opened 2025-10-27
[File based config] Support for Appd

robsunday:
This is support for AppD context. It includes service name and deployment environment name, as well as business transaction related data.
To make it working use the following snippet:
```yaml
instrumentation/development:
  java:
    cisco:
      ctx:
        enabled: true
```

This implementation utilizes BeforeAgentListener to copy service name and deployment environment name from resource attributes to corresponding fields of Appd context propagator. It is implemented this way for two main reasons:

1. Environment variable configuration [uses resource customizer for this task](https://github.com/signalfx/splunk-otel-java/blob/3ed744f78da9aae70e6ca47f8c69cb281263a8d5/custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusCustomizer.java#L60-L65). Resource customizers are not available in declarative configuration.
2. Declarative configuration resource detectors are run after all registered DeclarativeConfigurationCustomizerProvider already run. Resource detectors may affect service name and other resource attributes, so our code must run after all the detectors. I'm not aware of any other reliable place to put the code to be executed after SDK is initialized and before any signals are reported.

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusConfigurationCustomizerProvider.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusConfigurationCustomizerProvider.java
@@ -38,16 +38,10 @@ public final class AppdBonusConfigurationCustomizerProvider
   public void customize(DeclarativeConfigurationCustomizer autoConfiguration) {
     autoConfiguration.addModelCustomizer(
         model -> {
-          if (model.getInstrumentationDevelopment() == null
-              || model.getInstrumentationDevelopment().getJava() == null) {
+          if (!maybeAddAppdBonusPropagator(model)) {
             return model;
           }
-          Map<String, Object> properties =
-              model.getInstrumentationDevelopment().getJava().getAdditionalProperties();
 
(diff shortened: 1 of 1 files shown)


custom/src/main/java/com/splunk/opentelemetry/appd/AppdBonusConfigurationCustomizerProvider.java
```diff
@@ -0,0 +1,100 @@
...
+      compositeList = "";
+    } else if (!canAddPropagator(compositeList)) {
+      return false;
+    }
+
+    // Possible duplicates are handled by the upstream
+    compositeList =
+        compositeList.isEmpty()
+            ? AppdBonusPropagator.NAME
+            : AppdBonusPropagator.NAME + "," + compositeList;
```
laurit: Without declarative config we need to add the default propagators. Just checking whether this is also needed here or is it so that when you don't add propagators to yaml then you don't get any default ones?

breedx-splk (approved):
Had a couple small comments, but looks good to me.

PR #2505, opened 2025-10-29
[File based config] Initial configuration refactoring for profiler

robsunday:
Initial PR for profiler support with declarative config.
It is a cleanup and refactoring of profiler Configuration class.  This refactoring makes Configuration class reusable between env var config and declarative config approach.

There were some inconsistencies how code retrieved values from the config. In most places the code called utility methods form Configuration class, but in some places it called directly ConfigProperties::getXXX() instead (see JfrActivator, JfrSettingsOverride). Now the code consistently use Configuration class methods.


profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -20,74 +20,71 @@
...
+import java.util.function.Function;
 import java.util.logging.Logger;
 
-@AutoService(AutoConfigurationCustomizerProvider.class)
-public class Configuration implements AutoConfigurationCustomizerProvider {
+public class Configuration {
   private static final Logger logger = Logger.getLogger(Configuration.class.getName());
-  private static final boolean HAS_OBJECT_ALLOCATION_SAMPLE_EVENT = getJavaVersion() >= 16;
 
-  private static final String DEFAULT_RECORDING_DURATION = "20s";
```
robsunday: [for reviewer] This value is changed to avoid confusion due to discrepancy between env var config and declarative config data types accepted for duration properties.

Env var config supports nonstandard format with numeric value specified together with a time unit.
This is incompatible with the declarative config where specification clearly states that duration must be provided as an integer number.

laurit (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -20,74 +20,77 @@
...
+    }
+    log(CONFIG_KEY_CALL_STACK_INTERVAL, (it) -> getCallStackInterval(config));
+    log(CONFIG_KEY_INCLUDE_AGENT_INTERNALS, (it) -> getIncludeAgentInternalStacks(config));
+    log(CONFIG_KEY_INCLUDE_JVM_INTERNALS, (it) -> getIncludeJvmInternalStacks(config));
+    log(CONFIG_KEY_TRACING_STACKS_ONLY, (it) -> getTracingStacksOnly(config));
+    log(CONFIG_KEY_STACK_DEPTH, (it) -> getStackDepth(config));
+    logger.info("-----------------------");
   }
 
-  Map<String, String> defaultProperties() {
```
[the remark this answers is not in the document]
breedx-splk: The point is that you have one place to go where you can see all the defaults... but it's fine to remove.

profiler/src/main/java/com/splunk/opentelemetry/profiler/Configuration.java
```diff
@@ -20,74 +20,71 @@
...
   private static final Logger logger = Logger.getLogger(Configuration.class.getName());
-  private static final boolean HAS_OBJECT_ALLOCATION_SAMPLE_EVENT = getJavaVersion() >= 16;
 
-  private static final String DEFAULT_RECORDING_DURATION = "20s";
-  public static final boolean DEFAULT_MEMORY_ENABLED = false;
-  public static final Duration DEFAULT_CALL_STACK_INTERVAL = Duration.ofSeconds(10);
-  public static final boolean DEFAULT_INCLUDE_INTERNAL_STACKS = false;
-  public static final boolean DEFAULT_TRACING_STACKS_ONLY = false;
-  private static final int DEFAULT_STACK_DEPTH = 1024;
-  private static final boolean DEFAULT_MEMORY_EVENT_RATE_LIMIT_ENABLED = true;
```
breedx-splk: When someone asks what the default value behavior is for `foo`, you're going to have to go hunting now. I think this makes it way less findable, but whatever.

PR #2513, opened 2025-10-31
Add opamp client

laurit:
This pr adds an opamp client that just connects to the server and doesn't do anything else.

### DIFF
--- a/custom/src/test/java/com/splunk/opentelemetry/opamp/OpampActivatorTest.java
+++ b/custom/src/test/java/com/splunk/opentelemetry/opamp/OpampActivatorTest.java
@@ -64,6 +64,7 @@ class OpampActivatorTest {
 
   @Test
   void testOpamp() throws Exception {
+    // given
     Map<String, AgentConfigFile> configMap =
         Collections.singletonMap(
             "test-key",
@@ -104,8 +105,11 @@ class OpampActivatorTest {
             });
     cleanup.deferCleanup(client);
 
+    // when
     MessageData message = result.get(5, TimeUnit.SECONDS);
     AgentRemoteConfig remoteConfig = message.getRemoteConfig();
+
+    // then
     assertThat(remoteConfig).isNotNull();
     assertThat(remoteConfig.config.config_map.get("test-key").body.utf8()).isEqualTo("test-value");
   }


custom/src/test/java/com/splunk/opentelemetry/opamp/OpampActivatorTest.java
```diff
@@ -0,0 +1,112 @@
...
+    server.beforeTestExecution(null);
+  }
+
+  @AfterAll
+  static void cleanUp() {
+    server.stop();
+  }
+
+  @Test
+  void testOpamp() throws Exception {
```
robsunday: I'd love to see given, when, then sections in this test, but this is just my preference. It would make easier quick analysis of test scenario, because setup part is quite long here.

custom/src/test/java/com/splunk/opentelemetry/opamp/OpampActivatorTest.java
```diff
@@ -0,0 +1,112 @@
...
+                result.complete(messageData);
+              }
+            });
+
+    MessageData message = result.get(5, TimeUnit.SECONDS);
+    AgentRemoteConfig remoteConfig = message.getRemoteConfig();
+    assertThat(remoteConfig).isNotNull();
+    assertThat(remoteConfig.config.config_map.get("test-key").body.utf8()).isEqualTo("test-value");
+
+    client.close();
```
robsunday: It would be safer to put it some `finally`

custom/src/main/java/com/splunk/opentelemetry/opamp/OpampActivator.java
```diff
@@ -0,0 +1,105 @@
...
+    String serviceName = config.getString("otel.service.name");
+    if (serviceName != null) {
+      return serviceName;
+    }
+    Map<String, String> resourceAttributes = config.getMap("otel.resource.attributes");
+    serviceName = resourceAttributes.get(ServiceAttributes.SERVICE_NAME.getKey());
+    if (serviceName != null) {
+      return serviceName;
+    }
+    return resource.getAttribute(ServiceAttributes.SERVICE_NAME);
```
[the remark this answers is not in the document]
laurit: I copied from service name checker without thinking about it. I guess here we really don't need to do anything besides calling `resource.getAttribute(ServiceAttributes.SERVICE_NAME)`

custom/src/test/java/com/splunk/opentelemetry/opamp/OpampActivatorTest.java
laurit: usually we don't add these

breedx-splk (approved):

PR #2542, opened 2025-11-21
[File based config] Snapshot profiling span processor implementation

robsunday:
Snapshot profiling span processor can be used in YAML with the following snippet:
```yaml
tracer_provider:
  processors:
    - splunk_snapshot_profiling:
```

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessorComponentProvider.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessorComponentProvider.java
@@ -21,10 +21,8 @@ import io.opentelemetry.api.incubator.config.DeclarativeConfigProperties;
 import io.opentelemetry.sdk.autoconfigure.spi.internal.ComponentProvider;
 import io.opentelemetry.sdk.trace.SpanProcessor;
 
-@SuppressWarnings("rawtypes")
 @AutoService(ComponentProvider.class)
-public class SnapshotProfilingSpanProcessorComponentProvider
-    implements ComponentProvider<SpanProcessor> {
+public class SnapshotProfilingSpanProcessorComponentProvider implements ComponentProvider {
   private final TraceRegistry traceRegistry;
 
   public SnapshotProfilingSpanProcessorComponentProvider() {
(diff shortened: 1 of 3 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessorComponentProvider.java
```diff
@@ -0,0 +1,51 @@
...
+public class SnapshotProfilingSpanProcessorComponentProvider implements ComponentProvider {
+  private final TraceRegistry traceRegistry;
+
+  public SnapshotProfilingSpanProcessorComponentProvider() {
+    this(TraceRegistryHolder.getTraceRegistry());
+  }
+
+  SnapshotProfilingSpanProcessorComponentProvider(TraceRegistry traceRegistry) {
+    this.traceRegistry = traceRegistry;
+  }
```
breedx-splk: I don't think that you actually need to hold this global static in a field on this class -- just call `TraceRegistryHolder.getTraceRegistry()` in the `create()` method below right?

robsunday: The clue of `TraceRegistryHolder` class and static instance of `TraceRegistry` is to share this instance between `SnapshotProfilingConfigurationCustomizerProvider` and `SnapshotProfilingSpanProcessorComponentProvider`, which is impossible via YAML and `ComponentProvider` SPI.
This is not obvious and I should make a comment about it in code review initially.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceRegistryHolder.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler.snapshot;
+
+class TraceRegistryHolder {
```
laurit: Rest of the code uses `ConfigurableSupplier` in similar situations. I have a feeling that the existing code might be overcomplicating things we may need to revisit it at one point and see whether it could be improved.

PR #2550, opened 2025-12-05
[File based config] Use of dedicated .distribution node for Splunk distribution specific configuration

robsunday:
In declarative config there is a need to introduce a common location and format of Splunk specific configuration. This is aligned with discussion that take place in the community.

Community discussion about introducing vendor specific YAML configuration node is here: [Where should Vendor-Specific General Configuration be located?](https://github.com/open-telemetry/opentelemetry-configuration/issues/336)
Followup update of OpenTelemetry specification is here: [[declarative config] Support for distribution specific config](https://github.com/open-telemetry/opentelemetry-specification/issues/4770)
Splunk specific config format is discussed here: [Splunk specific file based configuration fields](https://github.com/signalfx/gdi-specification/pull/355)

**This PR contains the first part of common Splunk specific config support:**
- Introduction of `distribution` node to keep Splunk distribution specific config.
  Initially it contains profiler config only.
- Introduction of common Splunk config format for profiler. It required significant profiler config classes refactoring. This refactoring also enforced refactoring of client classes.

At the moment the `distribution` node is temporary located under `.instrumentation/development.java`. This will be moved to the root of YAML file once ConfigProvider has required changes implemented in the upstream. If this will take too long then I'll move it to the root and implement temporary `DeclarativeConfigurationCustomizer` that will process `.distribution` node.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
@@ -70,21 +70,7 @@ public class JfrActivator implements AgentListener {
 
   @Override
   public void afterAgent(AutoConfiguredOpenTelemetrySdk sdk) {
-    ProfilerConfiguration config;
-
-    if (AutoConfigureUtil.isDeclarativeConfig(sdk)) {
-      DeclarativeConfigProperties distributionConfig = AutoConfigureUtil.getDistributionConfig(sdk);
-      distributionConfig = Optional.ofNullable(distributionConfig).orElse(empty());
-      config =
-          new ProfilerDeclarativeConfiguration(
-              distributionConfig
-                  .getStructured("splunk", empty())
-                  .getStructured("profiling", empty())
-                  .getStructured("always_on", empty()));
-    } else {
-      ConfigProperties configProperties = AutoConfigureUtil.getConfig(sdk);
-      config = new ProfilerEnvVarsConfiguration(configProperties);
-    }
+    ProfilerConfiguration config = getProfilerConfiguration(sdk);
 
     if (notClearForTakeoff(config)) {
       return;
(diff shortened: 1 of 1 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilerConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+
+  int getStackDepth();
+
+  boolean getKeepFiles();
+
+  String getProfilerDirectory();
+
+  Duration getRecordingDuration();
+
+  Object getConfigProperties();
```
robsunday: This method returns instance of an Object class because there is no connection between `io.opentelemetry.api.incubator.config.DeclarativeConfigProperties` and `io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties`

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrActivator.java
```diff
@@ -67,23 +69,40 @@ public JfrActivator() {
...
+      config =
+          new ProfilerDeclarativeConfiguration(
+              distributionConfig
+                  .getStructured("splunk", empty())
+                  .getStructured("profiling", empty())
+                  .getStructured("always_on", empty()));
+    } else {
+      ConfigProperties configProperties = AutoConfigureUtil.getConfig(sdk);
+      config = new ProfilerEnvVarsConfiguration(configProperties);
+    }
```
breedx-splk: Factor out a method called `getProfilingConfiguration()` or similar.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilerEnvVarsConfiguration.java
```diff
@@ -52,44 +52,55 @@ public class Configuration {
...
 
-  public static void log(ConfigProperties config) {
+  private final ConfigProperties config;
+
+  public ProfilerEnvVarsConfiguration(ConfigProperties config) {
+    this.config = config;
+  }
+
+  @Override
+  public void log() {
```
breedx-splk: Can this implementation be shared between the two? Seems like it's effectively duplicated...and if they're only calling interface methods should be possible...

breedx-splk (approved):
Looks good to me. Had a few remarks, but nothing too intense. Thanks!!

laurit (approved):

PR #2556, opened 2025-12-11
Prevent 0 probability for snapshots

breedx-splk:
Aligning with https://github.com/signalfx/gdi-specification/pull/358

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingConfiguration.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingConfiguration.java
@@ -71,6 +71,14 @@ public class SnapshotProfilingConfiguration {
                 + "'");
         return MAX_SELECTION_PROBABILITY;
       }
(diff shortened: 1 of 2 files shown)


laurit (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingConfiguration.java
robsunday: I know this is not in the intial scope of this PR, but is it a valid MAX? [Specification](https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#snapshot-profiler) says max is 1.0

robsunday: It would be great if you could do this fix as well. If not then we will need to make a separate issue and PR for it.

breedx-splk: Separate PR please.

PR #2595, opened 2026-01-20
[File based config] SnapshotVolumePropagator component config taken form distribution node

robsunday:
Common Splunk profiling declarative config contains setting for snapshot `selection_probability` that is needed by `SnapshotVolumePropagator`.
**This is current format introduced with this PR**:
```yaml
# Required for callgraphs
propagator:
  composite:
    - tracecontext:
    - baggage:
    - splunk_snapshot_volume:
distribution:
  splunk:
    profiling:
      callgraphs:
        selection_probability: 0.123
        sampling_interval: 5
```
Unfortunately, there is one issue with this solution.
Convention established by declarative config recommends putting the component's config under the component's node in YAML. With this convention it should look like as follows:
```yaml
# Required for callgraphs
propagator:
  composite:
    - tracecontext:
    - baggage:
    - splunk_snapshot_volume:
       selection_probability: 0.123
distribution:
  splunk:
    profiling:
      callgraphs:
        sampling_interval: 5
```

It is incompatible with common Splunk distribution config idea, but I think that common config has a great value for the users.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotVolumePropagatorComponentProvider.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotVolumePropagatorComponentProvider.java
@@ -17,14 +17,14 @@
 package com.splunk.opentelemetry.profiler.snapshot;
 
 import com.google.auto.service.AutoService;
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


robsunday:
I'd really suggest rethinking if automatic adding `splunk_snapshot_volume` propagator in model customizer wouldn't make sense. Without this propagator the call graphs will not work anyway. So why force user to add one more line in YAML? This would also somehow hide this inconsistency, I mentioned in the description.

PR #2614, opened 2026-01-29
OpAMP client: Add additional identifying and recommended non-identifying res attrs

breedx-splk:
Here is some additional information from the spec about what should be placed into the opamp client request:

https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes


custom/src/test/java/com/splunk/opentelemetry/opamp/OpampActivatorTest.java
```diff
@@ -82,7 +86,7 @@ void testOpamp() throws Exception {
     OpampClient client =
         OpampActivator.startOpampClient(
             server.httpUri().toString(),
-            "test",
+            resource,
```
breedx-splk: I really wish that this `MockWebServerExtension` that we're using had some means of fetching a request body in order to make assertions about it...but it doesn't look possible.  This means that this new code goes untested or we find another mocking server/harness or something.

laurit: try `server.takeRequest().request().contentUtf8()` or `server.takeRequest().request().content().array()`

PR #2648, opened 2026-02-23
CSA build updates

breedx-splk:
We're going to eliminate the manual step of putting a csa release in github. So this will use the internal artifactory instead.


agent-csa-bundle/build.gradle.kts
```diff
@@ -117,6 +123,9 @@ tasks {
   }
 
   jar {
+    onlyIf("csa is only built from gitlab") {
```
robsunday: Wouldn't putting this condition only here be sufficient?
Maybe there is no need to repeat it in all the tasks that this top level task depends on.

breedx-splk: I thought I had tried it and the answer was no....but let's try it again....

PR #2652, opened 2026-02-24
Tests that use autoconfigure need to reset

breedx-splk:
I was seeing test failures in `NocodeInitializerTest` due to the global state being held in the `AgentDistributionConfig`. This reset after each test clears that up.


robsunday (approved):
Good catch!

PR #2663, opened 2026-02-27
Set petclinic rest version to 2.6.2

laurit:
Current image seems to be also using 2.6.2. Setting the version should let us rebuild the images using newer base image. Currently we can't rebuild the images because current petclinic-rest requires jdk17.


breedx-splk (approved):
Yeah I think this is fine...but it's a little worrisome that it's so old (Aug. 2024).

I guess we can't use jdk17 for this because we have some java8 stuff we're still testing with?

PR #2675, opened 2026-03-02
Add support for configuring opamp polling duration

breedx-splk:
Relates to https://github.com/signalfx/gdi-specification/pull/375

This will create parity with other languages.

Also removed the `enableRemoteConfig()` call, because this could be interpreted to mean that the agent already has this capability...which it does not.

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/opamp/OpampActivator.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/opamp/OpampActivator.java
@@ -108,8 +108,6 @@ public class OpampActivator implements AgentListener {
       HttpRequestService httpSender =
           HttpRequestService.create(okhttp, pollingDelay, DEFAULT_DELAY_BETWEEN_RETRIES);
       builder.setRequestService(httpSender);
-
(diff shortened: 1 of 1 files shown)


custom/src/main/java/com/splunk/opentelemetry/opamp/OpampActivator.java
```diff
@@ -84,11 +93,22 @@ public void onMessage(OpampClient opampClient, MessageData messageData) {}
...
+    // builder.enableEffectiveConfigReporting();
     if (endpoint != null) {
+      PeriodicDelay pollingDelay =
+          PeriodicDelay.ofFixedDuration(Duration.ofMillis(pollingDurationMillis));
+      OkHttpSender okhttp = OkHttpSender.create(endpoint);
+      HttpRequestService httpSender =
+          HttpRequestService.create(okhttp, pollingDelay, DEFAULT_DELAY_BETWEEN_RETRIES);
+      builder.setRequestService(httpSender);
+
       builder.setRequestService(HttpRequestService.create(OkHttpSender.create(endpoint)));
```
laurit:
```suggestion
```

robsunday (approved):
Looks good to me

PR #2680, opened 2026-03-03
Build petclinic image for jdk 21 and 25

laurit:


breedx-splk:
Are we still using this image? Should we consider not publishing it any longer?

PR #2696, opened 2026-03-10
Proxy replaced with delegating config properties

robsunday:
This is alternative for Proxy use introduced by https://github.com/signalfx/splunk-otel-java/pull/2695
The class implementing `ExtendedDeclarativeConfigProperties` interface is used instead of the Proxy.
Benefit from proposed solution is that any breaking changes made to ExtendedDeclarativeConfigProperties interface will raise some compilation error. Proxy based solution may get silently broken when `ExtendedDeclarativeConfigProperties.getConfigProvider` is renamed or removed.


profiler/src/main/java/com/splunk/opentelemetry/profiler/LogExporterBuilder.java
```diff
@@ -116,25 +117,94 @@ static LogRecordExporter buildHttpExporter(
...
+    private final ConfigProvider configProvider;
+
+    ExtendedDeclarativeConfigPropertiesImpl(
+        DeclarativeConfigProperties delegate, ConfigProvider configProvider) {
+      this.delegate = delegate;
+      this.configProvider = configProvider;
+    }
+
+    @Override
+    public ConfigProvider getConfigProvider() {
```
breedx-splk: Too bad java doesn't have first-class delegation support like kotlin. 😭

Issue #2689, opened 2026-03-06
Unsatisfactory behavior of snapshot selector


PeterF778:
There are at least three issues with the mechanism for selecting traces for collecting "snapshots":

- When dealing with distributed traces, the selection for the head service is inconsistent with the other services. Running a test where service A calls B, and choosing selection probability of 10%, I got the following results. Out of 4781 traces, service A was selected 480 times, and service B was selected 448 times. However, the number of traces where both services were selected was only 39.
- Traces which originate from spans different than SERVER or CONSUMER (like resulting from POJO instrumentation) are never selected (however, their downstream calls may still be selected).
- The selection algorithm for downstream services uses the same algorithm as TraceIdRatioBased sampler, which can lead to metrics skew if that sampler is actually used for sampling.

I believe the selection mechanism needs to be redesigned.

laurit:
Resolved with https://github.com/signalfx/splunk-otel-java/pull/2717

PR #2697, opened 2026-03-10
Issue #2689: Unsatisfactory behavior of snapshot selector

PeterF778:
Change the algorithm for snapshot profiling selection to be exclusively based on trace-id. Removing the concepts of snapshot Volume, SnapshotVolumePropagator, and ProbabilisticSnapshotSelector. Updating unit tests.

### DIFF
--- a/.github/config/lychee.toml
+++ b/.github/config/lychee.toml
@@ -1,3 +1,6 @@
+# Lychee configuration file
+# See https://lychee.cli.rs/guides/config/
+
 timeout = 30
 retry_wait_time = 5
 max_retries = 6
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,19 +13,9 @@ on:
 
 jobs:
   check-links:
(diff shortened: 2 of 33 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceIdBasedSnapshotSelector.java
```diff
@@ -16,48 +16,46 @@
...
+  }
+
+  private static String thresholdFor(double probability) {
+    int threshold = (int) (probability * MULTIPLIER);
+    if (threshold == 0) {
+      // Zero or near-zero probability, special case
+      return null;
+    }
+    String hexValue = String.format(HEX_FORMATTER, threshold);
+    return hexValue;
```
breedx-splk: the variable seems unnecessary
```suggestion
    return String.format(HEX_FORMATTER, threshold);
```

robsunday (approved):

PR #2726, opened 2026-03-23
Add current_span option to nocode instrumentation

johnbley:
Add a new `current_span: true` option to nocode rules, allowing you to add attributes to the currently active span by adding runtime instrumentation to a method as before.  This makes it easier to support usage like capturing a cookie value on a root servlet span, without needing to create a child span just to have the cookie value.

- `span_name`, `span_kind`, and `span_status` have no effect (besides logging a parse-time warning about this) on current-span rules
- if no span is active/current at runtime, nothing happens
- adding both span-creating/"regular"and current-span rules to a single method is supported and the order in the yaml config is the order they are evaluated at runtime

This PR was created by codex 5.4 extra-high from my spec (listed in comment for completeness), with some light tweaking/tuning/commenting by me.


laurit (approved):

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
```diff
@@ -37,26 +85,41 @@ public void onStart(
...
+    applyRuleAttributes(
+        mi, (key, value) -> setAttribute(attributesBuilderSetter, attributesBuilder, key, value));
+  }
+
+  static void applyToSpan(Span span, NocodeMethodInvocation mi) {
+    applyRuleAttributes(mi, (key, value) -> setAttribute(spanAttributeSetter, span, key, value));
+  }
+
+  private static void applyRuleAttributes(
+      NocodeMethodInvocation mi, BiConsumer<String, Object> attributeConsumer) {
```
breedx-splk: nit: `mi` is a poor variable name. How about simply `invocation`?

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
```diff
@@ -87,7 +87,11 @@ private List<NocodeRules.Rule> loadUnsafe(List<Map<String, Object>> ruleListNode
...
       ElementMatcher<MethodDescription> methodMatcher = methodMatcher(yamlRule.get("method"));
-      NocodeExpression spanName = toExpression(yamlRule.get("span_name"));
+      boolean isCurrentSpan = isCurrentSpanRule(yamlRule.get("current_span"));
+      NocodeExpression spanName =
+          isCurrentSpan
+              ? ignoreCurrentSpanField("span_name", yamlRule.get("span_name"))
+              : toExpression(yamlRule.get("span_name"));
       SpanKind spanKind = null;
       if (yamlRule.get("span_kind") != null) {
         String spanKindString = yamlRule.get("span_kind").toString();
```
breedx-splk, line 99: Does this mean you CAN change the span `kind` with this? Seems goofy, but do we care?

breedx-splk (approved):
A couple of nits but looks good.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/NocodeAttributesExtractor.java
johnbley: Yep, done.

instrumentation/nocode/src/main/java/com/splunk/opentelemetry/instrumentation/nocode/YamlParser.java
johnbley: Actually, it doesn't by implementation, but that's a good catch for documenting behavior and doing the same as name and status - log a parse-time warning about it not having any effect.  Working on that now.

PR #2749, opened 2026-04-07
Rework advice

laurit:
In upstream agent there is plan to transition from inline to non-inline advice. In non-inline advice some constructs such as `@Advice.Local` or `readOnly = false` are not available. This PR rewrites the advice so that they are ready to be used as non-inline but still work as inline for now. While some parts of the advice will be slightly more complicated non-inline advice may open up other options, such as advice having access to agent code without the need to go through boot loader, that make it worth it.

### DIFF
--- a/instrumentation/khttp/src/main/java/com/splunk/opentelemetry/instrumentation/khttp/RequestWrapper.java
+++ b/instrumentation/khttp/src/main/java/com/splunk/opentelemetry/instrumentation/khttp/RequestWrapper.java
@@ -33,6 +33,10 @@ public class RequestWrapper {
     this.parsedUri = parseUri(uri);
   }
 
+  public Map<String, String> getHeaders() {
+    return headers;
+  }
+
   private static URI parseUri(String uri) {
     if (uri == null) {
       return null;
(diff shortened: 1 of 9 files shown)


instrumentation/khttp/src/main/java/com/splunk/opentelemetry/instrumentation/khttp/RequestWrapper.java
```diff
@@ -33,6 +33,10 @@ public RequestWrapper(String method, String uri, Map<String, String> headers) {
     this.parsedUri = parseUri(uri);
   }
 
+  public Map<String, String> getHeaders() {
+    return headers;
```
robsunday: It would be safer to return unmodifiable map, but it is optional I think.

laurit: Usually in instrumentation we don't do such changes. Here returning an unmodifiable map would probably be fine because the comments in code indicate that we are replacing the headers map in first place because it might not be mutable.

robsunday: Yeah, leave it as it is then

PR #2765, opened 2026-04-15
[OpAMP] Full effective config reported for env vars

robsunday:
This is first attempt to report complete set of env vars used to configure the agent. This set is based on content of https://help.splunk.com/en/splunk-observability-cloud/manage-data/available-data-sources/supported-integrations-in-splunk-observability-cloud/apm-instrumentation/instrument-a-java-application/configure-the-java-agent with some minor discrepancies fixed.

There are still some open questions so for sure next PRs will come.
In this PR `EnvVarsEffectiveConfigFileFactory` replicates some pieces of configuration logic mostly from profiler and snapshot profiler configurations.
This logic is usually used to create a default value if some property is not provided. We need to decide how we want to handle it.


custom/src/main/java/com/splunk/opentelemetry/opamp/EnvVarsEffectiveConfigFileFactory.java
```diff
@@ -0,0 +1,471 @@
...
+        .add("otel.traces.sampler", "always_on")
+        .add("otel.traces.sampler.arg", "");
+  }
+
+  @VisibleForTesting
+  static String toEnvVarName(String name) {
+    return name.toUpperCase().replace('.', '_').replace('-', '_');
+  }
+
+  private static String getProfilerLogsEndpoint(ConfigProperties config) {
```
robsunday, line 314: [for reviewer] Do we want this logic?

PR #2766, opened 2026-04-16
[OpAMP] Dedicated module and full effective config reported for env vars

robsunday:
This is first attempt to report complete set of env vars used to configure the agent. This set is based on content of https://help.splunk.com/en/splunk-observability-cloud/manage-data/available-data-sources/supported-integrations-in-splunk-observability-cloud/apm-instrumentation/instrument-a-java-application/configure-the-java-agent with some minor discrepancies fixed.

There are still some open questions so for sure next PRs will come.
In this PR EnvVarsEffectiveConfigFileFactory replicates some pieces of configuration logic mostly from profiler and snapshot profiler configurations.
This logic is usually used to create a default value if some property is not provided. We need to decide how we want to handle it.

This PR also introduces **dedicated module for OpAMP** support which isolates OpAMP related code and also solves circular dependency issue that occurred when attempting to call profiler configuration code from :custom module.


opamp/src/main/java/com/splunk/opentelemetry/opamp/EnvVarsEffectiveConfigFileFactory.java
```diff
@@ -0,0 +1,508 @@
...
+      return addValue(
+          propertyName, valueTransformer.apply(config.getString(propertyName, defaultValue)));
+    }
+
+    String build() {
+      return stringBuilder.toString();
+    }
+
+    private String nullable(Object value) {
+      return value == null ? "" : value.toString();
```
laurit: should be same as `Objects.toString(value, "")`

opamp/src/main/java/com/splunk/opentelemetry/opamp/EnvVarsEffectiveConfigFileFactory.java
```diff
@@ -0,0 +1,505 @@
...
+        .add("otel.instrumentation.hystrix.experimental-span-attributes", false)
+        .add("otel.instrumentation.java-util-logging.experimental-log-attributes", false)
+        .add("otel.instrumentation.jaxrs.experimental-span-attributes", false)
+        .add("otel.instrumentation.jboss-logmanager.experimental.capture-event-name", false)
+        .add("otel.instrumentation.jboss-logmanager.experimental.capture-mdc-attributes", "")
+        .add("otel.instrumentation.jboss-logmanager.experimental-log-attributes", false)
+        .add("otel.instrumentation.jdbc.experimental.capture-query-parameters", false)
+        .add("otel.instrumentation.jdbc.experimental.sqlcommenter.enabled", false)
+        .add("otel.instrumentation.jdbc.experimental.transaction.enabled", false)
+        .add("otel.instrumentation.jdbc.statement-sanitizer.enabled", true)
```
laurit, line 263: some of these have complex defaults like jdbc sanitizer defaults to the value of `common.db-statement-sanitizer` which probably makes accurately reporting this too hard

laurit:
> This set is based on content of https://help.splunk.com/en/splunk-observability-cloud/manage-data/available-data-sources/supported-integrations-in-splunk-observability-cloud/apm-instrumentation/instrument-a-java-application/configure-the-java-agent with some minor discrepancies fixed.

that page is generated from the metadata we provide which unfortunately is not completely up to date

laurit (approved):

opamp/src/main/java/com/splunk/opentelemetry/opamp/EnvVarsEffectiveConfigFileFactory.java
robsunday: Right. In some cases there is some logic behind default value calculation. Thankfully sometimes we have it wrapped already in methods than we just can call, like methods from `ProfilerEnvVarsConfiguration`. At some point we will have to discuss general approach to complex defaults

robsunday:
> that page is generated from the metadata we provide which unfortunately is not completely up to date

Do we have any reliable source of truth for this? Anyway, the docs need to be reviewed and updated.

laurit:
> Do we have any reliable source of truth for this? Anyway, the docs need to be reviewed and updated.

I'm working on this.

opamp/src/main/java/com/splunk/opentelemetry/opamp/EnvVarsEffectiveConfigFileFactory.java
laurit: I personally think that it is futile. Might try to handle some well know values like otlp exporters and splunk specific properties but for rest might as well extract all keys that start with `otel` or `otel.instrumenation` from the config properties and send them as they are.

PR #2767, opened 2026-04-16
Upload built agent for pr builds

laurit:


.github/workflows/pr.yaml
```diff
@@ -37,6 +37,13 @@ jobs:
         if: ${{ !cancelled() && hashFiles('build-scan.txt') != '' }}
         run: cat build-scan.txt
 
+      - name: Upload agent jar
```
robsunday, line 40: As I understand, this will upload agent binary generated during PR build somewhere.
Will it be downloadable from the web?

robsunday (approved):

.github/workflows/pr.yaml
laurit: Have a look at the bottom of the build summary page https://github.com/signalfx/splunk-otel-java/actions/runs/24515305776?pr=2767

robsunday: cool!

PR #2771, opened 2026-04-17
Update splunk instrumentation metadata

laurit:


instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/JvmMetricsInstaller.java
```diff
@@ -33,7 +33,7 @@ public class JvmMetricsInstaller implements AgentListener {
   public void afterAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk) {
     ConfigProperties config = getConfig(autoConfiguredOpenTelemetrySdk);
     boolean metricsEnabled = config.getBoolean(PROFILER_MEMORY_ENABLED_PROPERTY, false);
-    if (!config.getBoolean("otel.instrumentation.jvm-metrics.splunk.enabled", metricsEnabled)) {
+    if (!config.getBoolean("otel.instrumentation.jvm-metrics-splunk.enabled", metricsEnabled)) {
```
laurit, line 36: old and new property normalize to the same value so the old property will still work

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -3824,15 +3827,11 @@ Some instrumentations (e.g. Spring Web MVC instrumentation) produce SpanKind.Int
     instrumentations.add(
         splunkInstrumentation("liberty").component("Liberty", "20.0 and higher").build());
     instrumentations.add(
-        splunkInstrumentation(List.of("tomcat", "tomcat-metrics-splunk"))
```
laurit, line 3827: Doesn't really do anything with metrics (maybe the metrics part was for tomcat connection pool that is now upstream?).

robsunday (approved):

PR #2773, opened 2026-04-17
Instrumentation metadata

laurit:
Use https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/instrumentation-list.yaml to generate instrumentation metadata. Move instrumentation settings under instrumentations like in dotnet. See https://help.splunk.com/en/splunk-observability-cloud/manage-data/available-data-sources/supported-integrations-in-splunk-observability-cloud/apm-instrumentation/instrument-a-.net-application/requirements and https://help.splunk.com/en/splunk-observability-cloud/manage-data/available-data-sources/supported-integrations-in-splunk-observability-cloud/apm-instrumentation/instrument-a-java-application/requirements
Due to the way upstream metadata is structured when different versions of the framework have different instrumentations we'll list them all. Known issues in upstream metadata are
- runtime telemetry includes experimental metrics that are not enabled by default
- some instrumentations use `otel.instrumentation.servlet.capture-request-parameters`, probably should be `otel.instrumentation.servlet.experimental.capture-request-parameters`
At least we don't have to manually keep this up to date any more.

### DIFF
--- a/metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
+++ b/metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
@@ -18,14 +18,21 @@ package com.splunk.opentelemetry.tools;
 
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.Writer;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
@@ -33,6 +40,25 @@ public class MetadataGenerator {
   private static final String BUNDLED_METRIC =
       "APM bundled, if data points for the metric contain `telemetry.sdk.language` attribute.";
   private static final String CUSTOM_METRIC = "Custom metric.";
+
+  // java instrumentation metrics marked as bundled by mts-category-generator
+  private static final Set<String> bundledMetrics =
+      new HashSet<>(
+          Arrays.asList(
+              "http.client.request.duration",
+              "http.server.request.duration",
+              "jvm.class.loaded",
+              "jvm.class.unloaded",
+              "jvm.class.count",
+              "jvm.cpu.time",
+              "jvm.cpu.count",
+              "jvm.cpu.recent_utilization",
+              "jvm.gc.duration",
+              "jvm.memory.used",
+              "jvm.memory.committed",
+              "jvm.memory.limit",
+              "jvm.memory.used_after_last_gc",
+              "jvm.thread.count"));
 
   private static String getRequiredSystemProperty(String name) {
     String value = System.getProperty(name);
@@ -866,7 +892,7 @@ public class MetadataGenerator {
             SettingCategory.INSTRUMENTATION));
 
     // DB statement sanitization
-    // https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#db-statement-sanitization
+    // https://opentelemetry.io/docs/zero-code/java/agent/instrumentation/#db-statement-sanitization
 
     /*
     System property: otel.instrumentation.common.db-statement-sanitizer.enabled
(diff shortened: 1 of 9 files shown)


metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -866,7 +892,7 @@ otel.exporter.otlp.metrics.temporality.preference	OTEL_EXPORTER_OTLP_METRICS_TEM
             SettingCategory.INSTRUMENTATION));
 
     // DB statement sanitization
-    // https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#db-statement-sanitization
+    // https://opentelemetry.io/docs/zero-code/java/agent/instrumentation/#db-statement-sanitization
```
laurit, line 895: There are probably other pages that have changed. I think we should try to tackle this in a subsequent PR.

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -1413,2441 +1439,151 @@ Some instrumentations (e.g. Spring Web MVC instrumentation) produce SpanKind.Int
...
-    instrumentations.add(
-        instrumentation(
-                List.of("kafka", "kafka-clients-metrics"),
-                "Consult documentation of the used kafka version for the exact metrics it generates.")
-            .component("Apache Kafka Producer/Consumer API", "0.11 and higher")
-            .component("Apache Kafka Streams API", "0.11 and higher")
-            .component("Apache Kafka Connect API", "2.6 and higher")
-            // https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/kafka/kafka-clients/kafka-clients-2.6/library/README.md
-            // these metrics are created by kafka, we only collect these as otel metrics
-            .customMetric(
```
laurit, line 2554: Upstream metadata doesn't include these. The exact metrics you'll get depends on the kafka client version. I think the plan is to disable the kafka metrics by default in the next major release.

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -4528,12 +2249,12 @@ Map<String, Object> build() {
       if (!metrics.isEmpty()) {
         signals.add(Collections.singletonMap("metrics", metrics));
       }
-      if (!dependencies.isEmpty()) {
```
laurit, line 4531: we never add dependencies to the instrumentations (I believe resource providers use this) so removed it

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -4221,6 +1957,93 @@ Some instrumentations (e.g. Spring Web MVC instrumentation) produce SpanKind.Int
...
+  }
+
+  private static List<Map<String, Object>> parseInstrumentations(URL url) throws IOException {
+    Yaml yaml = new Yaml();
+    Map<String, Object> metadata;
+    try (InputStream inputStream = url.openStream()) {
+      metadata = yaml.load(inputStream);
+    }
+
+    if (!"0.5".equals(metadata.get("file_format").toString())) {
```
robsunday, line 1982: Any chance it can be 0.5.1, which in theory should be backward compatible with 0.5 ?

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -4238,6 +2061,11 @@ private static Map<String, Object> setting(
       String defaultValue,
       SettingType type,
       SettingCategory category) {
+    if ("otel.instrumentation.spring-batch.item.enabled".equals(property)) {
+      // SplunkConfiguration overrides the default value of this settings to true
```
robsunday, line 2074: That is not the case for declarative config. It must be placed in yaml explcitly.

Also there is otel.instrumentation.spring-batch.enabled as well defaulting for env var config schema to `true`

robsunday (approved):
There are lots of changes hard to validate manually, but it looks good with few small comments.

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
laurit: Not really. When the format changes we expect to manually validate that the new version is fine.

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
laurit: > That is not the case for declarative config. It must be placed in yaml explcitly.

It is fine. This option was added for a certain customer that is not a customer any more. We can disable it by default in the next major release.

> Also there is otel.instrumentation.spring-batch.enabled as well defaulting for env var config schema to true

This is a bit annoying. Upstream metadata has a `disabled_by_default` tag. Perhaps we should append `This instrumentation is disabled by default.` to the description of the disabled instrumentations.

PR #2774, opened 2026-04-17
Fix of NPE when declarative config is in use

### LINKED ISSUE #2772: NPE when declarative config in use.
Agent is not starting when declarative config is in use.
Upstream in the most recent release 1.61.0 introduced a breaking change. AutoConfiguredOpenTelemetrySdk.getConfig() now returns null if declarative config is in use. It causes NPE in few places of our code preventing agent from starting.

Planned changes:
1) Add a test that detects agent is starting correctly with env var based config and with declarative config
2) Fix all the places where NPE is thrown

robsunday:
Fixes #2772

### DIFF
--- a/custom/src/main/java/com/splunk/opentelemetry/servicename/ServiceNameChecker.java
+++ b/custom/src/main/java/com/splunk/opentelemetry/servicename/ServiceNameChecker.java
@@ -48,14 +48,17 @@ public class ServiceNameChecker implements BeforeAgentListener {
 
   @Override
   public void beforeAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk) {
-    ConfigProperties config = getConfig(autoConfiguredOpenTelemetrySdk);
     Resource resource = getResource(autoConfiguredOpenTelemetrySdk);
-    if (serviceNameNotConfigured(config, resource)) {
-      if (AutoConfigureUtil.isDeclarativeConfig(autoConfiguredOpenTelemetrySdk)) {
+
+    if (AutoConfigureUtil.isDeclarativeConfig(autoConfiguredOpenTelemetrySdk)) {
+      if (serviceNameNotConfigured(resource)) {
         logWarn.accept(
             "The service.name resource attribute is not set. Your service is unnamed and will be difficult to identify.\n"
                 + " Set your service name in '.resource.attributes' node, or specify appropriate resource detector in the configuration YAML file.");
-      } else {
+      }
+    } else {
+      ConfigProperties config = getConfig(autoConfiguredOpenTelemetrySdk);
+      if (serviceNameNotConfigured(config, resource)) {
         logWarn.accept(
             "The service.name resource attribute is not set. Your service is unnamed and will be difficult to identify.\n"
                 + " Set your service name using the OTEL_SERVICE_NAME or OTEL_RESOURCE_ATTRIBUTES environment variable.\n"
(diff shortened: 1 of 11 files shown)


instrumentation/jvm-metrics/build.gradle.kts
```diff
@@ -5,6 +5,10 @@ plugins {
 dependencies {
   compileOnly(project(":custom"))
   compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
+
+  testImplementation(project(":custom"))
+  testCompileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
```
laurit: wouldn't `testImplementation` cover both `testCompileOnly` and `testRuntimeOnly`? Is this dependency needed at all?

custom/src/test/java/com/splunk/opentelemetry/appd/AppdBeforeAgentListenerTest.java
```diff
@@ -97,26 +98,30 @@ void shouldSetPropagatorProperties_envVarConfig() {
         Resource.create(
             Attributes.of(
                 DEPLOYMENT_ENVIRONMENT_NAME, "test-deployment-env", SERVICE_NAME, "test-service"));
-    when(AutoConfigureUtil.getConfig(sdkMock)).thenReturn(mockConfigProperties);
-    when(AutoConfigureUtil.getResource(sdkMock)).thenReturn(resource);
+    try (MockedStatic<AutoConfigureUtil> autoConfigureUtil = mockStatic(AutoConfigureUtil.class)) {
```
laurit: could consider using `autoCleanup.deferCleanup` to avoid having to warp everything in `try`

laurit (approved):

instrumentation/jvm-metrics/build.gradle.kts
robsunday: Good point.

custom/src/test/java/com/splunk/opentelemetry/appd/AppdBeforeAgentListenerTest.java
robsunday: Yes, it looks nicer now

custom/src/test/java/com/splunk/opentelemetry/servicename/ServiceNameCheckerTest.java
```diff
@@ -16,67 +16,128 @@
...
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
-@ExtendWith(MockitoExtension.class)
 class ServiceNameCheckerTest {
-  @Mock Consumer<String> logWarn;
+  @RegisterExtension final AutoCleanupExtension autoCleanup = AutoCleanupExtension.create();
```
laurit:
```suggestion
  @RegisterExtension static final AutoCleanupExtension autoCleanup = AutoCleanupExtension.create();
```

custom/src/main/java/com/splunk/opentelemetry/servicename/ServiceNameChecker.java
```diff
@@ -70,11 +74,25 @@ public int order() {
...
   }
 
-  static boolean serviceNameNotConfigured(ConfigProperties config, Resource resource) {
-    String serviceName = config.getString("otel.service.name");
-    Map<String, String> resourceAttributes = config.getMap("otel.resource.attributes");
-    return serviceName == null
-        && !resourceAttributes.containsKey(ServiceAttributes.SERVICE_NAME.getKey())
-        && "unknown_service:java".equals(resource.getAttribute(ServiceAttributes.SERVICE_NAME));
+  @VisibleForTesting
+  static boolean isServiceNameConfigured(ConfigProperties config, Resource resource) {
```
laurit, line 78: Previous code feels more straightforward. Not sure about the necessity of checking `s.trim().isEmpty()`, isn't checking for not null enough?

robsunday: For me this code is more straightforward because it separates declarative config path from env var config path. In previous implementation, for declarative config, the code was also validating values from ConfigProperties which are not applicable for declarative config.

> Not sure about the necessity of checking s.trim().isEmpty(), isn't checking for not null enough?

I think it may be good to add it because user can define resource this way:
```yaml
resource:
  attributes:
    - name: service.name
      value: ""
```
This will pass parser validation, but will give empty string as `service.name` value

PR #2788, opened 2026-04-23
More opamp identity changes

breedx-splk:
This better aligns with https://github.com/open-telemetry/opamp-spec/pull/312 and https://github.com/signalfx/gdi-specification/pull/387.

The type dispatch stuff was getting completely unwieldy, so I pulled it out of the `OpampActivator` into its own thing.


opamp/src/main/java/com/splunk/opentelemetry/opamp/OpAmpAgentAttributes.java
```diff
@@ -0,0 +1,224 @@
...
+          consumer.putIdentifying(key.getKey(), (double) makeValue(type, value));
+          break;
+        case BOOLEAN:
+          consumer.putIdentifying(key.getKey(), (boolean) makeValue(type, value));
+          break;
+        case STRING_ARRAY:
+          consumer.putIdentifying(key.getKey(), (String[]) makeValue(type, value));
+          break;
+        case LONG_ARRAY:
+          {
```
robsunday: These brackets and the brackets in cases below are not necessary. Please clean them up to improve readability.

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampAgentAttributes.java
```diff
@@ -0,0 +1,224 @@
...
+
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.AttributeType;
+import io.opentelemetry.sdk.resources.Resource;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+class OpampAgentAttributes {
```
robsunday, line 28: File name starts with `OpAmp`, while class name starts with `Opamp`

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampAgentAttributes.java
```diff
@@ -0,0 +1,224 @@
...
+          for (int i = 0; i < typedValueList.size(); i++) {
+            primitiveArray[i] = typedValueList.get(i);
+          }
+          return primitiveArray;
+        }
+    }
+    return null;
+  }
+
+  // Exists for testing
```
robsunday: I'm not a big fan of these interfaces. Yes, it is a clean solution, but it also generated a lot of additional code. For testing purposes we could use mocked OpampClientBuilder class, and get rid of all these interfaces and BuilderAttributeConsumer class (which just delegates everything to  BuilderAttributeConsumer)

robsunday (approved):

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpAmpAgentAttributes.java
breedx-splk: Sure. They're there because it shrinks the scope of the local variables, which will then have naming collisions with the blocks below...but I can rename those to be more specific and remove the brackets.

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampAgentAttributes.java
breedx-splk: Thanks yeah that was a rename casing snafu that I had fixed but didn't make it into git....because of ..... reasons. Fixed now.

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampAgentAttributes.java
breedx-splk: I don't love them either. Lemme see how bad it'll be with the mock....

breedx-splk: yeah that's an improvement. thanks again.

robsunday (approved):

laurit (approved):

PR #2794, opened 2026-04-27
Include experimental metrics in metadata

laurit:


metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -2029,19 +2029,60 @@ private static void handle(
...
+          // sort default configuration first
+          telemetry.sort(Comparator.comparingInt(c -> "default".equals(c.get("when")) ? -1 : 0));
+
+          boolean defaultSeen = false;
+          Set<String> seenMetrics = new HashSet<>();
           for (Map<String, Object> telemetryConfiguration : telemetry) {
-            if (!"default".equals(telemetryConfiguration.get("when"))) {
-              continue;
-            }
+            String configuration = telemetryConfiguration.get("when").toString();
```
robsunday, line 2038: Is it possible that there is no "when" key?

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -2029,19 +2029,60 @@ private static void handle(
...
+                String metricDescription = metric.get("description").toString();
+                if (isDefault) {
+                  seenMetrics.add(metricName);
+                } else if (seenMetrics.contains(metricName)) {
+                  // skip over metrics that we have already added
+                  continue;
+                } else {
+                  seenMetrics.add(metricName);
+                  // metric is not present in default configuration
+                  if (!metricDescription.isEmpty() && !metricDescription.endsWith("\n")) {
```
robsunday: I'd trim it first and then append "\n" if not empty. It is a minor thing, but just in case ...

robsunday (approved):

breedx-splk (approved):

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
laurit: I think it is always present. If it isn't this code will fail an we can fix it then.

PR #2796, opened 2026-04-28
Opamp effective config declarative config

robsunday:
In scope:

- Effective config reported when declarative configuration is in use
- Common format of effective config for both types declarative and envvars configuration


opamp/src/main/java/com/splunk/opentelemetry/opamp/EffectiveConfigFactory.java
```diff
@@ -0,0 +1,55 @@
...
+  String SPLUNK_PROFILER_CALL_STACK_INTERVAL = "SPLUNK_PROFILER_CALL_STACK_INTERVAL";
+  String SPLUNK_SNAPSHOT_PROFILER_ENABLED = "SPLUNK_SNAPSHOT_PROFILER_ENABLED";
+  String SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL = "SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL";
+  String OTEL_EXPORTER_OTLP_TRACES_ENDPOINT = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT";
+  String OTEL_EXPORTER_OTLP_METRICS_ENDPOINT = "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT";
+  String OTEL_EXPORTER_OTLP_LOGS_ENDPOINT = "OTEL_EXPORTER_OTLP_LOGS_ENDPOINT";
+
+  default AgentConfigFile createFile() {
+    ByteString content = new ByteString(buildFileContent().getBytes(UTF_8));
+    return new AgentConfigFile(content, "text/plain+properties");
```
breedx-splk, line 38: I think this is probably fine, but it doesn't seem like the ideal choice to favor one approach (in this case env/props) as the default.

breedx-splk: OH now that I've looked at more of this, it seems like the design is to use `test/plain+properties` for both styles -- sysprops/env AND for declarative config. I'm not sure that's what's expected, and I believe that some folks feel pretty strongly about having a yaml representation on the wire for declarative config. Let's try this for now and see how it goes....but I would expect this to need to be reworked when the format wants to be yaml.

breedx-splk (approved):
I'm fine with trying this approach for now. I'm always astonished how every declarative config PR (both here and in upstream) is like hundreds of lines longer than I'd expect it to be.... 😅

opamp/src/main/java/com/splunk/opentelemetry/opamp/EffectiveConfigFactory.java
robsunday: Yes, this is some initial solution currently agreed that is expected to evolve, most probably into yaml based representation which I'd also prefer.

robsunday:
> I'm fine with trying this approach for now. I'm always astonished how every declarative config PR (both here and in upstream) is like hundreds of lines longer than I'd expect it to be.... 😅

That is true... In this case there is quite a big code duplication, because exporter models have the same structure but are not connected by inheritance, as they are generated at build time from the schema.
Once upstream stabilizes the models and put them under version control, it may be possible to simplify this code a lot.

PR #2799, opened 2026-04-29
Opamp multi endpoints support

robsunday:
In the declarative config it is possible to define multiple exporters for a single signal. These exporters can send data to different endpoints, so it is a need to report multiple URLs in single property.

Reported URLs are now wrapped in double quotes and separated by comma:
`OTEL_EXPORTER_OTLP_TRACES_ENDPOINTS="https://traces.example.com", "http://localhost:4317"`

Endpoint related property names got converted to a plural form to reflect possibility of holding multiple values.


opamp/src/main/java/com/splunk/opentelemetry/opamp/EffectiveConfigFactory.java
```diff
@@ -29,9 +29,9 @@ interface EffectiveConfigFactory {
   String SPLUNK_PROFILER_CALL_STACK_INTERVAL = "SPLUNK_PROFILER_CALL_STACK_INTERVAL";
   String SPLUNK_SNAPSHOT_PROFILER_ENABLED = "SPLUNK_SNAPSHOT_PROFILER_ENABLED";
   String SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL = "SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL";
-  String OTEL_EXPORTER_OTLP_TRACES_ENDPOINT = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT";
-  String OTEL_EXPORTER_OTLP_METRICS_ENDPOINT = "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT";
-  String OTEL_EXPORTER_OTLP_LOGS_ENDPOINT = "OTEL_EXPORTER_OTLP_LOGS_ENDPOINT";
+  String OTEL_EXPORTER_OTLP_TRACES_ENDPOINTS = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINTS";
```
robsunday, line 32: [for reviewer] Property names pluralized to reflect multiple possible values. Is it OK?

opamp/src/test/java/com/splunk/opentelemetry/opamp/EnvVarsEffectiveConfigFileFactoryTest.java
```diff
@@ -49,11 +49,11 @@ void buildFileContent_reportsConfiguredValues() throws IOException {
             "SPLUNK_PROFILER_ENABLED", "true",
             "SPLUNK_PROFILER_MEMORY_ENABLED", "true",
             "SPLUNK_SNAPSHOT_PROFILER_ENABLED", "true",
-            "SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL", "26ms",
-            "SPLUNK_PROFILER_CALL_STACK_INTERVAL", "1235ms",
-            "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", "https://traces.example.com",
-            "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT", "https://metrics.example.com",
-            "OTEL_EXPORTER_OTLP_LOGS_ENDPOINT", "https://logs.example.com"));
+            "SPLUNK_SNAPSHOT_PROFILER_SAMPLING_INTERVAL", "\"26ms\"",
```
robsunday, line 52: [for reviewer] String values are now wrapped in double quotes, as the urls do. This should seamlessly work also if we switch to yaml instead of key=value format for effective config

breedx-splk: Why was the choice made to wrap with quotes? It it being overly cautious about embedded commas or something?

breedx-splk (approved):
I think this is fine for now.

I'm not convinced that the extra string/quote wrapping is completely necessary. I'm also not sure how many users are configuring mutliple endpoints, but I'd be willing to bet that it's very few indeed.

opamp/src/test/java/com/splunk/opentelemetry/opamp/EnvVarsEffectiveConfigFileFactoryTest.java
robsunday: Exactly. Wrapping with quotes allows splitting the list, since comma technically is allowed as a part of URL.

PR #2802, opened 2026-05-04
Dependencies cleanup

robsunday:
This PR started as a dependencies cleanup driven by the fact that `JvmMetricsInstaller` did not work correctly with declarative config, and also directly examined content of `PROFILER_MEMORY_ENABLED` config property instead of calling `ProfilerConfiguration` method.
Unnecessary dependencies were then removed from few modules.

`SplunkConfiguration.PROFILER_ENABLED_PROPERTY` accessed from `ProfilerEnvVarsConfiguration` was moved to this class, where it actually belongs.


custom/src/test/java/com/splunk/opentelemetry/SplunkConfigurationCustomizerTest.java
```diff
@@ -26,7 +26,7 @@
 import java.util.function.Supplier;
 import org.junit.jupiter.api.Test;
 
-class SplunkConfigurationTest {
+class SplunkConfigurationCustomizerTest {
```
robsunday, line 29: [for reviewer] This class was refactored in the past, and its only function is now a customization of the configuration. It triggered the rename.

custom/src/main/java/com/splunk/opentelemetry/RealmAccessTokenChecker.java
```diff
@@ -65,11 +65,12 @@ public int order() {
   private static boolean isRealmConfigured(ConfigProperties config) {
     String realm =
         config.getString(
-            SplunkConfiguration.SPLUNK_REALM_PROPERTY, SplunkConfiguration.SPLUNK_REALM_NONE);
-    return !realm.equals(SplunkConfiguration.SPLUNK_REALM_NONE);
+            SplunkConfigurationCustomizer.SPLUNK_REALM_PROPERTY,
+            SplunkConfigurationCustomizer.SPLUNK_REALM_NONE);
+    return !realm.equals(SplunkConfigurationCustomizer.SPLUNK_REALM_NONE);
```
breedx-splk: Seems like they weren't statically imported before, but it might make things read a little nicer.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilerEnvVarsConfiguration.java
```diff
@@ -94,7 +94,7 @@ private void log(String key, @Nullable Object value) {
 
   @Override
   public boolean isEnabled() {
-    return SplunkConfiguration.isProfilerEnabled(config);
+    return config.getBoolean(CONFIG_KEY_PROFILER_ENABLED, false);
```
breedx-splk, line 97: encapsulation is for suckers anyway. yolo!

breedx-splk (approved):

custom/src/main/java/com/splunk/opentelemetry/SplunkConfigurationCustomizer.java
```diff
@@ -27,12 +27,11 @@
 import java.util.logging.Logger;
 
 @AutoService(AutoConfigurationCustomizerProvider.class)
-public class SplunkConfiguration implements AutoConfigurationCustomizerProvider {
-  private static final Logger logger = Logger.getLogger(SplunkConfiguration.class.getName());
+public class SplunkConfigurationCustomizer implements AutoConfigurationCustomizerProvider {
```
laurit, line 30: if you wish to rename this class then perhaps it would make sense to extract the static constants into an interface named `SplunkConfiguration` so that other classes that use these constants wouldn't need to refer to `SplunkConfigurationCustomizer`

custom/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfigureUtil.java
```diff
@@ -60,7 +60,8 @@ private AutoConfigureUtil() {}
 
   /** Returns the {@link ConfigProperties} used for auto-configuration. */
   public static ConfigProperties getConfig(AutoConfiguredOpenTelemetrySdk sdk) {
-    ConfigProperties config = sdk.getConfig();
+    ConfigProperties config =
+        io.opentelemetry.sdk.autoconfigure.internal.AutoConfigureUtil.getConfig(sdk);
```
laurit, line 64: Initially the purpose of this class was to access package private methods like `getConfig`. The `io.opentelemetry.sdk.autoconfigure.internal.AutoConfigureUtil.getConfig` does the same thing as just calling `sdk.getConfig()`. Imo we should consider moving functionality that does not rely on package access out of this class, perhaps by now this class isn't needed at all.

testing/common/build.gradle.kts
```diff
@@ -3,7 +3,6 @@ plugins {
 }
 
 dependencies {
-  compileOnly(project(":bootstrap"))
   compileOnly(project(":custom"))
```
laurit: `custom` is also a bit unexpected here. I guess it could work for unit testing the agent but probably not for instrumentations.

custom/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfigureUtil.java
robsunday: That's a good observation. It seems indeed that it can be refactored and no longer function as a shim for private API. I'll create another PR for this.

custom/src/main/java/com/splunk/opentelemetry/SplunkConfigurationCustomizer.java
robsunday: I plan further changes to improve distribution config (from declarative config) and it will be a good place to make the change you recommend.

custom/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfigureUtil.java
robsunday: I created issue not to forget about it - https://github.com/signalfx/splunk-otel-java/issues/2803

laurit (approved):

testing/common/build.gradle.kts
robsunday: I did additional refactoring to get rid of dependency on `:custom `project.
`DeclarativeConfigTestUtil` got moved from `:testing.common` to `:custom.testFixtures` because `DeclarativeConfigTestUtil` depends on `io.opentelemetry.sdk.autoconfigure.AutoConfigureUtil.getDistributionConfig` which exists in `:custom`
Few subprojects got updated to make use of `:custom:testFixtures` in their tests.

laurit (approved):

PR #2806, opened 2026-05-07
Declarative config can be used to configure opamp

robsunday:
Declarative config YAML file can now define opamp client configuration under `.opamp/development` node
Structure and location is aligned with current state of [discussion in configuration schema repo](https://github.com/open-telemetry/opentelemetry-configuration/issues/544#issuecomment-4225244736)
Only HTTP is supported at the moment. WS support is coming soon.


opamp/src/test/java/com/splunk/opentelemetry/opamp/OpampClientConfigurationFactoryTest.java
```diff
@@ -0,0 +1,99 @@
...
+    assertThat(configuration.getEndpoint()).isEqualTo("https://opamp.example.com");
+    assertThat(configuration.getPollingInterval()).isEqualTo(3210);
+  }
+
+  @Test
+  void createConfiguration_readsDeclarativeConfig(@TempDir Path tempDir) throws IOException {
+    String yaml =
+        """
+            file_format: "1.0"
+            opamp/development:
```
robsunday, line 67: [question for reviewer] What is your opinion on using in our early development a location that may be the same as the upstream's will?
I see some risk of conflicts with future upstream development, but I used this location mainly for consistency with .net distro, that already committed it to their upstream repo.

laurit: I think it is fine. I know that it has happened before that features that we thought were sill in development suddenly get documented, but here it shouldn't be an issue yet.

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampClientConfiguration.java
```diff
@@ -0,0 +1,76 @@
...
+    return endpoint;
+  }
+
+  public long getPollingInterval() {
+    return pollingInterval;
+  }
+
+  @Override
+  public String toString() {
+    final StringBuffer sb = new StringBuffer("OpampClientConfiguration{");
```
laurit: Usually there is no need to add `final` to local variables, imo in upstream repo there is a checker that disallows it. Probably mean to use `StringBuilder` instead of `StringBuffer`. Here actually it would be better to just write `"OpampClientConfiguration{enabled='" + enabled + ...` the compiler will translate it to  the same code as you were using `StringBuilder` manually.

laurit (approved):

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampClientConfiguration.java
robsunday: Right. IntelliJ generated - I did not look carefully.

PR #2811, opened 2026-05-08
Simplify opamp attribute conversion

laurit:


robsunday (approved):
Nice

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampAgentAttributes.java
```diff
@@ -90,74 +94,60 @@ private Consumer<? super Map.Entry<AttributeKey<?>, Object>> putNonIdentifyingAt
...
+    return value.stream().mapToDouble(v -> (Double) v).toArray();
+  }
+
+  private boolean[] toBooleanArray(List<?> value) {
+    boolean[] result = new boolean[value.size()];
+    for (int i = 0; i < value.size(); i++) {
+      result[i] = (Boolean) value.get(i);
     }
-    return null;
+    return result;
```
breedx-splk, line 151: Why not use the `.stream()` here as well, to be consistent with the methods above?

breedx-splk: Oh, I think because there's no good way to do this with the primitive `boolean` and `mapToBoolean` doesn't exist. 

Java's types keep on giving.

breedx-splk (approved):
Notably cleaner. Thanks!

PR #2816, opened 2026-05-11
Enable security events in CSA by default

breedx-splk:
Secure App team wants these enabled by default.


breedx-splk:
LOL no they don't.

PR #2826, opened 2026-05-19
Use docker cli to determine docker host

laurit:
Since we have had a lot of failures on windows check the docker configuration for the host in case it is different from what we expect. Apparently it can also be `npipe:////./pipe/dockerDesktopLinuxEngine`

### DIFF
--- a/smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
+++ b/smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
@@ -106,8 +106,8 @@ public class WindowsTestContainerManager extends AbstractTestContainerManager {
 
     logger.info("Using docker host {}", uri);
     return DockerClientImpl.getInstance(
-        new DefaultDockerClientConfig.Builder().withDockerHost(NPIPE_URI).build(),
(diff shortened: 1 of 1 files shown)


smoke-tests/src/test/java/com/splunk/opentelemetry/helper/windows/WindowsTestContainerManager.java
```diff
@@ -74,16 +75,41 @@ public class WindowsTestContainerManager extends AbstractTestContainerManager {
...
+            exitCode,
+            text);
+      }
+    } catch (Exception e) {
+      logger.warn("Docker inspect command failed", e);
+    }
+
+    logger.info("Using docker host {}", uri);
+    return DockerClientImpl.getInstance(
+        new DefaultDockerClientConfig.Builder().withDockerHost(NPIPE_URI).build(),
```
robsunday: Is it OK, or it should be:
```java
       new DefaultDockerClientConfig.Builder().withDockerHost(uri).build()
       new ApacheDockerHttpClient.Builder().dockerHost(URI.create(uri)).build());
```
?

laurit: thanks

PR #2832, opened 2026-05-20
Prepare for 2.28.0 release

breedx-splk:
I don't normally like to lump the non-snapshot changeover into this, but I think it's fine this time.

I also acknowledge that the statement about and link to the upstream release notes is not typically part of our convention, but I think this might be a good thing to do, given that pull request 373 in gdi-specification (intentionally not linking to it 🤣 ) exists...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,23 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Unreleased
 
+- OpenTelemetry Java SDK has been updated to version 1.62.0.
+- OpenTelemetry Instrumentation for Java has been updated to version 2.28.0.
+- Changes to experimental OpAMP identity fields [#2788](https://github.com/signalfx/splunk-otel-java/pull/2788)
+- Updates to experimental OpAMP effective config 
+  [#2796](https://github.com/signalfx/splunk-otel-java/pull/2796),
+  [#2799](https://github.com/signalfx/splunk-otel-java/pull/2799),
+  [#2799](https://github.com/signalfx/splunk-otel-java/pull/2806)
+- We now include CSA jars as build artifacts, and publish docker images that include CSA.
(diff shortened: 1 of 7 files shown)


laurit:
> I also acknowledge that the statement about and link to the upstream release notes is not typically part of our convention

imo we should always do the same thing. If we add the link to upstream release notes now then we should do it for all subsequent releases too. Since 373 does not require it I think we should not do it for now. What we could do instead is linking the version numbers in the sdk and upstream agent versions like we do in the github release notes. Note that the link to upstream release notes won't always be super useful since we are going to `2.28.1` of upstream but the interesting things are in `2.28.0` release notes.
373 requests that we copy over relevant changes from the upstream release notes e.g. we'd copy over everything for the agent but exclude entries that are only for library instrumentations. Now to make things more fun we should also clarify whether we'd need to copy over the changes from the sdk too (and perhaps also contrib, if it has anything relevant). Since we bundle sdk through the upstream agent we kind of have more than 1 upstream project.
I'll remove the upstream link for now and also the opamp changes since it is not ready the changes to it aren't relevant to end users. Once we have figured out how we are going to follow 373 we can update the release notes for this release or start following it for some future release. I'll also do the rest of the release steps so that release could be completed tomorrow.

laurit (approved):

breedx-splk:
> imo we should always do the same thing. 

Sure, consistency is helpful, but we _are_ allowed to change.

> Since 373 does not require it I think we should not do it for now. 

Well, the idea was to be able to point at our release and say "we're already linking, we don't need to copy". Like, as a much better solution over what's proposed in 373. We can wait for 373 to figure itself out I guess.

> I'll remove the upstream link for now and also the opamp changes since it is not ready the changes to it aren't relevant to end users. 

We have internal opamp users. But that's also fine.

> I'll also do the rest of the release steps so that release could be completed tomorrow.

Thanks.

PR #2834, opened 2026-05-21
[OpAMP] Agent now returns value of telemetry.distro.version in identifying attrs

robsunday:
According to the documentation https://splunk.atlassian.net/wiki/spaces/PROD/pages/1079695409595/Server+-+Client+communication+RFC#5.2.1.-Required-Attribute-Set, agent's identifying attributes should contain service.version attribute with value copied from `telemetry.distro.version`

### DIFF
--- a/opamp/src/test/java/com/splunk/opentelemetry/opamp/OpampAgentAttributesTest.java
+++ b/opamp/src/test/java/com/splunk/opentelemetry/opamp/OpampAgentAttributesTest.java
@@ -24,6 +24,7 @@ import static io.opentelemetry.semconv.ServiceAttributes.SERVICE_INSTANCE_ID;
 import static io.opentelemetry.semconv.ServiceAttributes.SERVICE_NAME;
 import static io.opentelemetry.semconv.ServiceAttributes.SERVICE_NAMESPACE;
 import static io.opentelemetry.semconv.ServiceAttributes.SERVICE_VERSION;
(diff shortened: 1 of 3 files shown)


robsunday:
I'm not sure why someone decided to use service.version attribute name, instead of for any other name. This generates the collision with real service name from the resource. Possibly it was done to be in sync with collector, but collector could be updated as well to avoid confusion...

opamp/src/test/java/com/splunk/opentelemetry/opamp/OpampAgentAttributesTest.java
```diff
@@ -84,8 +88,9 @@ void addsNonIdentifyingAttributes() {
     OpampAgentAttributes testClass = new OpampAgentAttributes(resource);
 
     testClass.addNonIdentifyingAttributes(builder);
+    // Make sure original service name is reported in nonidentifying attributes
```
robsunday, line 91: [for reviewer] I think it is worth keeping the original service name in non identifying attributes.

breedx-splk (commented):
We should be trying to follow the guidance in the upstream otel spec, specifically here: https://github.com/open-telemetry/opamp-spec/blob/cc23ca3d209c640f5adec59329e7db7fe48388d6/opentelemetry-guidelines.md#identifying-attributes

`telemetry.distro.version` is not called out there, so I do not think we should include it.

Please see if the document author is open to editing their document to match upstream otel.

PR #2852, opened 2026-06-02
Fix NPE in snapshot profiler when using virtual threads

laurit:


robsunday (approved):
Looks fine

breedx-splk (approved):
Thanks, you rock!

PR #2855, opened 2026-06-03
Report instrumentation version for splunk jvm metrics

laurit:


instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelMeterProvider.java
```diff
@@ -19,12 +19,24 @@
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.metrics.MeterBuilder;
+import io.opentelemetry.instrumentation.api.internal.EmbeddedInstrumentationProperties;
 
 class OtelMeterProvider {
-  private static final String INSTRUMENTATION_NAME = "com.splunk.javaagent.jvm-metrics-splunk";
+  private static final String INSTRUMENTATION_NAME = "com.splunk.jvm-metrics";
```
laurit, line 26: This could be considered a breaking change but since as far as I remember apm doesn't show instrumentation scope it should be fine. @robsunday @breedx-splk if you guys think that we should keep the old instrumentation name and perhaps only change for 3.0 then that isn't hard too. Mostly changed it to align with the embedded version file generator.

robsunday (approved):
I think it is fine if APM doesn't care. I also did not find any public information about this instrumentation name.

PR #2860, opened 2026-06-04
Convert workflows to use setup-gradle action

laurit:


laurit:
Hopefully better caching will improve build stability. Though could backfire as with caching we might not always download the latest upstream snapshot.

breedx-splk (approved):
Seems worth it to me.

Why remove the flags from the muzzle check? Just because it's rarely a problem and mostly unnecessary?

laurit:
> Why remove the flags from the muzzle check? Just because it's rarely a problem and mostly unnecessary?

The `--scan` flag shouldn't be needed, the build scan is created even without it. Not sure why the `--no-daemon` flag was added so removed it. Added back `--stacktrace` to muzzle tasks where it was removed (build scan should have the stacktrace). Upstream does not have `--stacktrace`, don't remember missing it.

PR #2853, opened 2026-06-02
[OpAMP] Effective declarative config update to match GDI Spec

robsunday:
This is update of Effective Declarative Config to follow [GDI spec](https://github.com/signalfx/gdi-specification/blob/main/specification/opamp_datamodel.md#effective-declarative-config)

**This PR includes:**
- valid content type for effective config yaml
- valid file name for EDC according to GDI Spec
- endpoints for signals are reported
- reporting declarative config file name
- reporting profiler configuration
- reporting snapshot profiler configuration

**Plans for next PRs:**
- update structure of opamp/development node in declarative config file.

PR #2863, opened 2026-06-05
[OpAMP] Effective declarative config models use for YAML serialization

robsunday:
When I worked on https://github.com/signalfx/splunk-otel-java/pull/2853 I was not 100% happy with use of simple Maps and Lists together with `YamlNodeBuilder` class.

So in this PR I took another approach - use of annotated model classes instead of `YamlNodeBuilder` from #2853 PR. While it generates more classes I think it will be more maintainable in a long run, when we will add more effective config elements.

What is your opinion?


robsunday:
This PR also includes some package structure updates

PR #2871, opened 2026-06-08
Extract profling supervisor from jfr activator

breedx-splk:
In order to better handle the upcoming changes where we want to be able to turn the profiler on or off at runtime, we need some refactoring around how JFR and the profiler components get spun up.

This PR is a step in that direction. It introduces a new class called the `ProfilingSupervisor` which can read commands from a queue in order to start and stop the profiler. Only start is implemented currently. When the `JfrAgentListener` is triggered, it will ALWAYS create an instance of the supervisor.

This PR also factors out the creational responsibilities from the JFRActivator AND the new supervisor, so that it's well encapsulated in the `RecordingSequencerBuilder` now. I leveraged codex to assist with some of the tests.

Future work:

* Figure out how other components (opamp machinery) will be able to send messages to the supervisor.
* Implement status check for reporting effective config
* Implement stop


profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencerBuilder.java
```diff
@@ -199,29 +114,14 @@ private void activateJfrAndRunForever(ProfilerConfiguration config, Resource res
...
-      DeclarativeConfigProperties exporterConfig =
-          ((DeclarativeConfigProperties) configProperties).getStructured("exporter", empty());
-      return LogExporterBuilder.fromConfig(exporterConfig);
-    }
-    if (configProperties instanceof ConfigProperties) {
-      return LogExporterBuilder.fromConfig((ConfigProperties) configProperties);
-    }
-    throw new IllegalArgumentException(
-        "Unsupported config properties type: " + configProperties.getClass().getName());
+    return RecordingSequencer.builder()
```
robsunday: To me it is somehow confusing piece of code.  There is a call to `RecordingSequencer.builder()` method from `RecordingSequencerBuilder.build()`
Maybe `RecordingSequencerBuilder` should be refactored to `RecordingSequencerFactory`, or `RecordingSequencer.Builder` class can be removed?

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
```diff
@@ -0,0 +1,153 @@
...
+    this.config = config;
+    this.jfr = jfr;
+    this.sdk = sdk;
+    this.commandQueue = commandQueue;
+  }
+
+  static ProfilingSupervisor createAndStart(
+      AutoConfiguredOpenTelemetrySdk sdk, ProfilerConfiguration config) {
+    ExecutorService executor = HelpfulExecutors.newSingleThreadExecutor("JFR Profiler");
+    // TODO: What if already started?
```
robsunday, line 61: Maybe `executor` should be a static field of `ProfilingSupervisor` class, or could be passed as parameter?

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrAgentListenerTest.java
```diff
@@ -0,0 +1,168 @@
...
+    AutoConfiguredOpenTelemetrySdk sdk = createAutoConfiguredSdk(yaml, tempDir, autoCleanup);
+
+    var jfr = mock(JFR.class);
+    var supervisor = mock(ProfilingSupervisor.class);
+
+    when(jfr.isAvailable()).thenReturn(true);
+
+    JfrAgentListener listener =
+        new JfrAgentListener(jfr) {
+          @Override
```
robsunday, line 70: Something to consider - instead of overriding a method you could add a `Function` parameter to the `JfrAgentListener` constructor that exists for testing purposes.
This function would be a factory for supervisor executed from `makeProfilingSupervisor()`

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
```diff
@@ -0,0 +1,153 @@
...
+      return;
+    }
+    if (!jfr.isAvailable()) {
+      logger.warning(
+          "JDK Flight Recorder (JFR) is not available in this JVM. Profiling will not start.");
+      return;
+    }
+    config.log();
+    logger.info("Profiler is active.");
+    setupContextStorage();
```
robsunday, line 124: Looks like multiple calls to `tryStart()` will add multiple context wrappers.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
```diff
@@ -0,0 +1,153 @@
...
+
+  @VisibleForTesting
+  void start(ExecutorService executor) {
+    executor.submit(this::forever);
+  }
+
+  private void forever() {
+    while (true) {
+      try {
+        ProfilingCommand command = commandQueue.take();
```
robsunday, line 76: Optional optimization - we are actually interested in the last command in the queue, so you could discard all earlier commands if exist - just in case of some hypothetical scenario when there were some significant delays in switching profiler on and off, or there is more than one sender to this queue.

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencerBuilder.java
```diff
@@ -47,97 +37,22 @@
 import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
-
-@AutoService(AgentListener.class)
-public class JfrActivator implements AgentListener {
 
+class RecordingSequencerBuilder {
```
laurit, line 41: Imo `RecordingSequencerBuilder` isn't a good name since `RecordingSequencer` isn't a good name because it doesn't really sequence anything.

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrAgentListener.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import io.opentelemetry.javaagent.extension.AgentListener;
+import io.opentelemetry.sdk.autoconfigure.AutoConfigureUtil;
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+
+@AutoService(AgentListener.class)
+public class JfrAgentListener implements AgentListener {
+
+  private static final java.util.logging.Logger logger =
```
laurit: could import Logger

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
laurit, line 124: this won't work, `ContextStorage.addWrapper` can only be called before context storage is initialized

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencerBuilder.java
breedx-splk: Agreed, this is stupid. I did not intend to have two builders. I'll fix this.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
breedx-splk: I think we ultimately want to process all of the commands in the order they arrive. It's possible (but unlikely) that it gets toggled on and then back off back to back. We don't want to lose that. If you're saying that since the current command set is small, we could "resolve" them to the last...but that sounds overly complicated. I also expect the command set to grow to change other aspects of it, so let's not do that now. 

We can always optimize later if/when that becomes a practical problem.

profiler/src/main/java/com/splunk/opentelemetry/profiler/RecordingSequencerBuilder.java
breedx-splk: Yeah, I think I agree. This is probably a historical artifact -- I think it used to do more? Right now, it's entire job is to basically ensure that the recorder is started and to periodically flush it. How about `PeriodicRecordingFlusher`?

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrAgentListenerTest.java
breedx-splk: Good idea! I considered it, and even tried it before landing on this. I didn't think it was cleaner. If you feel strongly about it, I could refactor.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
breedx-splk: Interesting! Seems like it was already "broken" before this PR then...how was it working?

What should we do about that then? Do this via `BeforeAgentListener` or how might we get this in place correctly?

breedx-splk: Yeah @laurit also called this out for other reasons. We need to address it.

laurit: The current code works before apparently nothing touches the context before the wrapper is set. It might be a bit lucky considering that other `AgentListener` implementations may trigger initializing the context before `JfrActivator` manages to set things up. Currently context initialization is triggered by `JfrActivator` itself right after it has set up the context wrapper (calling `executor.submit` accesses context). 

> What should we do about that then? Do this via BeforeAgentListener or how might we get this in place correctly?

You must call it from the main thread whenever there is a possibility that the profiler might get enabled. We may also need to modify it so it wouldn't do anything unless the profiler is really enabled.

breedx-splk: >You must call it from the main thread

It looks like `BeforeAgentListener` is ultimately called from `premain()` so that should be ok...unless you still think that's not early enough. I hear your point about other listeners being invoked first and potentially screwing this up. Any ideas on making this robust?

>whenever there is a possibility that the profiler might get enabled.

Well, unfortunately by making it dynamic, that means that it might be enabled at any time....so I guess we need to always do this. Another downside of making the agent more stateful.... 😭

laurit: Imo doing it in the `AgentListener`, is fine since we haven't had issues with it. What you should not do is moving it to a background thread, even if it would work sometimes (I think it currently doesn't work at all) it could still race with something.

> Well, unfortunately by making it dynamic, that means that it might be enabled at any time....so I guess we need to always do this. Another downside of making the agent more stateful.... 

I guess the configuration can't be changed when opamp isn't enabled.

profiler/src/test/java/com/splunk/opentelemetry/profiler/JfrAgentListenerTest.java
robsunday: I'd prefer function, but I can live with current implementation :-).

breedx-splk:
Closing in favor of 2881.

PR #2881, opened 2026-06-12
[OpAMP] Enable/disable profiler via remote config

robsunday:
This PR is based on @breedx-splk PR https://github.com/signalfx/splunk-otel-java/pull/2871
Original PR has an issue with `JfrContextStorage` that might not fully work with dynamically enabled/disabled profiling.
This PR implements the following approach to support dynamic profiling:

- `JfrContextStorage` is always added to `ContextWrappers` regardless of profiling settings. It happens in static initializer of `ProfilingSupervisor` class
- `JfrContextStorage` has a method that enables/disables processing in `JfrContextStorage.attach` method.

Profiling can now be enabled/disabled via remote config according to [GDI spec](https://github.com/signalfx/gdi-specification/blob/main/specification/opamp_datamodel.md#data-format)

Sampling interval is not supported yet and Effective config is not updated and sent after getting remote config. It will be implemented in subsequent PRs


breedx-splk (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/JfrAgentListener.java
```diff
@@ -0,0 +1,83 @@
...
+  @VisibleForTesting
+  JfrAgentListener(JFR jfr) {
+    this.jfr = jfr;
+  }
+
+  @Override
+  public void afterAgent(AutoConfiguredOpenTelemetrySdk sdk) {
+
+    ProfilerConfiguration config = getProfilerConfiguration(sdk);
+    // Always start the supervisor, because we may need to start it later elsewhere.
```
laurit: Imo there is something wrong with this sentence, should `it` be replaced with `profiler` or `profiling`?

profiler/src/main/java/com/splunk/opentelemetry/profiler/PeriodicRecordingFlusher.java
```diff
@@ -20,29 +20,41 @@
...
-    executor.scheduleAtFixedRate(
-        this::handleInterval, 0, recordingDuration.toMillis(), TimeUnit.MILLISECONDS);
+    scheduledFlushFuture =
+        executor.scheduleAtFixedRate(
+            this::handleInterval, 0, recordingDuration.toMillis(), TimeUnit.MILLISECONDS);
+  }
+
+  public void stop() {
+    if (scheduledFlushFuture != null) {
+      scheduledFlushFuture.cancel(true);
```
laurit: I'm not sure about interrupting the thread. The problem with it is that it is unpredictable where exactly it gets interrupted. It will probably log the interrupted exception at error level. Might be easier to just let the task to run to completion when it is already running. Could also consider somehow waiting for the periodic task to complete (idk whether that is really necessary).

profiler/src/main/java/com/splunk/opentelemetry/profiler/PeriodicRecordingFlusher.java
```diff
@@ -20,29 +20,41 @@
...
 import java.util.logging.Logger;
 
 /** Responsible for periodically generating a sequence of JFR recordings. */
-class RecordingSequencer {
-  private static final Logger logger = Logger.getLogger(RecordingSequencer.class.getName());
+class PeriodicRecordingFlusher {
+  private static final Logger logger = Logger.getLogger(PeriodicRecordingFlusher.class.getName());
 
   private final ScheduledExecutorService executor =
       HelpfulExecutors.newSingleThreadedScheduledExecutor("JFR Recording Sequencer");
```
laurit: maybe also rename the thread? This thread name could also be used for filtering our threads from profiling ourput.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * This class oversees the profiling subsystem. It runs for the entire time that the agent is
+ * running.
+ */
+public class ProfilingSupervisor {
+  static {
+    setupContextStorage();
```
laurit: Imo this is hard to follow. We know that this has to be called before otel context is first used (maybe add a comment?). Tracking down where the `ProfilingSupervisor` class is initialized feels harder than having an explicit method call.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+  }
+
+  /**
+   * Try and start the profiler. This does not check configuration, just responds to a command
+   * request.
+   */
+  private void tryStart() {
+    if (isJfrRecordingActive()) {
+      logger.warning("JFR is already running, not starting again.");
```
laurit, line 132: Not sure that these warnings make sense. To me it looks that these are caused by our coding errors not by something user did or has control of.

profiler/src/main/java/com/splunk/opentelemetry/profiler/PeriodicRecordingFlusher.java
robsunday: I had some doubts here as well. You have a point that we do not know where the thread is interrupted, but this thread just flushes data to some temporary files and mostly waits. All IO operations in `JfrRecorder.flushSnapshot()` are inside try-with-resources so everything should be closed, so I assumed it is safe for interrupt the thread.
Also, `InterruptedException` will trigger log `"ProfilingSupervisor is shutting down"` printed from `ProfilingSupervisor.commandLoop()` method.

To be on the safe side, I'll switch to `false`, test and see if some additional development is needed.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
robsunday: I'd keep it. No harm caused, just a signal that something may be not optimal

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
robsunday: So I added explicit call in `JfrContextStorageInitializer` class that is a config customizer, so it is executed before SDK is set up. Do you see some better place for calling `ProfilingSupervisor.setupJfrContextStorage()`?

robsunday: I moved `ProfilingSupervisor.setupJfrContextStorage()` call to `JfrAGentListener` and it still works, so I removed `JfrContextStorageInitializer`. I can easily revert it if needed.

PR #2898, opened 2026-06-23
[OpAMP] Remote config -> effective config feedback

robsunday:


profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilerConfiguration.java
```diff
@@ -16,9 +16,13 @@
 
 package com.splunk.opentelemetry.profiler;
 
+import com.splunk.opentelemetry.profiler.util.OptionalConfigurableSupplier;
 import java.time.Duration;
 
 public interface ProfilerConfiguration {
+  OptionalConfigurableSupplier<ProfilerConfiguration> SUPPLIER =
```
robsunday, line 23: [for reviewer] Now both implementations of `ProfilerConfiguration` interface are accessible through the same `SUPPLIER`. Depending on configuration flavor - `SUPPLIER` will serve appropriate type of configuration. Before this refactoring `ProfilerEnvVarsConfiguration` was created in multiple places. Now it is instantiated only once, which is OK since `ConfigProperties` is immutable after SDK is initialized. It was necessary to support profiler remote config in a uniform way.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilerConfigurationInitializer.java
```diff
@@ -23,27 +23,53 @@
...
  * created, such as AgentListeners.
  */
-@AutoService(DeclarativeConfigurationCustomizerProvider.class)
-public class ProfilerDeclarativeConfigurationInitializer
-    implements DeclarativeConfigurationCustomizerProvider {
+@AutoService({
+  DeclarativeConfigurationCustomizerProvider.class,
+  AutoConfigurationCustomizerProvider.class
+})
+public class ProfilerConfigurationInitializer
```
robsunday, line 39: [for reviewer] This class supports environment and declarative configuration. Appropriate method is called to init `SUPPLIER` with right profiler configuration kind, depending on agent configuration flavor used.

opamp/src/main/java/com/splunk/opentelemetry/opamp/effectiveconfig/EffectiveConfigReporter.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import io.opentelemetry.sdk.autoconfigure.AutoConfigureUtil;
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+import java.util.HashMap;
+import java.util.Map;
+import okio.ByteString;
+import opamp.proto.AgentConfigFile;
+import opamp.proto.AgentConfigMap;
+
+public class EffectiveConfigReporter {
```
robsunday, line 31: [for reviewer] This class keeps recently reported effective config and reports a new effective config only if the content changed.
Maybe it is not the most effective way of handling effective config reporting but it is a good start. I'm open to any improvement ideas.

opamp/src/main/java/com/splunk/opentelemetry/opamp/ProfilerRemoteConfiguration.java
robsunday, line 1: [for reviewer] This class is a proxy over original configuration. Original profiler configuration is immutable once created. This class overwrites handling of the properties that are updatable via remote configuration, while preserving original values.
I'm still considering introducing a mutable ProfilerConfiguration interface implementation and letting appropriate factories to set it up. This would be served via SUPPLIER without any proxy objects.

PR #2905, opened 2026-06-25
[OpAMP] Remote config profiler call stack interval support

robsunday:
This PR includes:

- call stack sampling interval support for Remote Configuration
- additional tests and supporting refactorings
- JFR Recording Sequencer -> JFR Recording Flusher updates missed so far


profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
```diff
@@ -153,13 +172,25 @@ private void tryStop() {
     logger.info("Profiler is deactivated.");
   }
 
+  private void tryReinitialize() {
```
robsunday, line 175: [for reviewer] This was added for convenience and clarity. Its intent is to apply config that is served by the supplier.
There are two steps needed to perform it:
1) Stop the profiler if it is currently running
2) If current config enables the profiler then it is started with configuration available in the supplier.

This is enough to apply remote config.

profiler/src/main/java/com/splunk/opentelemetry/profiler/PeriodicRecordingFlusherFactory.java
```diff
@@ -37,25 +37,11 @@
 import java.time.Duration;
 import java.util.Map;
 
-class PeriodicRecordingFlusherBuilder {
+class PeriodicRecordingFlusherFactory {
```
robsunday, line 41: [for reviewer] Builder was converted into factory that is passed as an argument into [ProfilingSupervisor](https://github.com/signalfx/splunk-otel-java/pull/2905/changes#diff-9b1635d3a2e3dd451af57624ce4274f6ec4b489fde0fc0b392055baa4ff6cfa4R70) constructor used by tests. It simplifies testing of `ProfilingSupervisor`.

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilerConfiguration.java
```diff
@@ -170,6 +170,54 @@ public Object getConfigProperties() {
     return configProperties;
   }
 
+  @Override
+  public boolean equals(Object other) {
```
robsunday, line 174: [for reviewer] Used to detect if configuration changed in [RemoteConfigProcessor](https://github.com/signalfx/splunk-otel-java/pull/2905/changes#diff-f622df91bb8a7aa0500d8f1f6b86823f32768c4d2e061ec10dc4fde799403a76R90) and trigger profiler restart only if necessary

laurit (approved):

opamp/src/main/java/com/splunk/opentelemetry/opamp/RemoteConfigProcessor.java
```diff
@@ -100,12 +101,12 @@ public void applyConfig(AgentRemoteConfig remoteConfig, OpampClient opampClient)
           opampClient);
 
     } catch (Exception e) {
+      logger.log(Level.WARNING, "Remote config not applied due to exception", e);
```
laurit: for user visible error might be better to use full sentences that end with `.`. Replace `config` with `configuration`

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
```diff
@@ -145,21 +164,29 @@ private void tryStart() {
...
     deactivateJfrRecording();
     logger.info("Profiler is deactivated.");
   }
 
+  private void tryReinitialize() {
+    tryStop();
+    // Start the profiler with current settings if it is enabled. New settings will be applied.
+    if (configSupplier.get().isEnabled()) {
+      tryStart();
+    }
```
breedx-splk, line 180: Do we care about this being non-atomic? That is, if this is called concurrently by two threads, can the order of starting and stopping get confusing?

breedx-splk: But I guess if the supervisor can only do one command at a time from the queue it's probably fine.

robsunday: Exactly. This was my assumption

breedx-splk (approved):
Seems good to me. 🎈

PR #2919, opened 2026-07-01
[OpAmp] Memory profiling support for remote config

robsunday:
PR includes:

- Memory profiler can be enabled/disabled by remote config
- JVM memory related metrics are enabled/disabled according to memory profiler settings
- Redundant classes cleanup


custom/src/main/java/com/splunk/opentelemetry/logging/SplunkSlf4jLoggingCustomizer.java
```diff
@@ -59,7 +59,7 @@ private static void addCustomLoggingConfiguration() {
 
       // Silence the following warning on jdk21, this warning should go away with an update to jfr
       // parser
-      // [otel.javaagent 2023-10-09 14:38:17:665 +0000] [JFR Recording Sequencer] WARN
+      // [otel.javaagent 2023-10-09 14:38:17:665 +0000] [JFR Recording Flusher] WARN
```
laurit, line 62: looks like this PR is built on top of https://github.com/signalfx/splunk-otel-java/pull/2905

robsunday: It was. After merge this diff is gone.

agent/build.gradle.kts
```diff
@@ -38,6 +38,10 @@ val licenseReportDependencies = configurations.create("licenseReportDependencies
 }
 
 val otelInstrumentationVersion: String = rootProject.extra["otelInstrumentationVersion"] as String
+val javaagentInstrumentationProjects =
+  project(":instrumentation").subprojects.filter {
+    it.name !in setOf("compile-stub", "nocode-testing")
```
laurit: what was wrong with using `splunk.instrumentation-conventions` as the marker?

robsunday: There were two reasons:
1) After profiler included :instrumentation:jmv-metrics there were errors from gradle: "Cannot mutate the dependencies of configuration ':agent:javaagentLibs' after the configuration was resolved"
2) This marker caused including ":instrumentation:servlet-3-testing" module into production code.

I finally decided to explicitly list all relevant ":instrumentation" submodules.

agent/build.gradle.kts
robsunday, line 1: The way of handling `:instrumentation` submodules was changed because after `:profiler` included `:instrumentation:jmv-metrics` there were errors from gradle:

> "Cannot mutate the dependencies of configuration ':agent:javaagentLibs' after the configuration was resolved"


The `splunk.instrumentation-conventions` marker also caused including `:instrumentation:servlet-3-testing` module into production code.

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelAllocatedMemoryMetrics.java
```diff
@@ -23,25 +23,42 @@
...
-        .setUnit("bytes")
-        .setDescription("Approximate sum of heap allocations.")
-        .buildWithCallback(
-            measurement ->
-                measurement.record(
-                    allocatedMemoryMetrics.getCumulativeAllocationTotal(), attributes));
+    allocatedMemoryCounter =
+        meter
+            .counterBuilder(METRIC_NAME)
+            .setUnit("bytes")
```
laurit: Unit should be `"By"`

robsunday: Right. Good catch. It was in existing code but it is good thing to fix.

agent/build.gradle.kts
laurit: I wasn't able to reproduce the `Cannot mutate the dependencies of configuration` failure.

> The splunk.instrumentation-conventions marker also caused including :instrumentation:servlet-3-testing module into production code.

This can be avoided by renaming `splunk.instrumentation-conventions` to `splunk.instrumentation-testing-conventions` and creating `splunk.instrumentation-conventions` that adds the `splunk.instrumentation-testing-conventions` plugin. Then the testing module should still function and not get included. Upstream uses a similar split.

robsunday: Updated to your recommendation

buildSrc/src/main/kotlin/splunk.instrumentation-conventions.gradle.kts
```diff
@@ -34,249 +34,5 @@
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
```
laurit: this import isn't probably needed any more

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelAllocatedMemoryMetrics.java
```diff
@@ -23,25 +23,42 @@
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.metrics.ObservableLongCounter;
 
 public class OtelAllocatedMemoryMetrics {
```
laurit, line 28: maybe these classes should be moved to the profiling module

laurit (approved):

instrumentation/jvm-metrics/src/main/java/com/splunk/opentelemetry/instrumentation/jvmmetrics/otel/OtelAllocatedMemoryMetrics.java
robsunday: Makes sense to me, as they are now used only in tests and by `ProfilingSupervisor.`
There is also `io.opentelemetry.api.metrics.JvmMemory` class that is not used at all, so I guess it can be removed.
What do you think about it @breedx-splk ?

robsunday: I tried to do it but it generated lots of additional changes. I'd keep it for now as it is and make it as a separate task.

PR #2920, opened 2026-07-01
Profiler configuration validation

robsunday:
**Based on PR https://github.com/signalfx/splunk-otel-java/pull/2905.
Real changes are only in setters of [ProfilerConfiguration](https://github.com/signalfx/splunk-otel-java/pull/2920/changes#diff-c9550fecdd6e0ee388cb32a7c0a539caec54a87b892bbf3b3985c3bb88460830R309) class**

`Duration` type properties and call stack depth are now validated.
This implementation validates values in setters of `ProfilerConfiguration` class.

Validation throws `IllegalArgumentException` in case of failure.
It is triggered for:

- Environment configuration -> agent is not started
- Declarative configuration -> agent is not started
- Remote configuration -> remote config is not applied

For remote config OpAmp server gets the following remote config status response:
```json
  "status": {
    "lastRemoteConfigHash": {
      "base64": "9hkJef8xZ/Tw+fT1hMm7GNDP/yNt74+ujxxITVbfUpk=",
      "hex": "f6190979ff3167f4f0f9f4f584c9bb18d0cfff236def8fae8f1c484d56df5299",
      "size": 32
    },
    "status": "FAILED",
    "errorMessage": "Exception occurred: Invalid call stack interval: -10. Must be >= 0."
  }
```


laurit (approved):
Initially I thought that maybe we shouldn't prevent the agent from starting when the value is invalid but the sdk behaves the same way. Adding `-Dotel.exporter.otlp.protocol=x` will also prevent the agent from starting so this should be fine.

PR #2921, opened 2026-07-01
Rename csa to secureapp

breedx-splk:
There is a request to change the artifact name to not use `csa` in favor of `secureapp`.

This change also updates the docker image name.

These are both breaking changes for any users that have automation that leverages the old names.


laurit (approved):

robsunday (approved):
All is fine

breedx-splk:
Was told to hold off on this, so I'm just gonna close it for now. We can revisit if the decision is changed.

PR #2934, opened 2026-07-08
Add ability for the opamp client to use the x-sf-token header

breedx-splk:
If the opamp client is to be used without a collector, then it will need the ability to set the token header.

### DIFF
--- a/opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampClientConfiguration.java
+++ b/opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampClientConfiguration.java
@@ -19,6 +19,7 @@ package com.splunk.opentelemetry.opamp;
 public class OpampClientConfiguration {
   private boolean enabled;
   private String endpoint;
+  private String accessToken = null;
   private long pollingInterval;
 
   private OpampClientConfiguration() {}
@@ -39,6 +40,10 @@ public class OpampClientConfiguration {
     return pollingInterval;
   }
 
+  public String getAccessToken() {
+    return accessToken;
+  }
+
   @Override
   public String toString() {
     return "OpampClientConfiguration{"
@@ -47,6 +52,9 @@ public class OpampClientConfiguration {
         + '\''
         + ", endpoint='"
         + endpoint
+        + '\''
+        + ", accessToken='"
+        + (accessToken == null ? "<null>" : "***redacted***")
         + '\''
         + ", pollingInterval="
         + pollingInterval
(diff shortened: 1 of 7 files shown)


opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampClientConfiguration.java
```diff
@@ -19,6 +19,7 @@
 public class OpampClientConfiguration {
   private boolean enabled;
   private String endpoint;
+  private String accessToken = null;
```
robsunday: Initialization is redundant and looks inconsistent with other properties.

robsunday:
Are you going to add `OpampClientConfigurationFactory` and `OpampClientConfigurationFactoryTest` changes in this PR, or another one will be created?

robsunday (approved):

breedx-splk:
> Are you going to add `OpampClientConfigurationFactory` and `OpampClientConfigurationFactoryTest` changes in this PR, or another one will be created?

Why, does `OpampClientConfigurationFactory` require something extra/special? I thought this would already cover those.

robsunday:
`OpampClientConfiguration.Builder.withAccessToken()` is called in tests to setup confiuration so tests pass, but it is not called from production code when starting the agent.

`OpampClientConfiguration` is created in two methods:

- `OpampClientConfigurationFactory.createConfigurationFromDeclarativeConfig`
- `OpampClientConfigurationFactory.createConfiguration(ConfigProperties)` - used for environment config

`OpampClientConfiguration.Builder.withAccessToken()` should be called there as well.

opamp/src/main/java/com/splunk/opentelemetry/opamp/OpampClientConfigurationFactory.java
```diff
@@ -59,6 +61,7 @@ private static OpampClientConfiguration createConfiguration(ConfigProperties con
     return OpampClientConfiguration.builder()
         .withEnabled(config.getBoolean("splunk.opamp.enabled", false))
         .withEndpoint(config.getString("splunk.opamp.endpoint"))
+        .withAccessToken(config.getString(SPLUNK_ACCESS_TOKEN))
```
robsunday, line 64: We should either use constants in all places here, or just a plain strings. This constant is not used anywhere else, so to me it is not worth making this class bigger by adding a constant.

breedx-splk: That constant is used in a few other places. I prefer using existing constants over duplicating string literals, which is prone to creating inconsistencies. I didn't create constants for the OTHER string key literals because they are only used here.

PR #2951, opened 2026-07-17
Support the new upstream metadata format

breedx-splk:
This gets us past the failing metadata generator. I haven't been following upstream closely enough to know if this is a great approach, I just let codex take the reins mostly. Seems to be doing ok.

### DIFF
--- a/metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
+++ b/metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
@@ -1349,27 +1349,45 @@ public class MetadataGenerator {
     return parseInstrumentations(new URL(url));
   }
 
-  private static List<Map<String, Object>> parseInstrumentations(URL url) throws IOException {
+  static List<Map<String, Object>> parseInstrumentations(URL url) throws IOException {
     Yaml yaml = new Yaml();
     Map<String, Object> metadata;
     try (InputStream inputStream = url.openStream()) {
       metadata = yaml.load(inputStream);
     }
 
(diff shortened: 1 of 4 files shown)


metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -1349,27 +1349,45 @@ private static List<Map<String, Object>> parseInstrumentations(
...
+
     List<Map<String, Object>> result = new ArrayList<>();
 
-    handle(result, (List<Map<String, Object>>) metadata.get("libraries"));
-    handle(result, (List<Map<String, Object>>) metadata.get("custom"));
+    handle(
+        result,
+        (List<Map<String, Object>>) metadata.get("libraries"),
+        configurationDefinitions,
+        metricDefinitions);
```
robsunday: I think `metricDefinitions` is redundant parameter. It can be extracted inside `handle` method.

metadata-generator/src/main/java/com/splunk/opentelemetry/tools/MetadataGenerator.java
```diff
@@ -1349,27 +1349,45 @@ private static List<Map<String, Object>> parseInstrumentations(
     return parseInstrumentations(new URL(url));
   }
 
-  private static List<Map<String, Object>> parseInstrumentations(URL url) throws IOException {
+  static List<Map<String, Object>> parseInstrumentations(URL url) throws IOException {
```
robsunday, line 1354: `@VisibleForTests` would be good here

robsunday (approved):

PR #2952, opened 2026-07-17
Export thread lock and monitor info as profiling labels

breedx-splk:
This borrows some of the art from the opamp thread dump command and just ands the thread lock/monitor info as a standard part of the profiling signal. These labels aren't yet understood by the backend and are expected to be quietly dropped until then.

### DIFF
--- a/profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/StackTraceBuilder.java
+++ b/profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/StackTraceBuilder.java
@@ -16,6 +16,12 @@
 
 package com.splunk.opentelemetry.profiler.snapshot;
 
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.lang.management.LockInfo;
+import java.lang.management.MonitorInfo;
+import java.lang.management.ThreadInfo;
 import java.time.Duration;
 import java.time.Instant;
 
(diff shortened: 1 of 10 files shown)


profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/StackTraceBuilder.java
```diff
@@ -71,14 +83,20 @@ public StackTraceBuilder withSpanId(String spanId) {
...
-        spanContext.getSpanId(),
-        0);
+        timestamp, duration, threadInfo(), spanContext.getTraceId(), spanContext.getSpanId(), 0);
+  }
+
+  private ThreadInfo threadInfo() {
+    if (threadInfo != null) {
+      return threadInfo;
+    }
+    ThreadInfo result = mock(ThreadInfo.class);
```
robsunday, line 93: Not a big issue, but maybe do lazy initialization of threadInfo, or create static constant instead of creating new mock every time.

robsunday (approved):
Looks good with one small test related comment

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/StackTraceBuilder.java
breedx-splk: The `ThreadInfo` instance is created from parts of the builder at the time `build()` is called....so I'm not sure how a static would work. The builder can have different fields each time.

I'm not scared of creating a mock for each test. It also gives me some additional small assurance about test isolation/separation.

PR #2955, opened 2026-07-21
[OpAMP] Enable/disable snapshot profiling with remote config

robsunday:
Adds support for dynamic enable/disable snapshot profiling using remote configuration.


profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
```diff
@@ -64,14 +64,12 @@ public class ProfilingSupervisor {
       OptionalConfigurableSupplier<ProfilerConfiguration> configSupplier,
       JFR jfr,
       AutoConfiguredOpenTelemetrySdk sdk,
-      BlockingQueue<ProfilingCommand> commandQueue,
```
robsunday, line 67: [for reviewer] It was not used anywhere, so I removed this parameter from constructor.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSupervisor.java
robsunday, line 1: [for reviewer] This class follows the pattern of using async processing established in `ProfilingSupervisor` class.

laurit (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/ProfilingSupervisor.java
breedx-splk: Good catch. I think it might have been used for testing at some point, or the original idea was that other components might need to have the queue to send stuff into it....but it's encapsulated and I guess it's not needed. We can always add it back if/when that changes.

breedx-splk (approved):
Looks good, thanks. I like that it's following the pattern established by the other profiler, but I don't love that they now have two separate threads across both implementations. For now, though, I think it's fine....especially until we decide about how we generally want to handle additional mutable/state changes going forward.

I think it's worth pointing out, just for historical purposes, that there are still a number of ways in which the reported/actual state (seen through effective config) may not reflect the actual state of the profiler(s). We can address that later if/when it becomes an issue.

PR #2984, opened 2026-07-30
OpAMP thread dump - use custom message

breedx-splk:
Switch from using hacky remote config to using the also hacky custom message.

### DIFF
--- a/control/src/main/java/com/splunk/opamp/remotecontrol/CommandDispatcherImpl.java
+++ b/control/src/main/java/com/splunk/opamp/remotecontrol/CommandDispatcherImpl.java
@@ -38,7 +38,7 @@ public class CommandDispatcherImpl implements CommandDispatcher {
   }
 
   @Override
-  public void dispatch(String contentType, String body) {
+  public void dispatch(String type, String body) {
     String[] parts = body.split("\n");
     if (parts.length == 0) {
       logger.warning("Missing useful command body.");
--- a/control/src/main/java/com/splunk/opamp/remotecontrol/CommandDispatcherImpl.java
+++ b/control/src/main/java/com/splunk/opamp/remotecontrol/CommandDispatcherImpl.java
@@ -38,7 +38,7 @@ public class CommandDispatcherImpl implements CommandDispatcher {
   }
 
   @Override
-  public void dispatch(String type, String body) {
+  public void dispatch(String body) {
     String[] parts = body.split("\n");
     if (parts.length == 0) {
       logger.warning("Missing useful command body.");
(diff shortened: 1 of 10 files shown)


control/src/main/java/com/splunk/opamp/remotecontrol/CommandDispatcherImpl.java
```diff
@@ -38,7 +38,7 @@ public CommandDispatcherImpl(BigDumper threadDumper) {
   }
 
   @Override
-  public void dispatch(String contentType, String body) {
+  public void dispatch(String type, String body) {
```
robsunday: `type` parameter that is unused, yet passed always as a "command" string, looks awkward to me. After some time it will be confusing.
I think this parameter can be removed, since this class is already named as command dispatcher.
If you believe that it may be used in the future then value check should be added for it.

robsunday:
I like this solution much more than the previous one! Good change.

control/src/main/java/com/splunk/opamp/remotecontrol/CommandDispatcherImpl.java
breedx-splk: yeah it felt weird to build something that handles a custom message without allowing that thing to know the type of custom message it was handling. Your point is legit tho -- there really is only one of these and it's checked to only handle `command` messages, so I think it's probably good to remove it. Future proofing isn't warranted here (famous last words). I'll simplify it.

opamp/src/main/java/com/splunk/opentelemetry/opamp/ServerToAgentMessageHandler.java
```diff
@@ -21,11 +21,12 @@
...
 import io.opentelemetry.opamp.client.OpampClient;
 import io.opentelemetry.opamp.client.internal.response.MessageData;
-import opamp.proto.AgentConfigFile;
 import opamp.proto.AgentRemoteConfig;
+import opamp.proto.CustomMessage;
 
 public class ServerToAgentMessageHandler {
-  public static final String MAGIC_CMD_STRING = "COMMAND_HACKS";
+  public static final String HACKY_CMD_CAPABILITY = "com.splunk.opamp.experimental_command/v1";
+  public static final String HACKY_CMD_TYPE = "command";
```
laurit: rename these fields

laurit: so the current idea is to set the type to command `command` and get the actual command from the message body, did you consider having separate types for different commands e.g. `thread-dump`?

breedx-splk: > rename these fields

I think you're asking me to hide my shame.  🙈 

> so the current idea is to set the type to command command and get the actual command from the message body, did you consider having separate types for different commands e.g. thread-dump?

I did consider it, and @robsunday also pointed out something similar. I'm not sure that this will be around for long, or if it's even a good idea, so I thought it would be marginally better to keep it contained within one type. If we did want to promote commands up to `type` then we would probably want to consider a namespace so that all commands aren't in some flat hierarchy....but now it's already over-engineered. 

If you feel strongly about this we can change it. I certainly don't have a strong preference.

laurit: I was thinking that this way the consumer of the message wouldn't need to parse the message body to know it is a thread dump which might make it easier to handle.

laurit (approved):

PR #2988, opened 2026-08-06
Avoid warning

laurit:
> Task :testing:agent-for-testing:shadowJar
'inst/META-INF/kotlin-stdlib-jdk7.shadow.kotlin_module' is matched by com.github.jengelman.gradle.plugins.shadow.transformers.KotlinModuleMetadataTransformer_Decorated but its DuplicatesStrategy is EXCLUDE — duplicates may be silently dropped before the transformer processes them.
Set it to INCLUDE or WARN to ensure all duplicates are processed by the transformer.
See https://gradleup.com/shadow/configuration/merging/#handling-duplicates-strategy for more details.


robsunday (approved):
Looks good, however I'm not the `gradle` expert.

PR #2991, opened 2026-08-07
`APPLYING` status is set when processing of Remote Config starts

robsunday:
So far RemoteConfigStatus was set only at the end of remote config processing to `APPLIED` or `FAILED`.
According to the [SPEC](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#remoteconfigstatus-message) it can also be set to `APPLYING` at the beginning of remote config application.
Usually server will get final remote config status, but in case of some bug that prevents remote config application to finish the server should get the `APPLYING` status.

### DIFF
--- a/opamp/src/test/java/com/splunk/opentelemetry/opamp/RemoteConfigProcessorImplTest.java
+++ b/opamp/src/test/java/com/splunk/opentelemetry/opamp/RemoteConfigProcessorImplTest.java
@@ -18,6 +18,7 @@ package com.splunk.opentelemetry.opamp;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoInteractions;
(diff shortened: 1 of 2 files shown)


laurit (approved):

opamp/src/test/java/com/splunk/opentelemetry/opamp/RemoteConfigProcessorImplTest.java
```diff
@@ -68,6 +70,22 @@ void tearDown() {
...
+    ByteString configHash = ByteString.encodeUtf8("test-config-hash");
+    AgentRemoteConfig remoteConfig = createRemoteConfig(configHash, "test-config:");
+
+    // when
+    handler.applyConfig(remoteConfig, opampClient);
+
+    // then
+    RemoteConfigStatus status = getInitialReportedRemoteConfigStatus();
+    assertThat(status.last_remote_config_hash).isEqualTo(configHash);
+    assertThat(status.status).isEqualTo(RemoteConfigStatuses.RemoteConfigStatuses_APPLYING);
```
breedx-splk, line 85: 👍🏻

breedx-splk (approved):

PR #3000, opened 2026-08-11
Add workflow to automatically update csa release version.

breedx-splk:
So the idea here is to have a daily process that checks to see if there is a new release of CSA, and if there is, update which version we're using.

The script got a little long, mostly because there are a lot of safety checks along the way, which I figured was probably ok. There's also a dry run which feels overengineered but allows one to test locally without side effects.

Disclaimer: codex wrote most of this.


breedx-splk:
replaced by #3001

PR #3024, opened 2026-08-20
Remove the remote control command for taking a thread dump.

breedx-splk:
Yup.


PeterF778:
> Yup.

But why?

breedx-splk:
> > Yup.
> 
> But why?

It's not longer being asked for. Profiling features to be leveraged and enhanced instead. 🤷🏻

laurit (approved):

PR #3025, opened 2026-08-21
Automatically regenerate license reports for renovate PRs

breedx-splk:
We've had this problem forever, where renovate PRs fail the build due to the new version of a dep causing the license report to be outdated and fail the license check. Usually a human checks out the branch, regenerates the report, and then pushes it to the renovate PR branch. This is fine, but it's slow, error prone, and somewhat annoying.

So this modifies the PR workflow to automatically regenerate the license report and push it to the PR branch. It should only happen for renovate PRs.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,11 +1,17 @@
 name: PR build
 
 concurrency:
-  group: pr-${{ github.event.pull_request.number }}
+  group: pr-${{ github.event.pull_request.number || inputs.pull_request_number || github.ref_name }}
   cancel-in-progress: true
 
 on:
   pull_request:
+  workflow_dispatch:
+    inputs:
+      pull_request_number:
+        description: Pull request number associated with the branch
+        required: true
+        type: number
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -226,6 +239,60 @@ jobs:
...
+          git add licenses
+
+          if git diff --cached --quiet
+          then
+            echo "License report is already up-to-date on the Renovate branch."
+            exit 0
+          fi
+
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
laurit, line 291: is this used anywhere else? will it pass the cla check?

laurit (approved):

.github/workflows/pr.yaml
breedx-splk: CLA skips it apparently https://github.com/contributor-assistant/github-action/blob/v2.6.1/dist/index.js#L135

PR #3030, opened 2026-08-25
Config setting for snapshot thread monitor/lock info

breedx-splk:
This introduces a new config setting called `splunk.profiler.locks.enabled` which defaults to false. This effectively rolls back #2952 unless the user opts in with the new config setting.

(used codex)

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
@@ -44,7 +44,20 @@ class PeriodicStackTraceSampler implements StackTraceSampler {
 
   public PeriodicStackTraceSampler(
       Supplier<StagingArea> staging, Supplier<SpanTracker> spanTracker, Duration samplingPeriod) {
-    this(staging, spanTracker, new ThreadInfoCollector(), samplingPeriod, Clock.getDefault());
+    this(staging, spanTracker, samplingPeriod, false);
+  }
+
+  public PeriodicStackTraceSampler(
+      Supplier<StagingArea> staging,
+      Supplier<SpanTracker> spanTracker,
+      Duration samplingPeriod,
+      boolean locksEnabled) {
+    this(
+        staging,
+        spanTracker,
+        new ThreadInfoCollector(locksEnabled),
+        samplingPeriod,
+        Clock.getDefault());
   }
 
   @VisibleForTesting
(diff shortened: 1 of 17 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/PeriodicStackTraceSampler.java
```diff
@@ -44,7 +44,20 @@ class PeriodicStackTraceSampler implements StackTraceSampler {
 
   public PeriodicStackTraceSampler(
```
robsunday: For final solution I'd get rid of this constructor. It is now used directly only by tests and is redundant.

profiler/src/test/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingDeclarativeConfigurationFactoryTest.java
```diff
@@ -42,6 +42,8 @@ void shouldMapYamlToConfiguration() {
                     selection_probability: 0.0123 # SPLUNK_SNAPSHOT_SELECTION_PROBABILITY
                     stack_depth: 200              # SPLUNK_SNAPSHOT_STACK_DEPTH
                     staging_capacity: 7           # SPLUNK_SNAPSHOT_STAGING_CAPACITY
+                    locks:
```
robsunday: Not sure if additional nesting level gives any benefit here. I'd rather use:
`report_locks: true` or similar.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingConfiguration.java
```diff
@@ -79,6 +82,7 @@ public void log() {
     log("SamplingInterval", getSamplingInterval().toMillis() + "ms");
     log("ExportInterval", getExportInterval().toMillis() + "ms");
     log("StagingCapacity", getStagingCapacity());
+    log("LocksEnabled", getLocksEnabled());
```
laurit: does it make sense to announce a feature that can't really be used since the info isn't displayed anywhere even when you enable it?

laurit (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/ThreadInfoCollector.java
```diff
@@ -32,14 +32,25 @@ class ThreadInfoCollector {
   private static final Logger logger = Logger.getLogger(ThreadInfoCollector.class.getName());
 
   private final ThreadMXBean threadMXBean;
+  private final boolean locksEnabled;
 
   ThreadInfoCollector() {
```
robsunday: For clarity I'd recommend to keep only constructors that accept `locksEnabled` parameter

robsunday (approved):
I think it would benefit from some additional changes, but looks like a good quick fix

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingConfiguration.java
breedx-splk: "Announce" might be an overstatement, but you're right of course, logging this right now has basically no utility. I'll remove it.

PR #3031, opened 2026-08-27
[OpAMP] Snapshot profiling remote config

robsunday:
Until now it was only possible to turn on and off snapshot profiling with remote config.
This PR introduces support for sampling interval and snapshot selection probability settings.

### DIFF
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
@@ -26,4 +26,6 @@ public interface SnapshotProfilingSpanProcessor extends SpanProcessor {
   void setEnabled(boolean enabled);
 
   boolean isEnabled();
+
+  void setSnapshotSelectionProbability(double probability);
 }
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessorImpl.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessorImpl.java
@@ -41,6 +41,11 @@ public class SnapshotProfilingSpanProcessorImpl implements SnapshotProfilingSpan
   }
 
   @Override
+  public void setSnapshotSelectionProbability(double probability) {
+    this.selector.setSnapshotSelectionProbability(probability);
+  }
+
+  @Override
   public void onStart(Context context, ReadWriteSpan span) {
     if (!isEnabled()) {
       return;
--- a/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessorImpl.java
+++ b/profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessorImpl.java
@@ -42,7 +42,7 @@ public class SnapshotProfilingSpanProcessorImpl implements SnapshotProfilingSpan
 
   @Override
   public void setSnapshotSelectionProbability(double probability) {
-    this.selector.setSnapshotSelectionProbability(probability);
+    selector.setSnapshotSelectionProbability(probability);
   }
 
   @Override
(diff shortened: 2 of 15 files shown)


profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceIdBasedSnapshotSelector.java
```diff
@@ -18,22 +18,18 @@
...
+final class TraceIdBasedSnapshotSelector implements SnapshotSelector {
   // the length of trace-id represented as a hex string
   private static final int TRACE_ID_LENGTH = 32;
   // number of characters taken from the tail of trace-id as the trace randomness
   private static final int CHARS = 7;
   private static final String HEX_FORMATTER = "%07x";
   // to convert probability to threshold
   private static final int MULTIPLIER = (1 << (4 * CHARS)) - 1;
-  private final String threshold;
+  private volatile String threshold;
```
robsunday, line 29: [for reviewer] This value can be modified and read by multiple threads.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessor.java
```diff
@@ -26,4 +26,6 @@ public interface SnapshotProfilingSpanProcessor extends SpanProcessor {
   void setEnabled(boolean enabled);
 
   boolean isEnabled();
+
+  void setSnapshotSelectionProbability(double probability);
```
robsunday, line 30: [for reviewer] Instance of span processor is created in configuration customizers and possibly registered somewhere by SDK, so it is safer to keep the instance unchanged and just reconfigure it.

breedx-splk: s/reconfigure/mutate/

unfortunately, yeah.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/SnapshotProfilingSpanProcessorImpl.java
```diff
@@ -40,6 +40,11 @@ public class SnapshotProfilingSpanProcessorImpl implements SnapshotProfilingSpan
     this.orphanedTraceCleaner = new OrphanedTraceCleaner(registry);
   }
 
+  @Override
+  public void setSnapshotSelectionProbability(double probability) {
+    this.selector.setSnapshotSelectionProbability(probability);
```
breedx-splk: Unnecessary `this`?

```suggestion
   selector.setSnapshotSelectionProbability(probability);
```

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceIdBasedSnapshotSelector.java
```diff
@@ -47,13 +43,23 @@ private static String thresholdFor(double probability) {
 
   @Override
   public boolean select(SpanContext spanContext) {
-    if (threshold == null || !spanContext.isValid()) {
+    String currentThreshold = threshold;
```
breedx-splk, line 48: Why introduce a local variable here? Is this some trick to try and reduce the reads on the volatile? Does that even work? Seems like something the JITter would handle anyway eh?

breedx-splk (approved):
Seems fine to me. I don't love the fact that merely implementing this feature increases overhead with new volatile reads in the hot path, but here we are.

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceIdBasedSnapshotSelector.java
laurit: It is probably to ensure that the same value for `threshold` would be used for the whole method. Imagine that threshold is update while the method is running, if the new probability is `0` the threshold is set to `null`.

laurit (approved):

profiler/src/main/java/com/splunk/opentelemetry/profiler/snapshot/TraceIdBasedSnapshotSelector.java
robsunday: Yes, what @laurit said is the reason of introducing local variable. I added a comment with explanation.

breedx-splk: Ah yes thanks buds, that makes sense (why didn't I come up with that! Derp.).

PR #3046, opened 2026-09-04
Try an alternative way for commiting updated license report

laurit:
The problem with the current approach is that the commit isn't signed which makes the PR unmergeable because we have a rule that allows only signed commits. Instead of committing the changes manually this PR uses https://github.com/IAreKyleW00t/verified-bot-commit that goes through github api and should produce a verified commit from `github-actions[bot]`. Didn't test.


breedx-splk (approved):
Thanks. This looks like a viable approach! I don't love the additional action dependency, but let's try it out.

PR #3065, opened 2026-09-17
Fix that would allow use JDK 21+ to build the project

robsunday:
Build failed on JDK 21 because `JfrContextStorageTest.testAttachLifecycle` test failed.
It was caused by the following sequence of actions performed on test start:

1. Mockito’s inline mock maker retransforms ContextAttached.
2. Java 21’s JFR machinery detects that it is an Event subclass and retransforms it again.
3. JFR replaces begin(), commit(), and shouldCommit() with its own implementations. OpenJDK explicitly generates these methods during JFR instrumentation.
4. Mockito’s interception of shouldCommit() is consequently lost.
5. The real method returns false, so this expression effectively becomes: `when(false).thenReturn(true)` . Since Mockito recorded no invocation, it throws `MissingMethodInvocationException`.

With this PR Mockito can mock JFR events.


profiler/src/main/java/com/splunk/opentelemetry/profiler/events/ContextAttached.java
```diff
@@ -28,7 +28,7 @@
 @Label("otel context attached")
 @Category("opentelemetry")
 @StackTrace(false)
-public class ContextAttached extends Event {
+public class ContextAttached extends Event implements JfrEvent {
```
robsunday, line 31: The inherited final JFR methods satisfy the interface in production, while Mockito can mock the interface without competing with JFR bytecode instrumentation.

profiler/src/main/java/com/splunk/opentelemetry/profiler/events/JfrEvent.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.splunk.opentelemetry.profiler.events;
+
+public interface JfrEvent {
```
laurit, line 36: add a comment here that explains why this class is necessary

laurit (approved):

